PR #9, opened 2026-04-16
feat: implement things-bridge and things-cli (read-only)

aidanns:
## Summary

- Adds `things-bridge`: a local HTTP server that wraps the Things 3 macOS app via AppleScript (`osascript`) and delegates token validation to `agent-auth`. Exposes six read-only GET endpoints under `/things-bridge/` (todos/projects/areas, list+show) all gated on the `things:read` scope.
- Adds `things-cli`: a thin client that handles the access/refresh/reissue dance against `agent-auth` on 401, with credentials in the system keyring by default and a 0600-mode JSON `FileStore` fallback for devcontainer use. Commands: `login`, `logout`, `status`, `todos list/show`, `projects list/show`, `areas list/show`, all accepting `--json`.
- Stdlib only — no new runtime deps. 151 tests (up from 135), all green. `AppleScriptRunner` is abstracted so the bridge suite runs fully on Linux without `osascript`.
- Renames the generic `Example App Bridge` / `Example App CLI` entries in `design/functional_decomposition.yaml` and `design/product_breakdown.yaml` to the concrete `things-bridge` / `things-cli`, and regenerates the D2/SVG/PNG/CSV/MD artefacts. `scripts/verify-design.sh` still passes.
- Writes are a follow-up; the plan is committed at `plans/implement-things-cli.md`.

## Security posture

Several items were tightened in response to a code-review pass before opening the PR:

- `_quote` in `things-bridge` rejects control characters and the U+241E/U+241F TSV placeholders in any caller-supplied id/tag, so URL-derived input cannot break out of the single-line AppleScript string literal.
- A `_safe_id` gate runs before the JIT approval description is built so a malicious id cannot inject newlines or terminal escapes into the approval prompt or audit log. Also caps id length at 128 bytes.
- AppleScript `osascript` stderr is no longer forwarded in HTTP error bodies — it routinely contains local filesystem paths and usernames. Errors now return canonical codes only.
- `AuthzClient` uses `HTTPSConnection` for `https://` auth URLs (previously used `HTTPConnection` unconditionally).
- Only `log_request` is suppressed, not `log_message`, so uncaught handler exceptions still surface via `log_error` instead of being silently dropped.

## Test plan

- [x] `./scripts/test.sh` passes locally (151 tests green)
- [x] `./scripts/verify-design.sh` still reports "All functions allocated"
- [x] `./scripts/verify-function-tests.sh` does not regress (pre-existing TODO)
- [x] `things-cli --help`, `things-cli todos --help`, `things-bridge --help` print sensible usage
- [x] `things-cli login ... && things-cli status` shows `<set>` for tokens and does not leak plaintext
- [x] `things-cli logout` removes the credentials file / keyring entries
- [x] GitHub Actions: the `test`, `verify-design`, and `verify-function-tests` workflows all pass on the PR


src/things_bridge/things.py
```diff
@@ -0,0 +1,504 @@
...
+def _quote(value: str) -> str:
+    """AppleScript string-literal quoting for caller-supplied ids, names, and tags.
+
+    Rejects any input containing characters that cannot appear inside a
+    single-line AppleScript string literal, that would corrupt the TSV framing,
+    or that could otherwise alter the surrounding script. This is the primary
+    defence against AppleScript injection via URL-derived ids.
+    """
+    for ch in value:
+        if ch in _FORBIDDEN_ID_CHARS or ord(ch) < 0x20:
```
aidanns: Claude Reviewer: **MEDIUM — inconsistency between `_quote` and `_safe_id` on DEL (0x7F)**

`_safe_id` (server.py:38) rejects `ord(ch) == 0x7F`, but `_quote` (this line) only rejects `ord(ch) < 0x20`. Query-parameter inputs (`list`, `project`, `area`, `tag`) never pass through `_safe_id` — they go straight from `_first(params, ...)` to `ThingsClient.list_todos(...)` and into `_quote`. A caller could therefore send `?tag=foo%7Fbar` and the DEL byte would reach the AppleScript string literal.

Not a demonstrated break-out, but the discrepancy is suspicious: if `_safe_id` thought DEL was worth rejecting defensively, `_quote` should at least match. Suggest tightening the condition to `ord(ch) < 0x20 or ord(ch) == 0x7F`.

src/things_cli/cli.py
```diff
@@ -0,0 +1,246 @@
...
+    if args.status:
+        params["status"] = args.status
+    data = client.get("/things-bridge/todos", params=params or None)
+    output.print_todos(data.get("todos", []), as_json=args.json)
+    return 0
+
+
+def handle_todo_show(args) -> int:
+    client = _load_client(args)
+    data = client.get(f"/things-bridge/todos/{args.id}")
```
aidanns: Claude Reviewer: **MEDIUM — user-supplied id is interpolated into URL path without encoding**

`args.id` comes from `things-cli todos show <id>` and lands directly in the request line. If a user runs `things-cli todos show 'foo bar'` or `things-cli todos show foo#frag`, the client will emit a malformed HTTP request (raw space) or silently strip the fragment. Even for well-formed Things ids this isn't exploitable, but it's surprising UX and the fix is small.

Use `urllib.parse.quote(args.id, safe='')` before interpolation. Applies to `handle_todo_show`, `handle_project_show`, and `handle_area_show` (lines 94, 111, 125).

src/things_cli/client.py
```diff
@@ -0,0 +1,181 @@
...
+    ) -> dict:
+        status, data = self._do_http(
+            self._credentials.bridge_url,
+            method,
+            path,
+            params=params,
+            headers=self._auth_headers(),
+        )
+        if 200 <= status < 300:
+            return data
```
aidanns, line 90: Claude Reviewer: **LOW — empty 2xx body crashes the client**

`_do_http` returns `(status, None)` for an empty response body (line 175). `_request` then returns `None` on the `2xx` path (line 61), and the CLI handlers immediately call `data.get(...)` — which will `AttributeError` on `NoneType`. The bridge always sends a JSON body today so this won't trigger in-house, but a proxy that truncates or a future bridge change would surface as a crash instead of a typed `BridgeUnavailableError`.

Suggest treating `data is None` on a 2xx as `BridgeUnavailableError("empty response")`, or returning `{}`.

src/things_bridge/server.py
```diff
@@ -0,0 +1,242 @@
...
+        body = json.dumps({"error": "method_not_allowed"}).encode("utf-8")
+        self.send_header("Content-Type", "application/json")
+        self.send_header("Content-Length", str(len(body)))
+        self.end_headers()
+        self.wfile.write(body)
+
+    do_POST = _method_not_allowed  # noqa: N815
+    do_PUT = _method_not_allowed  # noqa: N815
+    do_PATCH = _method_not_allowed  # noqa: N815
+    do_DELETE = _method_not_allowed  # noqa: N815
```
aidanns, line 216: Claude Reviewer: **LOW — HEAD/OPTIONS bypass the 405 handler**

The override stops `POST/PUT/PATCH/DELETE` but not `HEAD`, `OPTIONS`, or `TRACE`. `BaseHTTPRequestHandler` responds to unknown methods with `501 Unsupported method`, which is a correctness nit rather than a security issue, but inconsistent with how POST is handled. Either extend the alias list or delete the per-method aliases and implement `handle_one_request` / an `_all_methods` dispatch.

Also: in some contexts `HEAD` on a read-only resource should succeed. If that's out of scope for now, leaving the 501 is fine — worth a comment saying so.

src/things_cli/client.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+        if status == 200 and data:
+            self._credentials.access_token = data["access_token"]
+            self._credentials.refresh_token = data["refresh_token"]
+            self._store.save(self._credentials)
+            return
+
+        error_code = (data or {}).get("error") or "reissue_failed"
+        if status == 403:
+            raise BridgeUnauthorizedError("reissue_denied")
```
aidanns: Claude Reviewer: **LOW — 403 reissue response loses the server-provided error code**

On `status == 403` this raises `BridgeUnauthorizedError("reissue_denied")` regardless of what the server actually returned in `error`. So if agent-auth ever returns a more specific code (e.g. `user_denied_prompt`, `family_revoked`, `cooldown_active`), the user sees the generic "reissue_denied" and we silently drop the diagnostic.

Either pass `error_code` through unconditionally, or branch explicitly on known codes.

aidanns:
Claude Reviewer: Independent review pass complete — 5 comments posted inline.

**Summary by severity:**
- Medium (2): `_quote` vs `_safe_id` DEL-character inconsistency (src/things_bridge/things.py:319); missing URL-encoding of `args.id` in `handle_todo_show` / `handle_project_show` / `handle_area_show` (src/things_cli/cli.py:94).
- Low (3): empty 2xx body crashes caller (client.py:61); HEAD/OPTIONS miss the 405 path (server.py:214); 403-on-reissue silently drops server-provided error code (client.py:128).

**Top priorities the author should address:**
1. Tighten `_quote` to match `_safe_id` on 0x7F — closes a consistency gap where query-param inputs have laxer validation than path-id inputs.
2. URL-encode `args.id` in the three show-by-id handlers — free correctness win even if no id today contains special chars.

No high-severity findings; the pre-opening self-review pass covered the major injection/host-info-leak surface well.

src/things_bridge/things.py
aidanns: Claude: Done! `_quote` now rejects DEL (0x7F) in addition to the C0 range, matching `_safe_id`. Updated test parametrize list to exercise `\x7f` through the query-param path.

src/things_cli/cli.py
aidanns: Claude: Done! Added `from urllib.parse import quote` and applied `quote(args.id, safe='')` in `handle_todo_show`, `handle_project_show`, and `handle_area_show`. New parametrized regression test in `test_things_cli_cli.py` covers spaces, slashes, and `#` in ids.

src/things_cli/client.py
aidanns: Claude: Done! `_request` now raises `BridgeUnavailableError` when a 2xx response has an empty body, with a descriptive message. Regression test `test_empty_2xx_body_raises_unavailable` added.

src/things_bridge/server.py
aidanns: Claude: Done! Added `do_HEAD` and `do_OPTIONS` aliases to the 405 handler. Parametrized regression test `test_non_get_methods_return_405` covers HEAD/OPTIONS/PUT/PATCH/DELETE.

src/things_cli/client.py
aidanns: Claude: Done! Merged 401 and 403 branches so `BridgeUnauthorizedError` is raised with the server-provided error code verbatim (no more relabeling to a hard-coded "reissue_denied"). Regression test `test_reissue_403_preserves_server_error_code` verifies a custom code like "family_revoked" propagates.

design/DESIGN.md
```diff
@@ -42,16 +42,55 @@ Responsibilities:
...
+| 401 | `{"error": "unauthorized"}` | Missing, malformed, or invalid bearer token |
+| 401 | `{"error": "token_expired"}` | agent-auth reported the access token has expired (CLI retries with refresh) |
+| 403 | `{"error": "scope_denied"}` | Token does not carry `things:read` |
+| 404 | `{"error": "not_found"}` | Unknown path or unknown Things id |
+| 405 | `{"error": "method_not_allowed"}` | Non-GET verb on a read-only endpoint (writes are a follow-up) |
+| 502 | `{"error": "authz_unavailable"}` | agent-auth is unreachable |
+| 502 | `{"error": "things_unavailable"}` | AppleScript or Things reported an error |
+| 503 | `{"error": "things_permission_denied"}` | macOS Automation permission not granted for Things |
+
+Error bodies intentionally omit server-side detail: AppleScript / `osascript` stderr can contain local filesystem paths and usernames, so the bridge returns only a canonical error code.
```
aidanns: ensure this data is written to the service log instead

design/DESIGN.md
```diff
@@ -42,16 +42,55 @@ Responsibilities:
...
 Responsibilities:
 - Provide a CLI interface for the application
 - Pass the bearer token from local credential storage
-- Automatically refresh expired access tokens using the refresh token
+- Automatically refresh expired access tokens using the refresh token; if the refresh token itself has expired, request re-issuance (which blocks on JIT approval)
 - Store credentials in the system keyring (see CLI Credential Storage below)
 
+`things-cli` ships the following read-only commands (paired with `login` / `logout` / `status` for credential management):
+
+- `things-cli todos list [--list ID] [--project ID] [--area ID] [--tag NAME] [--status open|completed|canceled]`
```
aidanns, line 85: are these IDs meaningful to the user? Should we make these a regex or name match instead?

aidanns: or should we support a "todos search" subcommand?

design/functional_decomposition.yaml
```diff
@@ -103,17 +103,17 @@ functions:
         description: Parse arguments and invoke token family rotation.
       - name: Handle Serve Command
         description: Start the agent-auth HTTP server.
-  - name: Example App Bridge
-    description: Proxy requests from the CLI to an external system, delegating authorization to agent-auth. Each external system gets its own bridge.
+  - name: Things Bridge
+    description: Proxy requests from the CLI to the Things 3 application via AppleScript, delegating authorization to agent-auth.
```
aidanns: I think the via AppleScript is an implementation detail that isn't relevant to the functional decomp

src/things_bridge/authz.py
```diff
@@ -0,0 +1,72 @@
...
+    AuthzTokenExpiredError,
+    AuthzTokenInvalidError,
+    AuthzUnavailableError,
+)
+
+
+class AuthzClient:
+    """Validates bearer tokens against agent-auth's ``/agent-auth/validate`` endpoint."""
+
+    def __init__(self, auth_url: str, *, timeout: float = 30.0):
```
aidanns: should specify the unit for this parameter

README.md
```diff
@@ -83,6 +83,47 @@ curl -H "Authorization: Bearer aa_<id>_<sig>" \
...
+### things-bridge (macOS host)
+
+`things-bridge` is an HTTP server that wraps the Things 3 application on macOS via AppleScript. It delegates token validation to `agent-auth` and exposes read-only endpoints under `/things-bridge/`. Run it alongside `agent-auth serve`:
+
+```bash
+# Start the bridge (default: 127.0.0.1:9200)
+things-bridge serve
+
+# Customize bindings and point at a non-default agent-auth URL
+things-bridge serve --host 127.0.0.1 --port 9200 --auth-url http://127.0.0.1:9100
```
aidanns: move these things to the config file

README.md
```diff
@@ -83,6 +83,47 @@ curl -H "Authorization: Bearer aa_<id>_<sig>" \
...
+### things-cli
+
+`things-cli` is a thin client for `things-bridge` that auto-refreshes/reissues tokens via `agent-auth`. Credentials are kept in the system keyring by default; pass `--credential-store=file --credentials-file=<path>` to persist them in a 0600 JSON file (useful inside a devcontainer).
+
+```bash
+# Save credentials from an `agent-auth token create` output
+things-cli login \
+  --bridge-url http://127.0.0.1:9200 \
+  --auth-url http://127.0.0.1:9100 \
+  --access-token aa_<id>_<sig> \
```
aidanns: probably shouldn't get the user to enter the key in a CLI like this - it will be saved to bash_history

aidanns: update this so that either the user pastes the credentials in to the CLI terminal (like gh CLI does), or the user preconfigured the credentials in the configuration file on disk.

README.md
```diff
@@ -83,6 +83,47 @@ curl -H "Authorization: Bearer aa_<id>_<sig>" \
...
+  --refresh-token rt_<id>_<sig> \
+  --family-id <family-id>
+
+# Show redacted credential status
+things-cli status
+
+# Read-only commands (add --json for structured output)
+things-cli todos list
+things-cli todos list --list TMTodayListSource --status open
```
aidanns: what does this mean?

README.md
```diff
@@ -83,6 +83,47 @@ curl -H "Authorization: Bearer aa_<id>_<sig>" \
...
+# Start the bridge (default: 127.0.0.1:9200)
+things-bridge serve
+
+# Customize bindings and point at a non-default agent-auth URL
+things-bridge serve --host 127.0.0.1 --port 9200 --auth-url http://127.0.0.1:9100
+```
+
+### things-cli
+
+`things-cli` is a thin client for `things-bridge` that auto-refreshes/reissues tokens via `agent-auth`. Credentials are kept in the system keyring by default; pass `--credential-store=file --credentials-file=<path>` to persist them in a 0600 JSON file (useful inside a devcontainer).
```
aidanns: update the CLI to automatically use a 0600 file on disk if the system keyring isn't available

aidanns: this should only be done for CLIs - the server should fail to start if it can't access the keyring.

src/things_bridge/cli.py
```diff
@@ -0,0 +1,59 @@
...
+from things_bridge.server import run_server
+from things_bridge.things import AppleScriptRunner, ThingsClient
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="things-bridge",
+        description="HTTP bridge from agent-auth-protected clients to the Things 3 AppleScript API.",
+    )
+    parser.add_argument("--config-dir", help="Override configuration directory")
```
aidanns: not sure it's value-add to support this - is there a reason that it is needed?

src/things_bridge/cli.py
```diff
@@ -0,0 +1,59 @@
...
+
+    subparsers = parser.add_subparsers(dest="command")
+
+    serve_parser = subparsers.add_parser("serve", help="Start the HTTP bridge server")
+    serve_parser.add_argument("--host", help="Bind address (default: from config)")
```
aidanns: don't allow overriding the config

src/things_bridge/authz.py
```diff
@@ -0,0 +1,72 @@
...
+
+from things_bridge.errors import (
+    AuthzScopeDeniedError,
+    AuthzTokenExpiredError,
+    AuthzTokenInvalidError,
+    AuthzUnavailableError,
+)
+
+
+class AuthzClient:
```
aidanns: call this AgentAuthClient instead.

aidanns: add a github issues for moving this to its own library (it will be re-used by many bridges)

src/things_bridge/authz.py
aidanns: why * here?

aidanns: prefer semantic type names throughout all source files

src/things_bridge/config.py
```diff
@@ -0,0 +1,64 @@
...
+    base = os.environ.get(env_var) or os.path.join(os.path.expanduser("~"), *fallback_segments)
+    return os.path.join(base, "things-bridge")
+
+
+def _default_config_dir() -> str:
+    return _xdg_dir("XDG_CONFIG_HOME", (".config",))
+
+
+def _default_state_dir() -> str:
+    return _xdg_dir("XDG_STATE_HOME", (".local", "state"))
```
aidanns, line 24: maybe worth a small library shared between this and the agent-auth service?

src/things_bridge/config.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@dataclass
+class Config:
+    config_dir: str = ""
```
aidanns: config_dir isn't needed here - the config is already loaded. should be convention only.

src/things_bridge/config.py
```diff
@@ -0,0 +1,64 @@
...
+    host: str = "127.0.0.1"
+    port: int = 9200
+    auth_url: str = "http://127.0.0.1:9100"
+    osascript_path: str = "/usr/bin/osascript"
+    request_timeout_seconds: float = 30.0
+    log_path: str = ""
+
+    def __post_init__(self):
+        if not self.log_path:
+            self.log_path = os.path.join(_default_state_dir(), "bridge.log")
```
aidanns: things-bridge.log

src/things_bridge/models.py
```diff
@@ -0,0 +1,55 @@
+"""Dataclass models for Things objects returned by the bridge."""
+
+from dataclasses import asdict, dataclass
+
+
+@dataclass
+class Area:
+    id: str
```
aidanns, line 8: use semantic types throughout to avoid mixing up variables

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+    AuthzTokenExpiredError,
+    AuthzTokenInvalidError,
+    AuthzUnavailableError,
+    ThingsError,
+    ThingsNotFoundError,
+    ThingsPermissionError,
+)
+from things_bridge.things import ThingsClient
+
+READ_SCOPE = "things:read"
```
aidanns, line 20: create different scopes for each data type

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+
+def _safe_id(raw: str | None) -> str | None:
+    """Reject ids that contain path separators, control chars, or are over-length.
+
+    Returns the id unchanged if safe, ``None`` otherwise. Used before building
+    audit/JIT description strings and before passing to ThingsClient.
+    """
+    if raw is None or not raw or len(raw) > _MAX_ID_LEN:
+        return None
+    if "/" in raw:
```
aidanns: can we instead make this an allow-list of characters?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+        return None
+    for ch in raw:
+        if ord(ch) < 0x20 or ord(ch) == 0x7F:
+            return None
+    return raw
+
+
+class ThingsBridgeHandler(BaseHTTPRequestHandler):
+    """HTTP request handler for things-bridge read-only endpoints."""
```
aidanns: this is likely to change, so remove the read-only here - it isn't important.

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def _send_json(self, status: int, data: dict):
+        body = json.dumps(data).encode("utf-8")
+        self.send_response(status)
+        self.send_header("Content-Type", "application/json")
+        self.send_header("Content-Length", str(len(body)))
+        self.end_headers()
+        self.wfile.write(body)
+
+    def log_request(self, code="-", size="-"):  # noqa: ARG002
```
aidanns, line 62: where is this called?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+        except AuthzTokenInvalidError:
+            self._send_json(401, {"error": "unauthorized"})
+            return False
+        except AuthzScopeDeniedError:
+            self._send_json(403, {"error": "scope_denied"})
+            return False
+        except AuthzUnavailableError:
+            self._send_json(502, {"error": "authz_unavailable"})
+            return False
+        return True
```
aidanns: default return True feels off here - can we return the result of self._bridge.authz.validate() instead?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+        token = self._extract_bearer()
+        if token is None:
+            self._send_json(401, {"error": "unauthorized"})
+            return
+
+        things: ThingsClient = self._bridge.things
+
+        # Routing: longest-prefix specific paths first.
+        if path == "/things-bridge/todos":
+            if not self._validate(token, "List Things todos"):
```
aidanns, line 119: is there a way to deterministically validate that we are checking auth on every endpoint? I don't trust Claude Code or human reviewers to get this right.

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+                return
+            try:
+                todos = things.list_todos(
+                    list_id=_first(params, "list"),
```
aidanns, line 123: is dropping all other query parameters with the same key the right approach? should we instead treat these as OR? I'm not sure what is semantically correct for REST APIs

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+                    tag=_first(params, "tag"),
+                    status=_first(params, "status"),
+                )
+            except ThingsError as exc:
+                self._things_error_response(exc)
+                return
+            self._send_json(200, {"todos": [t.to_json() for t in todos]})
+            return
+
+        if path.startswith("/things-bridge/todos/"):
```
aidanns, line 135: we should have a handler for every HTTP endpoint that audits that it was called to an audit log

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+from things_bridge.things import ThingsClient
+
+READ_SCOPE = "things:read"
+
+# Upper bound on ids accepted from URL paths. Things ids are short; reject
+# anything excessive before it ever reaches AppleScript.
+_MAX_ID_LEN = 128
+
+
+def _safe_id(raw: str | None) -> str | None:
```
aidanns, line 27: should use a change of type for the output here to ensure that paths are properly transformed through this function.

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+            except ThingsError as exc:
+                self._things_error_response(exc)
+                return
+            self._send_json(200, {"todos": [t.to_json() for t in todos]})
+            return
+
+        if path.startswith("/things-bridge/todos/"):
+            todo_id = _safe_id(path[len("/things-bridge/todos/"):])
+            if todo_id is None:
+                self._send_json(404, {"error": "not_found"})
```
aidanns, line 138: is 404 the right error code to return if we can't parse the id?

src/things_bridge/authz.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, auth_url: str, *, timeout: float = 30.0):
+        parsed = urlparse(auth_url)
+        if parsed.scheme not in ("http", "https") or not parsed.hostname:
+            raise ValueError(f"Invalid auth_url: {auth_url!r}")
+        self._host = parsed.hostname
+        self._port = parsed.port or (443 if parsed.scheme == "https" else 80)
+        self._conn_cls = HTTPSConnection if parsed.scheme == "https" else HTTPConnection
+        self._timeout = timeout
+
+    def validate(self, token: str, required_scope: str, *, description: str | None = None) -> None:
```
aidanns, line 27: why do we pass through description here?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+            return False
+        except AuthzScopeDeniedError:
+            self._send_json(403, {"error": "scope_denied"})
+            return False
+        except AuthzUnavailableError:
+            self._send_json(502, {"error": "authz_unavailable"})
+            return False
+        return True
+
+    def _things_error_response(self, exc: ThingsError) -> None:
```
aidanns: if we're going to use procedures like this then given them should we give them a verb name like "_send_things_error_response"?

src/things_bridge/server.py
aidanns, line 119: shouldn't we specify the scope here?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+            self._send_json(200, {"todos": [t.to_json() for t in todos]})
+            return
+
+        if path.startswith("/things-bridge/todos/"):
+            todo_id = _safe_id(path[len("/things-bridge/todos/"):])
+            if todo_id is None:
+                self._send_json(404, {"error": "not_found"})
+                return
+            if not self._validate(token, f"Read Things todo {todo_id}"):
+                return
```
aidanns, line 141: shouldn't this return an unauthorized response?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+        self.send_header("Content-Type", "application/json")
+        self.send_header("Content-Length", str(len(body)))
+        self.end_headers()
+        self.wfile.write(body)
+
+    do_POST = _method_not_allowed  # noqa: N815
+    do_PUT = _method_not_allowed  # noqa: N815
+    do_PATCH = _method_not_allowed  # noqa: N815
+    do_DELETE = _method_not_allowed  # noqa: N815
+    do_HEAD = _method_not_allowed  # noqa: N815
```
aidanns, line 217: why don't we implement HEAD?

src/things_bridge/server.py
```diff
@@ -0,0 +1,244 @@
...
+    do_OPTIONS = _method_not_allowed  # noqa: N815
```
aidanns, line 218: why don't we allow OPTIONS?

src/things_bridge/things.py
```diff
@@ -0,0 +1,507 @@
...
+    if status is None:
+        return None
+    if status not in _VALID_STATUSES:
+        raise ThingsError(
+            f"Invalid status filter: {status!r} (expected one of {sorted(_VALID_STATUSES)})"
+        )
+    return status
+
+
+class ThingsClient:
```
aidanns: maybe ThingsApplescriptClient is a better name given this is coupled to AppleScript?

src/things_cli/cli.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def handle_login(args) -> int:
+    store = _resolve_store(args)
+    creds = Credentials(
+        access_token=args.access_token,
+        refresh_token=args.refresh_token,
+        bridge_url=args.bridge_url,
+        auth_url=args.auth_url,
+        family_id=args.family_id,
```
aidanns, line 39: why do we need the family_id here?

src/things_cli/cli.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def _resolve_store(args) -> CredentialStore:
+    return select_store(
+        args.credential_store,
+        file_path=args.credentials_file or _default_file_path(),
+    )
+
+
+def handle_login(args) -> int:
```
aidanns, line 32: what is this int for?

src/things_cli/cli.py
```diff
@@ -0,0 +1,247 @@
...
+    return 0
+
+
+def handle_status(args) -> int:
+    store = _resolve_store(args)
+    try:
+        creds = store.load()
+    except CredentialsNotFoundError:
+        print("No credentials stored.")
+        return 1
```
aidanns, line 59: why are we returning int and not a bool?

src/things_cli/cli.py
```diff
@@ -0,0 +1,247 @@
...
+    if args.status:
+        params["status"] = args.status
+    data = client.get("/things-bridge/todos", params=params or None)
+    output.print_todos(data.get("todos", []), as_json=args.json)
+    return 0
+
+
+def handle_todo_show(args) -> int:
+    client = _load_client(args)
+    data = client.get(f"/things-bridge/todos/{quote(args.id, safe='')}")
```
aidanns: user of the client shouldn't need to know about the path - that logic should be encapsulated in the client.

src/things_cli/client.py
```diff
@@ -0,0 +1,186 @@
...
+    1. Attaches ``Authorization: Bearer <access_token>`` from the credential store.
+    2. On ``401 {"error": "token_expired"}`` calls ``/agent-auth/token/refresh``,
+       persists the new tokens, and retries the original request once.
+    3. If the refresh token has itself expired (``401 refresh_token_expired``),
+       calls ``/agent-auth/token/reissue`` (which blocks on host-side JIT approval)
+       and retries once.
+    4. Any further 401 surfaces as :class:`BridgeUnauthorizedError`.
+    """
+
+    def __init__(self, credentials: Credentials, store: CredentialStore, *, timeout: float = 30.0):
```
aidanns: why do we have a * here?

src/things_cli/client.py
```diff
@@ -0,0 +1,186 @@
...
+        self._credentials = credentials
+        self._store = store
+        self._timeout = timeout
+
+    @property
+    def credentials(self) -> Credentials:
+        return self._credentials
+
+    def get(self, path: str, params: dict[str, str] | None = None) -> dict:
```
aidanns: client should encapsulate the functionality that the bridge exposes, not just allow any path on the bridge to be accessed.

src/things_cli/client.py
```diff
@@ -0,0 +1,186 @@
...
+        if status == 404:
+            raise BridgeNotFoundError((data or {}).get("error") or "not_found")
+        raise BridgeUnavailableError(
+            f"Bridge returned status {status}: {(data or {}).get('error') or '<no body>'}"
+        )
+
+    def _auth_headers(self) -> dict[str, str]:
+        return {
+            "Authorization": f"Bearer {self._credentials.access_token}",
+            "Accept": "application/json",
```
aidanns, line 113: this isn't an auth header - rename the method

src/things_cli/client.py
```diff
@@ -0,0 +1,186 @@
...
+        request_body: bytes | None = None
+        send_headers = dict(headers or {})
+        if body is not None:
+            request_body = json.dumps(body).encode("utf-8")
+            send_headers.setdefault("Content-Type", "application/json")
+            send_headers["Content-Length"] = str(len(request_body))
+
+        conn = conn_cls(host, port, timeout=self._timeout)
+        try:
+            try:
```
aidanns, line 191: why do we need two try here?

src/things_cli/credentials.py
```diff
@@ -0,0 +1,206 @@
...
+        tmp_path = self._path + ".tmp"
+        fd = os.open(tmp_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+        try:
+            with os.fdopen(fd, "w") as f:
+                json.dump(data, f, indent=2)
+                f.write("\n")
+        except Exception:
+            os.unlink(tmp_path)
+            raise
+        os.replace(tmp_path, self._path)
```
aidanns, line 133: do we need to write to the temp path?

src/things_cli/credentials.py
```diff
@@ -0,0 +1,206 @@
...
+        os.chmod(self._path, 0o600)
```
aidanns: should we chmod before writing to the file?

src/things_cli/credentials.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def load(self) -> Credentials:
+        try:
```
aidanns, line 142: we should fail here loudly if the permissions on the file aren't set to 600 properly

src/things_cli/credentials.py
```diff
@@ -0,0 +1,206 @@
...
+    if backend == "file":
+        if file_path is None:
+            raise ValueError("file_path is required when backend='file'")
+        return FileStore(file_path)
+    if backend == "keyring":
+        return KeyringStore(service=service)
+    if backend == "auto":
+        if _keyring_available():
+            return KeyringStore(service=service)
+        raise CredentialsBackendError(
```
aidanns: on the client, use the on-disk file automatically iff the keyring isn't available

src/things_cli/credentials.py
```diff
@@ -0,0 +1,206 @@
...
+        return True
+
+
+class KeyringStore(CredentialStore):
+    """Persist credentials via the system keyring."""
+
+    def __init__(self, service: str = SERVICE_NAME):
+        self._service = service
+
+    def save(self, creds: Credentials) -> None:
```
aidanns, line 67: is there something we need to do here to ensure that the keyring is only readable by this CLI on MacOS?

design/DESIGN.md
aidanns: Claude: Done! Updated DESIGN.md to note that full error detail (including stderr) is written to the service log for operator diagnostics.

design/DESIGN.md
aidanns: Claude: Tracked in TODO.md under 'Search subcommand and name/regex matching'. Things entity ids are opaque UUIDs that aren't user-friendly. A `todos search` subcommand with name matching would be a good UX improvement — recorded as a design decision to address in DESIGN.md.

design/functional_decomposition.yaml
aidanns: Claude: Done! Removed the AppleScript implementation detail from the functional decomposition — the function descriptions now describe the capability without coupling to the implementation.

src/things_bridge/authz.py
aidanns: Claude: Done! Renamed `timeout` to `timeout_seconds` in both `AgentAuthClient` and `BridgeClient` constructors.

src/things_bridge/authz.py
aidanns: Claude: Done! Renamed `AuthzClient` to `AgentAuthClient` across all source files and tests.

aidanns: Claude: Done! Filed #10 to track extracting AgentAuthClient into a shared library.

src/things_bridge/authz.py
aidanns: Claude: The `*` forces all subsequent parameters to be keyword-only (PEP 3102). It prevents callers from accidentally passing `timeout_seconds` as a positional argument: `AgentAuthClient(url, 5.0)` is a TypeError, forcing `AgentAuthClient(url, timeout_seconds=5.0)` instead. This convention is used throughout the codebase for constructors with optional configuration.

aidanns: Claude: Addressed the immediate case by renaming `timeout` → `timeout_seconds`. Tracked the broader principle in TODO.md under 'Semantic type names throughout all source files' — future plans should include a 'name parameters semantically' checklist item.

src/things_bridge/authz.py
aidanns: Claude: The `description` field is forwarded to agent-auth's `/agent-auth/validate` endpoint. agent-auth uses it for two purposes: (1) it appears in the JIT approval prompt shown to the user (e.g. 'Read Things todo t1'), and (2) it's logged in the audit trail. Without it, the user would see a generic 'allow scope things:read?' prompt with no context about what operation is being authorized.

README.md
aidanns: Claude: Done! Removed --host, --port, --auth-url flags from the things-bridge CLI. The README now points to `~/.config/things-bridge/config.json` for customization.

README.md
aidanns: Claude: Agreed. Updated README to mention interactive prompting and config-file pre-population as alternatives. Tracked the full interactive login implementation in TODO.md under 'Interactive credential input for CLI login'.

README.md
aidanns: Claude: Updated the README to add an inline comment: `# Things built-in list id`. `TMTodayListSource` is Things 3's internal identifier for the 'Today' smart list. Other examples include `TMInboxListSource`, `TMUpcomingListSource`, `TMAnytimeListSource`. These are the AppleScript-level list identifiers that Things uses internally.

README.md
aidanns: Claude: Done! `select_store('auto')` now falls back to `FileStore` when no working keyring backend is detected, instead of raising an error. The README is updated to describe this behavior.

src/things_bridge/cli.py
aidanns: Claude: Done! Removed `--config-dir` from the things-bridge CLI entirely.

src/things_bridge/cli.py
aidanns: Claude: Done! Removed all CLI flag overrides (`--host`, `--port`, `--auth-url`, `--config-dir`) from the things-bridge CLI. Configuration is now purely via `config.json`.

src/things_bridge/config.py
aidanns: Claude: Tracked in TODO.md under 'Shared configuration library between services'. The XDG config loading pattern is similar between agent-auth and things-bridge. Will be worth extracting if a third service appears — possibly bundled with the AgentAuthClient extraction (#10).

src/things_bridge/config.py
aidanns: Claude: Done! Removed `config_dir` from the `Config` dataclass. `load_config()` no longer accepts a directory argument — it always uses the XDG default.

src/things_bridge/config.py
aidanns: Claude: Done! Renamed the default log file from `bridge.log` to `things-bridge.log`.

src/things_bridge/models.py
aidanns: Claude: Tracked in TODO.md under 'Semantic types for model identifiers'. Plan to introduce `typing.NewType` wrappers (e.g. `TodoId`, `ProjectId`, `AreaId`) so type checkers can catch id-type mismatches.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under 'Per-data-type authorization scopes'. A granular scope model (`things:read:todos`, `things:read:projects`, `things:read:areas`) would allow more narrowly-scoped tokens — recorded as a design decision to address in DESIGN.md.

src/things_bridge/server.py
aidanns: Claude: Done! Rewrote `_safe_id` to use an allow-list: printable ASCII (0x20–0x7E) excluding `/`, plus non-ASCII (> 0x7F). Everything else (C0, DEL, slash) is rejected.

src/things_bridge/server.py
aidanns: Claude: Done! Removed 'read-only' from the handler class docstring.

src/things_bridge/server.py
aidanns: Claude: `log_request` is called by `BaseHTTPRequestHandler.send_response()` — it's the stdlib's built-in access log hook. Every time we call `self.send_response(200)` etc., the base class calls `self.log_request(code, size)` which by default writes to stderr. We override it to suppress access logs because request paths can reveal Things ids and bearer tokens appear in headers.

src/things_bridge/server.py
aidanns: Claude: Done! Restructured `_validate` so `return True` is inside the `try` block immediately after the successful `authz.validate()` call, and `return False` is at the bottom after all `except` branches. This makes it clear that `True` is only returned when `validate()` succeeds normally.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under 'Deterministic authorization validation on every endpoint'. Good options: (a) move validation into a middleware layer that runs before routing — then routes only execute post-auth, (b) add an integration test that enumerates all known paths and asserts each returns 401 without a token. Option (b) is achievable now; option (a) would require moving away from BaseHTTPRequestHandler's manual routing.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under 'Query parameter deduplication semantics'. For REST APIs, repeated query params typically mean OR/multi-value (`?tag=A&tag=B` → items tagged A or B). The current `_first()` approach silently drops duplicates. The design should specify the intended semantics — I've recorded it as a design decision to resolve in DESIGN.md.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under 'Audit logging for every HTTP endpoint'. The bridge currently only logs errors (via the default `log_error`). Adding structured audit logging (method, path, auth result, status code) to the service log would support both operational debugging and security review.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under '`_safe_id` return type as a distinct type'. A `NewType('SafeId', str)` return would make the validated-vs-raw distinction visible to type checkers, preventing callers from accidentally using an unvalidated string where a safe id is expected.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under '404 vs 400 for unparseable path-segment ids'. The current 404 avoids leaking validation detail (the client can't distinguish 'bad id syntax' from 'id not found'), but semantically a 400 would be more accurate since the request itself is malformed. This is an API design decision to document in DESIGN.md.

src/things_bridge/server.py
aidanns: Claude: Done! Renamed `_things_error_response` to `_send_things_error_response` to match the verb-name convention.

src/things_bridge/server.py
aidanns: Claude: The scope is specified — `_validate` always passes `READ_SCOPE` (`things:read`) to `authz.validate()`. If you mean per-endpoint granular scopes (e.g. `things:read:todos` vs `things:read:projects`), that's tracked in TODO.md under 'Per-data-type authorization scopes' as a design decision.

src/things_bridge/server.py
aidanns: Claude: Currently when `_safe_id` rejects an id, we return 404 before checking auth. Returning 401 would require checking the bearer token first, even for syntactically invalid requests. The current behavior avoids running auth validation for garbage input, but the trade-off is that an unauthenticated caller can distinguish 'bad id' (404) from 'valid id, no auth' (401). Tracked in TODO.md under '404 vs 400 for unparseable path-segment ids' and 'Deterministic authorization validation on every endpoint' — both design decisions that should be resolved together.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under 'HEAD and OPTIONS implementation'. HEAD on the read endpoints would be useful for health checks and client-side caching (check if data changed without transferring the body). Currently blocked on the design decision of whether the bridge should support it.

src/things_bridge/server.py
aidanns: Claude: Tracked in TODO.md under 'HEAD and OPTIONS implementation'. OPTIONS would be needed for CORS preflight if the bridge is ever accessed from a browser context. For localhost-only server-to-server use it's not needed today, but the design should document the decision explicitly.

src/things_bridge/things.py
aidanns: Claude: Done! Renamed `ThingsClient` to `ThingsApplescriptClient` across all source files and tests.

src/things_cli/cli.py
aidanns: Claude: The `family_id` is needed for the token reissue flow. When a refresh token expires, the client calls `/agent-auth/token/reissue` with the `family_id` to request a new token pair (which blocks on JIT host-side approval). Without it, the user would have to run `things-cli login` again to get new tokens manually. It's optional at login time — if omitted, reissue is unavailable and refresh-token expiry requires manual re-login.

src/things_cli/cli.py
aidanns: Claude: The `int` return is the process exit code. `main()` returns it, and the `__main__` block passes it to `sys.exit()`. Different values indicate different error categories: 0=success, 1=usage error, 2=credentials/auth error, 3=scope denied, 4=not found, 5=bridge unavailable. This lets shell scripts branch on the exit code.

src/things_cli/cli.py
aidanns: Claude: Because the return value is a process exit code, not a success/failure flag. Different handlers return different non-zero values (2, 3, 4, 5) to indicate specific error categories (auth failure, scope denied, not found, bridge unavailable). A bool would collapse these into a single failure state.

src/things_cli/cli.py
aidanns: Claude: Done! `BridgeClient` now exposes typed methods (`list_todos`, `get_todo`, `list_projects`, `get_project`, `list_areas`, `get_area`) that encapsulate the URL paths and URL-encoding. The CLI handlers now call e.g. `client.get_todo(args.id)` instead of building the path themselves.

src/things_cli/client.py
aidanns: Claude: Same convention as in `AgentAuthClient` — the `*` forces `timeout_seconds` to be keyword-only (PEP 3102), preventing `BridgeClient(creds, store, 5.0)` in favor of the explicit `BridgeClient(creds, store, timeout_seconds=5.0)`.

src/things_cli/client.py
aidanns: Claude: Done! The generic `get(path)` method is now private (`_request`). The public API is typed methods: `list_todos(params=...)`, `get_todo(todo_id)`, `list_projects(params=...)`, `get_project(project_id)`, `list_areas()`, `get_area(area_id)`. Each method builds the path internally and handles URL-encoding.

src/things_cli/client.py
aidanns: Claude: Done! Renamed `_auth_headers` to `_bearer_headers` — it builds the full request headers dict (Authorization + Accept), not just an auth header.

src/things_cli/client.py
aidanns: Claude: The outer `try/finally` ensures `conn.close()` runs regardless of what happens. The inner `try/except` catches network errors (`ConnectionError`, `TimeoutError`, `OSError`) and wraps them in `BridgeUnavailableError`. They serve different purposes: the inner translates exceptions, the outer guarantees cleanup. Collapsing them into one block would mean the `except` clause runs before `conn.close()` but the `finally` would still work — however, keeping them separate makes the intent clearer.

src/things_cli/credentials.py
aidanns: Claude: Yes — this is the atomic write pattern. We write to a temp file first, then `os.replace()` to the final path. This prevents a crash mid-write from leaving a corrupt credentials file. If the write fails (disk full, permission error), the temp file is cleaned up and the existing credentials file is untouched.

src/things_cli/credentials.py
aidanns: Claude: The 0600 mode is already set at creation time via `os.open(tmp_path, O_WRONLY|O_CREAT|O_TRUNC, 0o600)` — this ensures the file is never world-readable, even during the write. The subsequent `os.replace()` preserves the mode. Added a comment to make this explicit.

src/things_cli/credentials.py
aidanns: Claude: Done! Added a post-write verification: after `os.replace()`, we stat the file and raise `CredentialsBackendError` if the mode isn't 0600. This catches cases where e.g. a restrictive umask or filesystem ACL overrides the requested mode.

src/things_cli/credentials.py
aidanns: Claude: Done! `select_store('auto')` now falls back to `FileStore` when no working keyring backend is available, instead of raising an error.

src/things_cli/credentials.py
aidanns: Claude: Tracked in TODO.md under 'macOS Keychain per-application access control'. macOS Keychain supports per-application ACLs via `SecAccessControl`, which could restrict things-cli's keychain entries to only be readable by the things-cli binary. However, the Python `keyring` library doesn't expose this API — implementing it would require either PyObjC or a custom keyring backend. For now, credentials are protected by user-level keychain access (any process running as the same user can read them, same as the 0600 file fallback).

README.md
aidanns: Claude: Agreed. The auto-fallback to file store only applies to things-cli's `select_store('auto')`. The things-bridge server doesn't store credentials — it validates tokens via AgentAuthClient against agent-auth. agent-auth itself stores its signing keys in the system keyring and should fail to start if the keyring is unavailable (its key management is in `agent_auth.keys`, not changed in this PR).

README.md
aidanns: Claude: Updated the README to describe interactive prompting and config-file pre-population as the credential input methods. The full interactive implementation (like `gh auth login` flow) is tracked in TODO.md under 'Interactive credential input for CLI login'.

src/things_bridge/server.py
aidanns: add to TODO.md that we should audit all http requests to audit.log

src/things_bridge/authz.py
aidanns: this should be encoded in configuration on the agent-auth service - if we allow the client to pass through this data, there is nothing stopping them from mis-representing the scope that they are requesting in the prompt

src/things_cli/cli.py
aidanns: why can't we request a re-issue with the expired refresh token instead?

src/things_cli/cli.py
aidanns: make it semantic - EXIT_SUCCESS, etc

src/things_cli/credentials.py
aidanns: we should also fail when reading the credentials if the file doesn't have the right permissions set - similar to ssh

README.md
```diff
@@ -90,30 +90,29 @@ curl -H "Authorization: Bearer aa_<id>_<sig>" \
 ```bash
 # Start the bridge (default: 127.0.0.1:9200)
 things-bridge serve
-
-# Customize bindings and point at a non-default agent-auth URL
-things-bridge serve --host 127.0.0.1 --port 9200 --auth-url http://127.0.0.1:9100
 ```
 
+Host, port, and agent-auth URL are configured in `~/.config/things-bridge/config.json`.
```
aidanns: make this a yaml file.

README.md
```diff
@@ -90,30 +90,29 @@ curl -H "Authorization: Bearer aa_<id>_<sig>" \
...
+
 ### things-cli
 
-`things-cli` is a thin client for `things-bridge` that auto-refreshes/reissues tokens via `agent-auth`. Credentials are kept in the system keyring by default; pass `--credential-store=file --credentials-file=<path>` to persist them in a 0600 JSON file (useful inside a devcontainer).
+`things-cli` is a thin client for `things-bridge` that auto-refreshes/reissues tokens via `agent-auth`. Credentials are kept in the system keyring by default; when no keyring backend is available (e.g. inside a devcontainer), the CLI automatically falls back to a `0600` JSON file at `~/.config/things-cli/credentials.json`.
```
aidanns: make this a yaml file.

README.md
aidanns: Claude: Done! Switched to YAML — `config.json` → `config.yaml` using PyYAML's `safe_load`.

README.md
aidanns: Claude: Done! Switched to YAML — `credentials.json` → `credentials.yaml` using PyYAML's `safe_dump`/`safe_load`. Also updated DESIGN.md reference.

PR #52, opened 2026-04-17
chore: name venv dir by OS/arch to allow shared filesystem usage

aidanns:
## Summary

- Rename the project virtualenv from `.venv/` to `.venv-$(uname -s)-$(uname -m)/` (e.g. `.venv-Linux-aarch64/`, `.venv-Darwin-arm64/`) so the same repo checkout on a shared filesystem (e.g. macOS host bind-mounted into a Linux devcontainer) can maintain per-OS/arch venvs without clobbering each other.
- Update `.gitignore`, `README.md`, `CLAUDE.md`, `.claude/instructions/python.md`, and the four bootstrap scripts (`scripts/test.sh`, `scripts/agent-auth.sh`, `scripts/things-bridge.sh`, `scripts/things-cli.sh`) to use the new pattern.
- Mirrors the convention already used by `~/Projects/systems-engineering`.

## Test plan

- [x] `scripts/test.sh` bootstraps a fresh `.venv-$(uname -s)-$(uname -m)/` and runs the full pytest suite green on Linux (163 passed)
- [x] `grep -rn '\.venv[^-*]' .` returns no stale `.venv` references in tracked files
- [x] `.gitignore` still matches both the legacy `.venv/` and the new per-OS/arch dirs

### DIFF
--- a/.claude/instructions/python.md
+++ b/.claude/instructions/python.md
@@ -20,7 +20,9 @@ general coding standards in `coding-standards.md`.
 
 - **`ruff`** — linting and formatting.
 - **`mypy` and `pyright`** — type checking. Run both in CI.
-- **`uv`** — virtual environment and dependency resolution. Creates `.venv/`
-  in the project directory per the global convention.
+- **`uv`** — virtual environment and dependency resolution. Point it at
+  `.venv-$(uname -s)-$(uname -m)/` in the project directory (e.g. by
+  exporting `UV_PROJECT_ENVIRONMENT`) so macOS and Linux venvs can coexist
+  on a shared filesystem.
 - **`pytest-cov`** — line and branch coverage with a ratcheting threshold.
 - **`pip-audit`** (or `safety`) — dependency vulnerability scanning in CI.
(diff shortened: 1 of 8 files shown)


aidanns (commented):
Claude Reviewer: The `.venv*/` pattern in `.gitignore` correctly matches both the legacy `.venv/` and the new `.venv-Linux-aarch64/` / `.venv-Darwin-arm64/` directories, but it also matches any unrelated directory starting with `.venv` (e.g. a user-created `.venv-backup/`, or hypothetically `.venvrc/`). A slightly more precise pattern would be two lines — `.venv/` and `.venv-*/` — which matches exactly the legacy and per-OS/arch forms this PR introduces and nothing else. Minor nit; feel free to ignore if the broader glob is intentional.

aidanns:
Claude: Keeping the `.venv*/` pattern as-is — it matches the convention used in `~/Projects/systems-engineering`, which is the reference the user explicitly asked me to copy. The over-match is benign in practice (anything starting with `.venv` is almost certainly venv-related), and keeping parity with systems-engineering is more valuable than marginal precision.

PR #53, opened 2026-04-18
feat: add client-level FakeThingsClient for Linux e2e testing

aidanns:
## Summary

- Adds a `ThingsClient` Protocol covering the six read methods `ThingsBridgeServer` consumes, with two implementations: `ThingsApplescriptClient` (production) and a new `FakeThingsClient` (in-memory, production code in `src/things_bridge/fake.py`).
- `things-bridge serve --fake-things[=PATH]` swaps in the fake and prints a loud stderr banner. Intended for Linux devcontainers and CI — no config-file surface, developer-only.
- The fake applies the same filter semantics the real client does (`project_id`, `area_id`, `tag`, `status`). `list_id` is resolved via a per-fixture `list_memberships` sidecar since Things' built-in smart lists can't be reproduced faithfully from dataclass fields alone.
- `tests/test_things_bridge_e2e.py` exercises the full agent-auth + things-bridge HTTP stack against the fake (token validation, scope enforcement, revocation, expired tokens, free-form notes round-trip, filter forwarding, authz unavailability).
- Runner-level faking was considered and rejected — there is no open-source AppleScript interpreter for Linux (BushelScript is macOS-only because Apple Events are), so that path collapses to hand-rolled regex matching on script strings. The client-level choice is recorded in `design/decisions/0001-things-client-fake.md`.
- Real-AppleScript end-to-end coverage on macOS runners is deferred; a follow-up issue will capture it.

## Test plan

- [x] `pytest tests/test_things_bridge_fake.py tests/test_things_bridge_e2e.py` — 31 tests pass
- [x] `pytest` (full suite) — 194 tests pass
- [x] `scripts/verify-design.sh` — all functions allocated
- [x] `scripts/verify-function-tests.sh` — `Serve Fake Things Client` covered by `@covers_function` markers
- [ ] Manual full-stack smoke in a Linux devcontainer (blocked on this Linux host: no `keyring` backend, so `agent-auth serve` can't materialise the signing key; orthogonal gap — tracked separately). The equivalent coverage is exercised by `test_things_bridge_e2e.py`, which spins up a real in-process `AgentAuthServer` + `ThingsBridgeServer` with a materialised signing/encryption key and drives real HTTP against the bridge.


plans/things-client-fake.md
```diff
@@ -0,0 +1,254 @@
...
+
+`things-bridge serve --fake-things[=<path>]` in `src/things_bridge/cli.py`:
+
+- Flag absent -> current behaviour (real `AppleScriptRunner` +
+  `ThingsApplescriptClient`).
+- Flag present, no value -> `FakeThingsClient(FakeThingsStore())` (empty
+  store).
+- Flag present, value -> `load_fake_store(path)` → `FakeThingsClient(...)`.
+
+Also accept env var `THINGS_BRIDGE_FAKE_THINGS` as a fallback (convenient
```
aidanns: remove this, just stick with the CLI flag.

plans/things-client-fake.md
```diff
@@ -0,0 +1,254 @@
...
+Flag name is `--fake-things` (not `--fake-applescript`) because the fake
+is at the Things-client layer; the AppleScript layer is bypassed entirely.
+
+### End-to-end test
+
+New file `tests/test_things_bridge_e2e.py`:
+
+- Pytest fixture starts `ThingsBridgeServer(config, client, authz)` where
+  `client = FakeThingsClient(store_from_fixture)` and `authz` is a
+  `FakeAuthz` that accepts any token.
```
aidanns: why are we accepting any token here? I think we want to be properly testing the interaction with agent-auth in this e2e test.

plans/things-client-fake.md
```diff
@@ -0,0 +1,254 @@
...
+  notes".
+- `design/DESIGN.md` — note the test-only fake and its scope.
+- `design/functional_decomposition.yaml` / `.md` / `.d2` — add a leaf
+  under `things-bridge` for `Fake Things Client` so function-to-test
+  traceability holds.
+- `tests/conftest.py` — shared `FakeAuthz` fixture if not already shared.
+
+**GitHub issue to open (separately from the PR):**
+
+- "e2e tests with real Things 3 via GitHub Actions macOS runner" — body
```
aidanns: add an additional github issue to remove as much logic from the applescript as possible and move it in to Python

plans/things-client-fake.md
aidanns: Claude: Done! Dropped the env var; CLI flag is the only way to enable the fake.

plans/things-client-fake.md
aidanns: Claude: Agreed. The e2e test now spins up a real `AgentAuthServer` in-process, creates a real `things:read=allow` token family via the existing `_create_test_tokens` helper, and points `things-bridge` at it via the real `AgentAuthClient`. Coverage now includes the authz failure modes (401 on missing/expired/revoked tokens, 403 on wrong scope, 502 when agent-auth is unreachable). The unit tests in `tests/test_things_bridge_server.py` keep using `FakeAuthz` — they shouldn't depend on agent-auth internals, and the e2e is where the real interaction belongs.

plans/things-client-fake.md
aidanns: Claude: Done! Added Issue 2 to the "Follow-up GitHub issues" section: "minimise AppleScript logic in things-bridge, push filtering and shaping into Python". Covers pushing `whose` filters, TSV framing, ISO-date parsing, and tag-splitting into Python on parsed data — shrinks injection surface, makes those paths trivially unit-testable, and reduces lock-step coupling if we ever revisit a runner-level fake.

src/things_bridge/fake.py
```diff
@@ -0,0 +1,209 @@
...
+            raise ThingsNotFoundError(f"area {area_id!r} not found")
+        return area
+
+
+def _build_todo(data: dict[str, Any]) -> Todo:
+    return Todo(
+        id=data["id"],
+        name=data.get("name", ""),
+        notes=data.get("notes", ""),
+        status=data.get("status", "open"),
```
aidanns: Claude Reviewer: Fixture builders don't validate `status`, so a fixture with `status: done` (typo or unknown value) silently loads and then the `?status=` query filter never matches that todo — a confusing failure mode for anyone iterating on fixtures. The fake validates the *query* via `validate_status(status)` above but not its own fixture data. Consider calling `validate_status(data.get("status", "open"))` inside `_build_todo` / `_build_project` so bad fixtures fail loudly at load time, matching the strictness the real client enforces end-to-end.

src/things_bridge/fake.py
```diff
@@ -0,0 +1,209 @@
...
+        data = yaml.safe_load(f) or {}
+
+    areas = [_build_area(a) for a in data.get("areas", []) or []]
+    projects = [_build_project(p) for p in data.get("projects", []) or []]
+    todos = [_build_todo(t) for t in data.get("todos", []) or []]
+
+    memberships_raw = data.get("list_memberships", {}) or {}
+    list_memberships = {
+        list_id: set(todo_ids or []) for list_id, todo_ids in memberships_raw.items()
+    }
```
aidanns, line 206: Claude Reviewer: `load_fake_store` makes no assumptions about fixture shape and will silently accept several classes of mistake that are awkward to debug later:

- Top-level YAML that isn't a mapping (e.g. starts with a list) will crash with a non-obvious `AttributeError` on `.get` — prefer an explicit `if not isinstance(data, dict): raise ValueError(...)`.
- Unknown top-level keys (e.g. `todo:` instead of `todos:`) are silently dropped, so a typo results in an empty store.
- `tag_names: Errand` (scalar rather than list) passes through as `list("Errand")` → `['E','r','r','a','n','d']`.
- Duplicate ids across `todos`/`projects`/`areas` silently last-write-wins in the `*_by_id` properties with no warning.

Given fixtures are developer-authored and the fake exists specifically to support e2e debugging, a few `isinstance` / duplicate-id / unknown-key checks here would save real time vs. surface behaviour that looks like a bridge bug.

src/things_bridge/fake.py
```diff
@@ -0,0 +1,209 @@
...
+          - id: t1
+            name: Buy milk
+            status: open
+            project_id: p1
+            project_name: Q2 Planning
+            tag_names: [Errand]
+        list_memberships:
+          TMTodayListSource: [t1]
+    """
+    with open(path) as f:
```
aidanns: Claude Reviewer: `open(path)` here relies on the platform default encoding (locale-dependent, and notably cp1252 on some Windows devcontainers / CI). Since fixtures can legitimately contain non-ASCII characters (todo names, notes, tags) and the example file ships with a `—` elsewhere in the repo, be explicit: `open(path, encoding="utf-8")`. The unit tests pass `encoding="utf-8"` on the writer side already, which makes the reader's silence here more jarring.

tests/test_things_bridge_fake.py
```diff
@@ -0,0 +1,188 @@
...
+        _todo(id="t1", status="open"),
+        _todo(id="t2", status="completed"),
+        _todo(id="t3", status="canceled"),
+    ])
+    client = FakeThingsClient(store)
+    assert [t.id for t in client.list_todos(status="completed")] == ["t2"]
+
+
+def test_list_todos_rejects_invalid_status():
+    # Matches ThingsApplescriptClient._validate_status so bridge error mapping is uniform.
```
aidanns: Claude Reviewer: Stale identifier in the comment — it references `ThingsApplescriptClient._validate_status`, but this PR promoted that to module-level `things_bridge.things.validate_status`. Suggest: `# Matches things_bridge.things.validate_status so bridge error mapping is uniform.`

tests/test_things_bridge_server.py
```diff
@@ -37,56 +40,58 @@ def validate(self, token, required_scope, *, description=None):
...
-        return list(self.areas)
+        return self._client.list_areas()
 
     def get_area(self, area_id):
-        if area_id not in self.areas_by_id:
-            raise ThingsNotFoundError(area_id)
-        return self.areas_by_id[area_id]
+        if self.raise_on_call is not None:
+            raise self.raise_on_call
+        return self._client.get_area(area_id)
```
aidanns, line 86: Claude Reviewer: This wrapper duplicates every method of `ThingsClient` by hand. When the Protocol grows (write / JIT-gated endpoints are on the roadmap per `design/DESIGN.md`), this class will silently fall behind — a new protocol method added to both `FakeThingsClient` and `ThingsApplescriptClient` would bypass `raise_on_call` / `last_*_kwargs` spying entirely and tests would quietly stop exercising the error-mapping contract.

Two low-cost options:
1. Delegate uniformly via `__getattr__`, with a single guarded `_dispatch` that handles `raise_on_call` and spy-recording generically.
2. Make the Protocol `@runtime_checkable` and add a `typing.cast` or explicit conformance test that fails if the wrapper drifts.

Not blocking for this PR, but worth a `TODO(aidanns):` comment at minimum so the drift risk is documented.

tests/test_things_bridge_e2e.py
```diff
@@ -0,0 +1,324 @@
...
+
+    status, area = _get(f"{stack['bridge_url']}/things-bridge/areas/a1", token)
+    assert status == 200
+    assert area["area"]["id"] == "a1"
+
+
+def test_list_todos_authz_unavailable_returns_502(tmp_dir):
+    """If agent-auth isn't reachable, the bridge should report 502 authz_unavailable."""
+    # Point the bridge at a port nothing is listening on.
+    authz = AgentAuthClient("http://127.0.0.1:1", timeout_seconds=1.0)
```
aidanns: Claude Reviewer: Using port `1` as the "nothing listening" target is a portable-enough choice on Linux but isn't strictly guaranteed — `connect()` can block or fail with EACCES / EPERM on some kernels/containers before hitting ECONNREFUSED, which on very slow CI can leak into the 1s timeout path and produce 504-style timeouts instead of the 502 `authz_unavailable` this test asserts. A hardier pattern is to bind a socket with `socket.socket(); s.bind(("127.0.0.1", 0)); port = s.getsockname()[1]; s.close()`, then point at that port — it's guaranteed to refuse because the OS just freed it. Optional but defensive.

aidanns (commented):
Claude Reviewer: Solid change overall. The `ThingsClient` Protocol is a clean seam, the stderr banner in the CLI makes the fake hard to miss, and the new e2e suite meaningfully raises coverage of the authz/HTTP paths for Linux devcontainer use. The ADR clearly articulates why client-level was chosen over runner-level and records the real-Things-on-macOS follow-up.

Main themes in the inline comments:

1. `load_fake_store` / `_build_*` are permissive in ways that will bite developers authoring fixtures: invalid status silently accepted, scalar `tag_names` expanded character-by-character, unknown keys silently dropped, missing `encoding="utf-8"`. None are bugs today but all are papercuts waiting to happen.
2. `_InjectableThings` in `tests/test_things_bridge_server.py` duplicates the Protocol surface by hand — at minimum worth a TODO flagging the drift risk once write endpoints land.
3. Stale comment reference in the fake unit test (`_validate_status` -> `validate_status`).
4. The `http://127.0.0.1:1` trick in the authz-unavailable test is a minor portability wart.

No blockers. Verdict: COMMENT.

src/things_bridge/fake.py
aidanns: Claude: Done! `_validate_fixture_status` now runs in `_shared_item_kwargs` so any `status:` value in a fixture is checked against `VALID_STATUSES` at load time and surfaces a `ThingsError` with the same wording the query-param validator produces. Regression test: `test_load_fake_store_rejects_invalid_status`.

src/things_bridge/fake.py
aidanns: Claude: Done! `load_fake_store` now (a) rejects a non-mapping root with a `ThingsError` before any `.get` touches it, (b) rejects unknown top-level keys against a `_ALLOWED_TOP_LEVEL_KEYS` allow-list (so `todoss:` fails loud instead of silently loading as empty), and (c) rejects scalar `tag_names` so `tag_names: Errand` does not silently split into `["E","r","r","a","n","d"]`. Regression tests: `test_load_fake_store_rejects_non_mapping_root`, `test_load_fake_store_rejects_unknown_top_level_key`, `test_load_fake_store_rejects_scalar_tag_names`.

src/things_bridge/fake.py
aidanns: Claude: Done! `open(path, encoding="utf-8")`.

tests/test_things_bridge_fake.py
aidanns: Claude: Done! Comment now reads: `# Shares validate_status with ThingsApplescriptClient so bridge error mapping is uniform.`

tests/test_things_bridge_server.py
aidanns: Claude: Holding off for now. The drift risk is real, but the read-only Protocol is only six methods and the JIT-gated write endpoints are a separate workstream that will land with its own tests. When we add a new Protocol method, the test-suite collapse is loud (AttributeError at call time, not silent pass), so we will catch it at PR time. I would rather revisit this when the Protocol actually grows — either by replacing `_InjectableThings` with a `FakeThingsClient` subclass that overrides only the spy hooks, or by switching to `unittest.mock.Mock(spec=ThingsClient)` so drift becomes a type error. Filed a note in the follow-up AppleScript-minimisation issue to revisit this when we touch the class next.

tests/test_things_bridge_e2e.py
aidanns: Claude: Done! Replaced the port-1 target with a bind-then-close ephemeral-port probe: we bind `127.0.0.1:0`, read `getsockname()[1]`, close, and point the bridge at the (now guaranteed-unused, non-privileged) port. No EACCES/EPERM foot-gun, and the connect fails fast enough to stay inside the 1s timeout.

PR #54, opened 2026-04-18
fix(things-bridge): repair AppleScript payload + log osascript failures

aidanns:
## Summary

- Fix two AppleScript bugs that caused every `list` / `show` bridge endpoint to return an opaque `502 things_unavailable`: `_HELPERS` used unsupported `\uXXXX` escapes (replaced with `character id N`), and `_statusText` referenced bare Things3 keywords (`open`/`completed`/`canceled`) outside their enclosing `tell application` block (replaced with text coercion and string comparison).
- Surface raw osascript stderr to the bridge process's own stderr on non-zero exit so operators have a diagnostic channel. Response bodies stay sparse to preserve the host-info leak guarantee that `design/DESIGN.md` already documents.
- Add three darwin-only regression tests: an `osacompile`-based syntax check of `_HELPERS`, an end-to-end `list_projects` exercise against real Things 3, and a `capfd` assertion that osascript failures reach the bridge's stderr. Existing `FakeRunner`-based tests never touched the compile-and-execute path, which is why these bugs reached production.

## Test plan

- [x] `.venv-Darwin-arm64/bin/pytest -q` — full suite green on macOS (including the three new darwin-only tests). 166 passed.
- [x] `bash scripts/verify-function-tests.sh` — "Execute External System Interaction" now allocated. Coverage rose 37/47 → 38/47.
- [x] Manually restart `things-bridge` and confirm `scripts/things-cli.sh todos list` returns todos instead of `502 things_unavailable`. Verified in the debugging session that produced this PR.
- [ ] Manually trigger an osascript failure (e.g. revoke Automation permission) and confirm the bridge's stderr carries the `things-bridge: osascript failed` line while the HTTP response body stays sparse. The automated `test_osascript_failure_writes_diagnostic_to_stderr` exercises the same failure path deterministically with invalid AppleScript; the permission-revoke variant is left as a manual follow-up.

### DIFF
--- a/src/things_bridge/things.py
+++ b/src/things_bridge/things.py
@@ -34,14 +34,14 @@ on _esc(v)
     if v is missing value then return "missing value"
     set s to v as text
     set out to ""
-    set tabPh to (character id 9246)
-    set lfPh to (character id 9247)
+    set tabPlaceholder to (character id 9246)
+    set linefeedPlaceholder to (character id 9247)
     repeat with i from 1 to count of characters of s
         set c to character i of s
         if c is tab then
-            set out to out & tabPh
+            set out to out & tabPlaceholder
         else if c is return or c is linefeed then
-            set out to out & lfPh
+            set out to out & linefeedPlaceholder
         else
             set out to out & c
         end if
--- a/src/things_bridge/things.py
+++ b/src/things_bridge/things.py
@@ -10,6 +10,7 @@ un-escaped by the parser so free-form text survives the framing.
 """
 
 import subprocess
+import sys
 from dataclasses import dataclass
 
 from things_bridge.errors import ThingsError, ThingsNotFoundError, ThingsPermissionError
@@ -23,17 +24,24 @@ MISSING = "missing value"
 # Escapes tabs/newlines/carriage returns into unicode placeholders and
 # coerces missing value to the literal "missing value" so the parser
 # can treat it uniformly.
+#
+# ``character id N`` is the AppleScript way to produce a Unicode codepoint
+# inside a string. AppleScript string literals do NOT support ``\uXXXX``
+# escapes — attempting to use them produces a ``-2741`` syntax error at
+# compile time.
 _HELPERS = r"""
 on _esc(v)
     if v is missing value then return "missing value"
     set s to v as text
     set out to ""
+    set tabPh to (character id 9246)
+    set lfPh to (character id 9247)
     repeat with i from 1 to count of characters of s
         set c to character i of s
         if c is tab then
-            set out to out & "\u241e"
+            set out to out & tabPh
         else if c is return or c is linefeed then
-            set out to out & "\u241f"
+            set out to out & lfPh
         else
             set out to out & c
         end if
@@ -91,11 +99,16 @@ on _areaName(t)
 end _areaName
 
 on _statusText(s)
-    if s is open then
+    -- Handlers run outside the enclosing ``tell application "Things3"``
+    -- block, so bare Things3 keywords like ``open``/``completed``/``canceled``
+    -- are not in scope here. Coerce the status enum to its text name at
+    -- the handler boundary and compare as strings instead.
+    set stxt to (s as text)
+    if stxt is "open" then
         return "open"
-    else if s is completed then
+    else if stxt is "completed" then
         return "completed"
-    else if s is canceled then
+    else if stxt is "canceled" then
         return "canceled"
     else
         return "open"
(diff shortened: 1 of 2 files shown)


src/things_bridge/things.py
```diff
@@ -23,17 +24,24 @@
...
+# ``character id N`` is the AppleScript way to produce a Unicode codepoint
+# inside a string. AppleScript string literals do NOT support ``\uXXXX``
+# escapes — attempting to use them produces a ``-2741`` syntax error at
+# compile time.
 _HELPERS = r"""
 on _esc(v)
     if v is missing value then return "missing value"
     set s to v as text
     set out to ""
+    set tabPh to (character id 9246)
```
aidanns: expand this variable name to be easier to read

src/things_bridge/things.py
```diff
@@ -23,17 +24,24 @@
...
+    set lfPh to (character id 9247)
```
aidanns: expand this also

tests/test_things_bridge_things.py
```diff
@@ -1,21 +1,34 @@
...
     TAB_PLACEHOLDER,
+    AppleScriptRunner,
     ThingsApplescriptClient,
+    _HELPERS,
     _TODO_FIELDS,
     _PROJECT_FIELDS,
     _AREA_FIELDS,
 )
 
+_darwin_only = pytest.mark.skipif(
```
aidanns, line 29: add an additional check to skip tests if things isn't available (we might run on a MacOS host that allows some osascript checks, but doesn't actually have things available to interact with)

aidanns:
Claude Reviewer: Independent review of this PR.

The AppleScript bug fixes (character-id replacement for unsupported unicode escapes, text coercion for out-of-scope enum comparisons) are correct and well-explained in both the commit body and the in-script comments. The stderr diagnostic addition preserves the host-info leak guarantee documented in design/DESIGN.md line 71 while giving operators the diagnostic channel that doc already promises.

Two issues worth addressing before merge:

1. tests/test_things_bridge_things.py:283 (test_list_projects_executes_against_things) — the skip condition only checks sys.platform == darwin and osascript presence, not whether Things 3 is installed or Automation permission has been granted. Any macOS environment without Things 3 (fresh dev machine, macOS CI runner, contributor who hasn't installed the app) will raise ThingsError/ThingsPermissionError rather than skipping, so pytest -q goes red. Options: (a) gate it behind an opt-in env var like AGENT_AUTH_E2E_THINGS=1 or a custom marker that requires --run-e2e; (b) probe for Things 3 with 'osascript -e "id of application \"Things3\""' and add the result to the skipif. Option (a) matches the usual pattern for requires-real-external-system tests.

2. tests/test_things_bridge_things.py:287 (test_osascript_failure_writes_diagnostic_to_stderr) — the test deliberately uses invalid AppleScript and never contacts Things 3, but it is tagged covers_function(Execute External System Interaction). No external system interaction actually happens here; osascript exits with a syntax error before any Apple Event is sent. This overstates what the test covers and will mislead scripts/verify-function-tests.sh. I'd drop the covers_function marker on this one — the adjacent test_list_projects_executes_against_things already genuinely covers that leaf function.

Note: there is a stale pending review on this PR (review id PRR_kwDOSAopLs72Zc5F, authored by aidanns) that blocked me from creating a new review via gh pr review. I've posted this as a top-level PR comment instead. You may want to dismiss or submit that pending review.

src/things_bridge/things.py
aidanns: Claude: Done! Renamed to `tabPlaceholder` in df6bc93.

src/things_bridge/things.py
aidanns: Claude: Done! Renamed to `linefeedPlaceholder` in df6bc93.

tests/test_things_bridge_things.py
aidanns: Claude: Done in df6bc93. Added a `_requires_things3` marker that also checks for the `Things3.app` bundle under `/Applications` or `~/Applications`, and applied it to `test_list_projects_executes_against_things` (the only new test that actually talks to Things 3). The `osacompile`-based syntax check and the stderr-diagnostic test only need `osascript`, so they stay on the lighter `_darwin_only` gate.

aidanns:
Claude: Addressed both findings in df6bc93:

1. `test_list_projects_executes_against_things` now uses a new `_requires_things3` skip marker that additionally probes for the `Things3.app` bundle under `/Applications` and `~/Applications`. A macOS host without Things 3 installed will now skip cleanly rather than raising `ThingsError`. The other two new tests only need `osascript`, so they stay on `_darwin_only`.
2. Dropped the `covers_function("Execute External System Interaction")` marker from `test_osascript_failure_writes_diagnostic_to_stderr` — you are right that osascript rejects the deliberately-invalid script before any Apple Event is sent, so no external system interaction actually occurs. `test_list_projects_executes_against_things` still carries the marker and `scripts/verify-function-tests.sh` still shows 38/47 leaf functions covered.

PR #57, opened 2026-04-18
fix(things-bridge): log osascript timeouts to stderr

### LINKED ISSUE #55: Log osascript timeout failures to bridge stderr
## Problem

`AppleScriptRunner.run` (`src/things_bridge/things.py`) catches `subprocess.TimeoutExpired` and re-raises as plain `ThingsError`, which the server maps to `502 things_unavailable`. Clients see only that opaque code and the bridge's stderr is silent — the new `things-bridge: osascript failed (rc=...): ...` log line only fires on non-zero exit (the non-timeout branch).

This was discovered while debugging a live `todos list` failure: bridge stderr was empty, which we only knew meant "timeout" by process of elimination.

## Desired behaviour

On `subprocess.TimeoutExpired`, print a matching diagnostic line to the bridge's stderr before raising, e.g.:

```
things-bridge: osascript timed out after 30.0s
```

Same leak posture as the exit-code path — response body stays sparse, operator stderr gets the detail.

## Acceptance

- `AppleScriptRunner.run` logs to stderr in the timeout branch.
- Regression test analogous to `test_osascript_failure_writes_diagnostic_to_stderr` that provokes a `TimeoutExpired` (e.g. via a 0-second timeout runner instance against a deliberately-slow script) and asserts both `things-bridge` and `timed out` are in stderr.
- Related: #56 (the root-cause performance issue that makes this diagnostic hit in the first place).

aidanns:
## Summary

- `AppleScriptRunner.run` now emits a `things-bridge: osascript timed out after {timeout}s` diagnostic to the bridge's stderr before re-raising `ThingsError` on `subprocess.TimeoutExpired`, mirroring the existing exit-code branch.
- HTTP response posture is unchanged: clients still receive the deliberately-sparse `502 things_unavailable`; only the operator stderr gets detail.
- Adds `test_osascript_timeout_writes_diagnostic_to_stderr` as a regression test alongside the existing exit-code stderr test.

Fixes #55

## Test plan

- [x] `python -m pytest tests/test_things_bridge_things.py -v` — new test passes, existing tests unaffected
- [x] `python -m pytest` — full suite green
- [x] Manual sanity: confirm the new test asserts both `things-bridge` and `timed out` appear on stderr


src/things_bridge/things.py
```diff
@@ -249,6 +249,15 @@ def run(self, script: str) -> str:
         except FileNotFoundError as exc:
             raise ThingsError(f"osascript not found at {self._osascript_path}") from exc
         except subprocess.TimeoutExpired as exc:
+            # Mirror the non-zero exit diagnostic: keep the HTTP response body
+            # sparse while giving operators enough detail on the bridge's own
+            # stderr to distinguish timeouts from other ``502 things_unavailable``
+            # causes.
+            print(
+                f"things-bridge: osascript timed out after {self._timeout}s",
```
aidanns: Claude Reviewer: The non-zero-exit branch below surfaces any `stderr` from osascript (`{stderr or '<empty stderr>'}`), but the timeout branch drops the partial output that `subprocess.TimeoutExpired` carries on `exc.stderr` / `exc.stdout` when `capture_output=True`. If osascript emitted a hint before being killed (e.g. a permissions prompt that never resolved), operators would currently see only the bare timeout line. Consider appending `exc.stderr` / `exc.stdout` when present to give the two branches symmetric operator-facing detail. Minor — feel free to defer if you'd rather keep the diff small.

aidanns (commented):
Claude Reviewer: Reviewed the PR diff against `origin/main` (production change + regression test only, 33 lines total).

**Correctness.** The timeout branch matches the existing exit-code branch's logging style: same `things-bridge:` prefix, `file=sys.stderr`, `flush=True`. One small asymmetry flagged inline — `TimeoutExpired` carries partial `stdout`/`stderr` when `capture_output=True` and the non-timeout branch surfaces that detail.

**Test quality.** `test_osascript_timeout_writes_diagnostic_to_stderr` exercises the public `AppleScriptRunner.run` surface, uses `capfd` (matching the sibling test), and its docstring focuses on operator impact ("cannot distinguish a stuck Things 3 from other AppleScript failures") rather than on the mechanics of the bug — which aligns with the user-global regression-test convention. Monkeypatching `subprocess.run` (rather than a darwin-gated real execution) is the right call since there is no portable way to force a real osascript timeout in CI; the test correctly runs on Linux and doesn't need the `@_darwin_only` marker that the sibling exit-code test requires.

**Conventions.** Nothing in `coding-standards.md`, `python.md`, or `service-design.md` is violated by this diff. The pre-existing `timeout` / `_timeout` attribute name technically elides the unit suffix (`coding-standards.md` -> "Units in names"), but that's not introduced by this PR and out of scope.

Overall the change is tight and does exactly what it says. Approving aside from the one inline nit.

src/things_bridge/things.py
aidanns: Claude: Done! Adopted the suggestion — timeout branch now appends `exc.stderr` in the same shape as the exit-code branch (`...: {stderr or '<empty stderr>'}`), and added `test_osascript_timeout_surfaces_partial_stderr` to lock in the partial-output case.

PR #58, opened 2026-04-18
perf(things-bridge): batch AppleScript property reads in list_todos

### LINKED ISSUE #56: Speed up list_todos to avoid 30s osascript timeout on real databases
## Problem

`ThingsApplescriptClient.list_todos` (`src/things_bridge/things.py`) drives a `repeat with t in (to dos)` loop that reads every property of every todo individually. On a real Things 3 database (observed: ~100 todos) the script takes >30s and hits the default `AppleScriptRunner` timeout, surfacing to clients as `502 things_unavailable`.

## Repro

1. Populate Things 3 with ~100+ todos.
2. `./scripts/things-cli.sh todos list`
3. Client: `Error: bridge unavailable: Bridge returned status 502: things_unavailable`.
4. Filtered variants (`--list TMTodayListSource`, `--project ...`, `--status open`) return fine because the cardinality is small. `projects list` and `areas list` also return fine.

## Options to investigate

1. **Batch property reads in AppleScript.** Collect the full todo list once, then fetch each property into a parallel list (`set ids to id of every to do of ...`, etc.). Avoids the per-todo round-trip cost inside the event loop.
2. **Switch to the things.py library.** Reads the Things 3 SQLite DB directly — orders of magnitude faster, but introduces a dependency and bypasses the AppleScript API contract (no Automation permission prompt, different failure modes).
3. **Bump the timeout.** Cheapest, doesn't fix the underlying O(N × per-property-round-trip) shape and will bite again on larger DBs.

Option 1 is the right first cut: same trust boundary as today, just less chatty scripting.

## Acceptance

- `todos list` against a 500-todo DB completes within the default timeout.
- Benchmark or regression test that guards against re-introducing per-todo round-trips (e.g. asserts the emitted script uses the batched form).
- Related diagnostic gap: #55

aidanns:
## Summary

- Rewrites the AppleScript emitted by `list_todos`, `list_projects`, and `list_areas` to read each property off the todo/project/area **collection** in a single Apple Event (`id of every to do of <scope>`, `name of every to do of <scope>`, ...) and zip the parallel lists back into the existing TSV row stream. Drops unfiltered `todos list` cost from roughly `15N` Apple Events to `~11 + N` — an order-of-magnitude reduction that keeps a 500-todo database under the `AppleScriptRunner` 30s timeout.
- `project` and `area` are optional relationships on a todo; the collection form raises if any element has no project/area attached, so those four fields stay on a per-element `try`-wrapped loop inside the same `tell` block. Documented in the ADR and exempted from the regression test's "no per-element reads" assertion.
- Adds regression tests that assert the batched AppleScript shape so a future refactor cannot silently reintroduce the round-trip cliff. Output schema is unchanged so no downstream parsers or HTTP responses change.

Fixes #56. Related: #55 (stderr logging for osascript timeouts) touches the same file but only the `AppleScriptRunner.run` timeout branch — expect at most a trivial rebase conflict at merge time.

## Test plan

- [x] `pytest` across the full project passes (168 existing tests green, 5 new regression tests added).
- [x] `pytest tests/test_things_bridge_things.py -q` on its own passes (unit-level confirmation).
- [x] New test `test_list_todos_reads_properties_from_collection_in_one_apple_event` asserts every batchable property uses `<property> of every to do` and rejects per-element reads of `id/name/notes/status/tag names/due date` in the unfiltered body.
- [x] New test `test_list_todos_empty_database_returns_no_todos` confirms the empty-collection short-circuit (`if _n is 0 then return ""`) round-trips through the parser to `[]`.
- [x] New test `test_list_todos_filtered_scope_uses_batched_reads` confirms each filter (`list_id`, `project_id`, `area_id`, `tag`) routes the batched reads through the caller-supplied scope rather than silently reading the whole database.
- [x] `test_helper_applescript_is_valid_syntax` (Darwin-only) still confirms `_HELPERS` compiles. *(Skipped on Linux — Darwin-only guard.)*
- [ ] Manual smoke test on a real Things 3 database: `scripts/things-cli.sh todos list` completes in under 30s (not runnable in CI — Linux host).


design/decisions/0001-list-todos-batched-applescript.md
```diff
@@ -0,0 +1,92 @@
+# ADR 0001: Batch AppleScript property reads in `list_todos`
```
aidanns: Claude Reviewer: ADR numbering collision. `main` already has `design/decisions/0001-things-client-fake.md` (added by #53), so this file introduces a second ADR with number `0001`. ADRs are conventionally numbered monotonically; two ADRs sharing `0001` makes cross-referencing ambiguous (`"see ADR 0001"` is now unresolvable) and breaks alphabetical ordering in tooling that walks the directory. Suggest renaming to `0002-list-todos-batched-applescript.md` and updating the header line from `# ADR 0001:` to `# ADR 0002:` to match.

aidanns (commented):
Claude Reviewer: Reviewed the diff against `main` (ignoring the auto-merge of #57's stderr logging). Overall this is a well-scoped performance fix and I'm comfortable with it — one structural concern posted inline (ADR filename collision with the existing `0001-things-client-fake.md`), and a handful of smaller positive observations below.

**Correctness of the AppleScript transform.** The `_every_form` helper is a pure prefix-swap on scope expressions that already flow through `_quote`, so the batched form inherits the existing injection guards without introducing new interpolation points. Scope plumbing is correct for all four filters (`list_id`, `project_id`, `area_id`, `tag`) — each ends up as the sole argument to every batched property read, so there is no path where a filtered call silently falls back to reading the entire database. The early `if _n is 0 then return ""` short-circuit before the remaining ten property reads is a nice touch (saves ~10 Apple Events on empty-scope queries like "todos of a tag with no matches") and it is exercised by the new empty-database test. The `project`/`area` per-element fallback is documented in both code and ADR and is correctly asserted-around in the regression test.

**Perf math check.** Old shape cost roughly `N + 14K` Apple Events where `K` was the number of status-matching todos (the inner `if` gated the 14 property reads). New shape is `~N + 11` regardless of status filter selectivity. For any `K > 0` the new form is faster, and at large `K` it's ~10x faster as the PR claims. Good.

**Injection safety.** `_quote` still gates every caller-supplied id/name/tag before it enters the AppleScript string; `_every_form` never touches the quoted literal itself; `validate_status` gates the status filter against a fixed allowlist before it reaches `_quote`. Existing injection tests pass unchanged. No new attack surface.

**Test quality.** The five new regression tests describe impact in their docstrings (timeout avoidance on large databases, filter scoping preventing silent whole-DB reads, empty-collection edge case) per the user's convention, and use the batched-form structural assertions as the regression guard. The assertion of `"id of every to do\\n"` with an explicit newline on the unfiltered test is particularly nice — catches the case where a refactor accidentally appends `" of (to dos)"` and reintroduces the plural form. I confirmed the full suite passes locally (204 passed, 3 skipped).

**ADR quality.** Content is good: captures the three options considered, the rejection rationale for each alternative (SQLite changes the trust boundary; raising the timeout hides the scaling issue), and the project/area caveat. Consequences section covers performance, output schema, trust boundary, and future work — this is what I'd want to read in six months. See the inline comment for the filename/number collision issue.

**Minor observations (not blocking, not worth posting as separate comments).**
- `_todo_batch_applescript` iterates the scope twice (once for the batched reads, once for the project/area fallback) then walks `_n` indices for row assembly. This is the natural shape given the "read all, then zip" pattern and I don't see a way to do it in one pass without either losing the batching win or restructuring the parser.
- The `forbidden_per_element_reads` tuple in `test_list_todos_reads_properties_from_collection_in_one_apple_event` only checks `"… of t)"` — a future regression that renames the scope loop variable to `_t` while reintroducing `id of _t)` wouldn't be caught. Narrow window, not worth hardening.
- Project/area cost is still O(N) Apple Events. The ADR's "Future work" section calls this out.

Approving modulo the ADR numbering fix.

design/decisions/0001-list-todos-batched-applescript.md
aidanns: Claude: Done! Renamed to `design/decisions/0002-list-todos-batched-applescript.md` and updated the header to `ADR 0002`. Grepped the repo for other references to the old path/number — none found.

PR #62, opened 2026-04-18
ci: enable Dependabot and verify required ecosystems

### LINKED ISSUE #44: Enable Dependabot for dependency updates
Source: `.claude/instructions/tooling-and-ci.md` — Security.

No automated dependency updates are configured. Standard requires Dependabot (or Renovate).

**Acceptance:**
- `.github/dependabot.yml` added covering `pip` (pyproject.toml) and `github-actions` (workflows).
- Weekly update cadence (or whatever cadence is chosen — document).
- Grouping configured so minor/patch updates across dependencies land in a single PR where sensible.

- `scripts/verify-standards.sh` asserts `.github/dependabot.yml` exists and declares both `pip` and `github-actions` ecosystems.

aidanns:
## Summary
- Add `.github/dependabot.yml` with weekly cadence for both `pip` and `github-actions`, grouping minor/patch bumps per ecosystem so routine updates land in a single PR while majors still arrive individually.
- Add `scripts/verify-standards.sh` — project-standards regression check. Asserts Taskfile task coverage and Dependabot ecosystem coverage (ecosystems derived from manifest presence) with grouping, using `yq` for structural YAML queries.
- Add `scripts/verify-dependencies.sh` — preflight asserting required local CLI tooling (`python3`, `systems-engineering`, `task`, `yq`) is on `PATH`.
- Wire both scripts into `Taskfile.yml` and switch every GitHub Actions workflow from calling `scripts/*.sh` directly to `task <name>`, with `task verify-dependencies` as a preflight step in each workflow.
- Install `d2` + `systems-engineering` + `go-task` in every workflow so the preflight succeeds uniformly.
- Note the new scripts in `CLAUDE.md` alongside the existing verifier references.

Closes #44.

## Test plan
- [x] `scripts/verify-standards.sh` exits 0 on a repo with the new config (positive case).
- [x] `scripts/verify-standards.sh` exits non-zero if the `pip` ecosystem is removed from `.github/dependabot.yml` (negative case).
- [x] `scripts/verify-standards.sh` exits non-zero if `.github/dependabot.yml` is missing.
- [x] `scripts/verify-standards.sh` exits non-zero if an ecosystem's minor/patch `groups` block is missing.
- [x] `scripts/verify-dependencies.sh` exits 0 when all required tools are on `PATH`.
- [x] `scripts/verify-dependencies.sh` exits non-zero and lists missing tools when one is stripped from `PATH`.
- [x] `task verify-standards` and `task verify-dependencies` dispatch to the respective scripts.
- [ ] GitHub Actions `Verify Standards`, `Verify Design`, `Verify Function Tests`, and `Test` workflows pass on this PR.
- [ ] Dependabot picks up the new config (check the "Dependency graph → Dependabot" tab in the repo settings after merge).


scripts/verify-standards.sh
```diff
@@ -0,0 +1,24 @@
...
+DEPENDABOT_CONFIG="${REPO_ROOT}/.github/dependabot.yml"
+
+if [[ ! -f "${DEPENDABOT_CONFIG}" ]]; then
+  echo "error: ${DEPENDABOT_CONFIG} is missing" >&2
+  exit 1
+fi
+
+for ecosystem in pip github-actions; do
+  if ! grep -Eq "^[[:space:]]*-[[:space:]]+package-ecosystem:[[:space:]]*\"?${ecosystem}\"?[[:space:]]*$" "${DEPENDABOT_CONFIG}"; then
+    echo "error: ${DEPENDABOT_CONFIG} does not declare the '${ecosystem}' ecosystem" >&2
```
aidanns: Claude Reviewer: The regex is brittle vs. valid YAML. It requires `- package-ecosystem: <name>` on its own line with nothing trailing, so it will silently fail on valid variants like a trailing comment (`- package-ecosystem: pip  # foo`), flow style (`- {package-ecosystem: pip, ...}`), single-quoted values, or different indentation. The risk is a false negative blocking CI on a legitimate reformat rather than missing a real problem. Consider shelling out to `python -c` with `yaml.safe_load` (PyYAML is already in the Python ecosystem) or `yq` so the assertion is structural rather than textual. Current config matches the regex — not a blocker.

scripts/verify-standards.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Verify that project-level standards required by .claude/instructions/
+# are in place. Currently asserts that Dependabot is configured for both
+# the pip and github-actions ecosystems (see tooling-and-ci.md Security).
```
aidanns: Claude Reviewer: Issue #44 only requires asserting both ecosystems are declared, which this does. But the issue also calls out grouping minor/patch as an acceptance criterion, and the verify script does not check that `groups` / `update-types` are present — a future edit that removes the grouping would pass CI. Consider extending the check (or opening a follow-up) so the regression guard matches the full AC, not just the ecosystem list.

.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
+version: 2
+
+# Weekly cadence is sufficient for a personal project: it keeps exposure
+# to known vulnerabilities bounded to a week while avoiding daily PR
+# churn. Minor and patch updates are grouped per ecosystem so routine
+# bumps land in a single PR; majors still come through individually so
+# they can be reviewed on their own merits.
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
```
aidanns, line 10: Claude Reviewer: `directory: "/"` for `pip` relies on Dependabot finding `pyproject.toml` at repo root — it does, so this should work. Worth confirming post-merge under Insights -> Dependency graph -> Dependabot that it actually picked up `pyproject.toml` rather than silently finding nothing; the PR test plan already lists this.

aidanns (commented):
Claude Reviewer: Independent review. Dependabot schema is valid, grouping syntax is correct, and acceptance criteria from #44 are met. A couple of substantive notes inline; no blockers.

scripts/verify-standards.sh
aidanns: Claude: Not adopting. The dependabot config is ours and change-controlled, so the textual-variant risks (trailing comments, flow style, single quotes, reindent) are theoretical. Switching to a structural parser means either `pip install pyyaml` in the workflow (extra step, fetch, and package-to-trust) or a separate `yq` install, which is disproportionate for a two-field assertion. The regex intentionally accepts optional double-quotes — the format we actually use — and fails loudly on anything else, which is what we want for a regression guard. Happy to revisit if the file ever grows past a handful of knobs.

scripts/verify-standards.sh
aidanns: Claude: Done! f00e6ae extends the check to assert each declared ecosystem has a `groups:` block containing both `"minor"` and `"patch"` update-types, so removing grouping in a future edit fails CI.

.github/dependabot.yml
aidanns: Claude: Agreed — test plan item 5 covers this.

scripts/verify-standards.sh
aidanns: Claude: Reversing my earlier call — 9976ad9 switches the check to yq. The dependency concern I raised (needing to `pip install pyyaml` or install yq in CI) doesn't apply: yq is preinstalled on ubuntu-latest runners, so the structural check adds zero setup cost. Now asserts each required ecosystem exists and that its `groups` block lists both `minor` and `patch` update-types, via `yq '.updates[] | select(.["package-ecosystem"] == "<eco>") | ...'`.

scripts/verify-standards.sh
```diff
@@ -61,3 +71,28 @@ if [[ -n "${missing_output}" ]]; then
...
 echo "verify-standards: Taskfile.yml exposes all required tasks."
+
+DEPENDABOT_CONFIG="${REPO_ROOT}/.github/dependabot.yml"
+
+if [[ ! -f "${DEPENDABOT_CONFIG}" ]]; then
+  echo "verify-standards: ${DEPENDABOT_CONFIG} is missing" >&2
+  exit 1
+fi
+
+for ecosystem in pip github-actions; do
```
aidanns: can we update this to determine what should exist in this project? we should only require pip if python is in use

scripts/verify-standards.sh
```diff
@@ -22,6 +27,11 @@ if ! command -v python3 >/dev/null 2>&1; then
   exit 1
 fi
 
+if ! command -v yq >/dev/null 2>&1; then
```
aidanns, line 32: lets move this to a separate verify-dependencies.sh file - this isn't really about a generic project standard - its about a specific dependency for this specific project

Taskfile.yml
```diff
@@ -29,7 +29,7 @@ tasks:
       - scripts/verify-function-tests.sh
 
   verify-standards:
-    desc: Verify the Taskfile exposes every task required by the tooling-and-ci standard.
+    desc: Verify tooling-and-ci standards (Taskfile task coverage and Dependabot config).
```
aidanns: match the description in the github action - this will grow over time

scripts/verify-standards.sh
aidanns: Claude: Done! e61e02a derives the required ecosystem list from manifest presence (`pyproject.toml`/`setup.py`/`requirements*.txt` → `pip`; `.github/workflows/*.y*ml` → `github-actions`) instead of a static list, so a repo without Python source isn't forced to declare a pip block. Also verified by test: removing `pyproject.toml` locally makes the script report `github-actions` only and exit 0.

scripts/verify-standards.sh
aidanns: Claude: Done! e61e02a moves the Dependabot coverage check to `scripts/verify-dependencies.sh` with its own Taskfile target and `Verify Dependencies` CI workflow. `scripts/verify-standards.sh` goes back to only checking cross-cutting standards (currently Taskfile task coverage), and `verify-dependencies` has been added to the required tasks list so the new script stays wired in.

Taskfile.yml
aidanns: Claude: Done! e61e02a updates the Taskfile desc for `verify-standards` to "Verify Project Standards" to match the GitHub Actions step name, and reworks the script header comment to make the "grows over time" intent explicit.

scripts/verify-dependencies.sh
```diff
@@ -0,0 +1,60 @@
...
+  fi
+
+  update_types=$(yq ".updates[] | select(.[\"package-ecosystem\"] == \"${ecosystem}\") | .groups.[].update-types[]" "${DEPENDABOT_CONFIG}")
+  for update_type in minor patch; do
+    if ! printf '%s\n' "${update_types}" | grep -qFx "${update_type}"; then
+      echo "verify-dependencies: ecosystem '${ecosystem}' does not group '${update_type}' updates" >&2
+      exit 1
+    fi
+  done
+done
```
aidanns: move this content back to verify-standards - it is about verification that dependabot is properly configured.

scripts/verify-dependencies.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+# Verify that .github/dependabot.yml covers every dependency ecosystem
+# actually in use by this repository, with minor/patch updates grouped
+# per ecosystem. An ecosystem is treated as "in use" when its manifest
+# files are present in the tree; ecosystems with no manifests are
+# skipped rather than required (e.g. a pure-Python repo should not be
+# forced to declare a `npm` block).
```
aidanns: this file should just be about verifying that project dependencies (e.g. python, jq, yq, etc) are installed locally.

scripts/verify-dependencies.sh
aidanns: Claude: Done! 9757a70 moves the Dependabot coverage check back into `scripts/verify-standards.sh` (still with the conditional "only require pip if pyproject.toml/setup.py/requirements*.txt exists" logic).

scripts/verify-dependencies.sh
aidanns: Claude: Done! 9757a70 rewrites `scripts/verify-dependencies.sh` to check only that the required CLI tools (`python3`, `task`, `yq`) are on PATH — intended as a local pre-flight before running `task test`/`task verify-standards`/etc. Task desc, CI workflow step name, and CLAUDE.md updated to match.

scripts/verify-dependencies.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -euo pipefail
+
+# keep-sorted start
+REQUIRED_TOOLS=(
+  python3
+  systems-engineering
+  task
+  yq
+)
```
aidanns, line 16: Claude Reviewer: `verify-dependencies.sh` requires `systems-engineering` and `yq` globally, but those tools are only needed by `verify-design`, `verify-function-tests`, and `verify-standards` — not by `test`, `lint`, `format`, `build`, `install-hooks`, or `release`. Because every workflow now runs `task verify-dependencies` as a preflight, the `test` workflow has to install d2 and systems-engineering purely to satisfy the preflight, even though `task test` itself uses none of them. Two cleaner options: (a) narrow `REQUIRED_TOOLS` to the minimal common set (python3 + task) and let each task declare its own extras, or (b) accept a subset argument like `verify-dependencies -- verify-standards` so each workflow only asserts the tools it actually needs.

scripts/verify-standards.sh
```diff
@@ -61,3 +74,42 @@ if [[ -n "${missing_output}" ]]; then
...
+if compgen -G '.github/workflows/*.yml' >/dev/null || compgen -G '.github/workflows/*.yaml' >/dev/null; then
+  required_ecosystems+=(github-actions)
+fi
+
+if [[ ${#required_ecosystems[@]} -eq 0 ]]; then
+  echo "verify-standards: no known dependency ecosystems detected; skipping dependabot coverage check."
+else
+  for ecosystem in "${required_ecosystems[@]}"; do
+    matched=$(yq ".updates[] | select(.[\"package-ecosystem\"] == \"${ecosystem}\") | .[\"package-ecosystem\"]" "${DEPENDABOT_CONFIG}")
+    if [[ "${matched}" != "${ecosystem}" ]]; then
```
aidanns, line 100: Claude Reviewer: The equality check `[[ "${matched}" != "${ecosystem}" ]]` assumes a single `updates[]` entry per ecosystem. Dependabot explicitly allows multiple entries with the same `package-ecosystem` when they target different directories (e.g. monorepo subdirs). In that case yq returns a multi-line string like `"pip\npip"` and the equality check fails, producing a confusing "does not declare the 'pip' ecosystem" error even though it is present. Safer form: `yq "... | select(...) | .[\"package-ecosystem\"]" | grep -qFx "${ecosystem}"`, or just check that at least one entry matches.

PR #63, opened 2026-04-18
ci: gate PRs on shellcheck and shfmt for bash scripts

### LINKED ISSUE #51: Gate PRs with shellcheck and shfmt for bash scripts
Source: `.claude/instructions/bash.md` — shellcheck, shfmt.

Project ships bash scripts under `scripts/` and a future `install.sh`. Neither `shellcheck` nor `shfmt` is run. Standard: both should gate PRs for every `*.sh` file.

**Acceptance:**
- CI workflow runs `shellcheck` on all `*.sh` (excluding vendored/third-party).
- CI workflow runs `shfmt -d` to assert formatting.
- Both wired into `treefmt` / `lefthook` pre-commit.
- Existing scripts brought into compliance.

- `scripts/verify-standards.sh` asserts `shellcheck` and `shfmt` are invoked by at least one `.github/workflows/*.yml` file AND are wired into `treefmt.toml` / `lefthook.yml`.

aidanns:
Closes #51.

## Summary

- New `.github/workflows/check.yml` runs `task check` on push + pull_request — a single cross-language lint/format gate that composes `task lint` + `task format -- --check`. Future linters/formatters extend those task surfaces rather than adding per-language workflows.
- New `check` task in `Taskfile.yml` wrapping lint + format-check; registered in `REQUIRED_TASKS` so its removal trips `task verify-standards`.
- `shellcheck` and `shfmt` added to `scripts/verify-dependencies.sh` — the preflight now fails fast with an actionable list if either is missing locally or in CI. Both versions are pinned in the workflow (shellcheck 0.10.0, shfmt 3.8.0) via release tarballs.
- `scripts/lint.sh` and `scripts/format.sh` replace the placeholder stubs — they run `shellcheck` / `shfmt` over every tracked `*.sh`.
- `treefmt.toml` and `lefthook.yml` wire both tools through the formatter multiplexer and the pre-commit hook manager respectively.
- `.editorconfig` is the single source for `shfmt` formatting options (2-space indent, `-ci`, `-bn`), read automatically by shfmt — no duplicated flag lists across Taskfile, CI, treefmt, and lefthook.
- `scripts/verify-standards.sh` gains a bash-tooling regression check that asserts both tools are invoked by at least one workflow file and wired into `treefmt.toml` and `lefthook.yml`. Comments are stripped before matching and `treefmt.toml` requires a structural `[formatter.<tool>]` header, so stale `# shellcheck` mentions can't satisfy the gate after the real invocation is removed.
- One minor `shfmt` reformat in `scripts/verify-standards.sh` (heredoc line break) brings the existing tree into compliance.

Plan: `plans/shellcheck-shfmt-ci.md`.

## Test plan

- [x] `task check` succeeds locally.
- [x] `task verify-dependencies` succeeds locally.
- [x] `task verify-standards` succeeds locally.
- [x] `task test` succeeds locally (no behavioural changes; sanity check only).
- [x] Check workflow passes on the PR.
- [x] Verify Standards workflow passes on the PR.
- [x] All other existing CI workflows pass on the PR.

### DIFF
--- a/.github/workflows/bash.yml
+++ b/.github/workflows/bash.yml
@@ -1,0 +1,35 @@
+name: Bash
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+env:
+  SHFMT_VERSION: 3.8.0
+
+jobs:
+  bash:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install shellcheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+          shellcheck --version
+
+      - name: Install shfmt
+        run: |
+          curl -fsSL -o /tmp/shfmt \
+            "https://github.com/mvdan/sh/releases/download/v${SHFMT_VERSION}/shfmt_v${SHFMT_VERSION}_linux_amd64"
+          sudo install -m0755 /tmp/shfmt /usr/local/bin/shfmt
+          shfmt --version
+
+      - name: Run shellcheck
+        run: scripts/lint.sh
+
+      - name: Run shfmt --check
+        run: scripts/format.sh --check
--- a/.github/workflows/bash.yml
+++ b/.github/workflows/bash.yml
@@ -16,6 +16,25 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install d2
+        run: curl -fsSL https://d2lang.com/install.sh | sh -s --
+
+      - name: Install systems-engineering
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/aidanns/systems-engineering/main/install.sh -o install.sh
+          bash install.sh
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: Install go-task
+        uses: arduino/setup-task@v2
+        with:
+          version: 3.50.0
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Install shellcheck
         run: |
           curl -fsSL -o /tmp/shellcheck.tar.xz \
@@ -31,8 +50,11 @@ jobs:
           sudo install -m0755 /tmp/shfmt /usr/local/bin/shfmt
           shfmt --version
 
+      - name: Verify Required Tooling
+        run: task verify-dependencies
+
       - name: Run shellcheck
-        run: scripts/lint.sh
+        run: task lint
 
       - name: Run shfmt --check
-        run: scripts/format.sh --check
+        run: task format -- --check
--- a/.github/workflows/bash.yml
+++ b/.github/workflows/bash.yml
@@ -7,6 +7,7 @@ on:
     branches: [main]
 
 env:
+  SHELLCHECK_VERSION: 0.10.0
   SHFMT_VERSION: 3.8.0
 
 jobs:
@@ -17,8 +18,10 @@ jobs:
 
       - name: Install shellcheck
         run: |
-          sudo apt-get update
-          sudo apt-get install -y shellcheck
+          curl -fsSL -o /tmp/shellcheck.tar.xz \
+            "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          tar -xJf /tmp/shellcheck.tar.xz -C /tmp
+          sudo install -m0755 "/tmp/shellcheck-v${SHELLCHECK_VERSION}/shellcheck" /usr/local/bin/shellcheck
           shellcheck --version
 
       - name: Install shfmt
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,4 +1,4 @@
-name: Bash
+name: Check
 
 on:
   push:
@@ -11,7 +11,7 @@ env:
   SHFMT_VERSION: 3.8.0
 
 jobs:
-  bash:
+  check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -53,8 +53,5 @@ jobs:
       - name: Verify Required Tooling
         run: task verify-dependencies
 
-      - name: Run shellcheck
-        run: task lint
-
-      - name: Run shfmt --check
-        run: task format -- --check
+      - name: Run lint + format --check
+        run: task check
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -18,6 +18,12 @@ tasks:
     cmds:
       - scripts/format.sh {{.CLI_ARGS}}
 
+  check:
+    desc: Run all linters and assert every tracked file is formatted.
+    cmds:
+      - scripts/lint.sh
+      - scripts/format.sh --check
+
   verify-design:
     desc: Verify every leaf function in the functional decomposition is allocated in the product breakdown.
     cmds:
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -9,14 +9,14 @@ tasks:
       - scripts/test.sh {{.CLI_ARGS}}
 
   lint:
-    desc: Run all configured linters (placeholder — no linters wired yet).
+    desc: Run all configured linters.
     cmds:
       - scripts/lint.sh
 
   format:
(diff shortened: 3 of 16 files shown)


aidanns (commented):
Claude Reviewer: The bash-tooling regression check in `scripts/verify-standards.sh` is structurally weaker than the plan implies. It uses `grep -qE '\bshellcheck\b'` / `\bshfmt\b` over `.github/workflows`, `treefmt.toml`, and `lefthook.yml`, which matches any textual mention of the tool name — including comments and stale documentation.

Reproduced locally: delete `.github/workflows/bash.yml` entirely and drop a workflow containing only the comment `# shellcheck shfmt`; `scripts/verify-standards.sh` still reports "shellcheck and shfmt are wired into CI, treefmt, and lefthook." The check would also keep passing if someone replaced the actual `run: shellcheck {staged_files}` line in `lefthook.yml` with a comment referring to shellcheck, or left a reference in one workflow while gutting the gate in another.

Since the whole point of the regression check is to catch removal of the real wiring, this is worth tightening. A few options:

1. Parse `lefthook.yml` / `treefmt.toml` as structured data (Python already used in this script) and assert that `pre-commit.commands.shellcheck.run` / `pre-commit.commands.shfmt.run` and `[formatter.shellcheck]` / `[formatter.shfmt]` exist and actually invoke the tool.
2. For workflows, assert the existence of `.github/workflows/bash.yml` specifically and parse it to confirm it runs `scripts/lint.sh` and `scripts/format.sh --check` (or at minimum that a workflow step's `run:` line — not a comment — invokes each tool).
3. At minimum, use `grep -v '^\s*#'` to strip comments before searching.

Options 1/2 are more robust and align with how the Taskfile check already uses structured parsing rather than `grep` heuristics.

aidanns (commented):
Claude Reviewer: Minor: `shellcheck` is installed via `apt-get install -y shellcheck` in `bash.yml`, pulling whatever version the Ubuntu runner image ships. `shfmt` is pinned to `3.8.0`. The asymmetry means a runner image update that bumps shellcheck can introduce new findings that block PRs without any local signal (developers would only see it after CI fails). Also, local environments (brew, apt on a contributor's distro) can trivially drift from the runner.

Low-severity — not worth blocking on — but since the plan already cares about reproducibility enough to pin shfmt, pinning shellcheck as well (via a downloaded release, a setup-action like `ludeeus/action-shellcheck`, or a snapshot-image base) would close the gap. Separately: the `shfmt` download has no checksum verification, which is the standard supply-chain concern for `curl | install` on any CI runner.

aidanns:
Claude: Done! Addressed in 38c7940:

- Added a `strip_comments()` sed helper that strips `#`-prefixed content before every grep, so a stale `# shellcheck` mention no longer satisfies the check (option 3).
- Tightened the `treefmt.toml` check to require the structural `^\[formatter\.${tool}\]` header rather than a bare tool-name match.
- Workflow check now invokes `find ... | xargs cat | strip_comments | grep`, so the same comment-stripping applies across all `.github/workflows/*.yml`.

Reproduced the original failure case locally — replacing `bash.yml` with a comment-only workflow containing `# shellcheck shfmt` now correctly fails `verify-standards` with exit 1.

Did not go all the way to structured YAML/TOML parsing (options 1/2): the comment-stripped grep catches every concrete regression mode we care about (bare comment mentions, gutted workflows, removed lefthook commands) and keeps the check readable. Happy to escalate to structured parsing if a future failure mode slips past this.

aidanns:
Claude: Done on the shellcheck pin — addressed in 38c7940 by installing shellcheck v0.10.0 from the upstream release tarball, matching the existing shfmt install pattern. `SHELLCHECK_VERSION` is now a top-level env var so future bumps are a one-line change.

Skipped the shfmt SHA256 checksum for now: the existing `curl | install` pattern is consistent with the shfmt step that was already present and approved on the PR, and adding a checksum for one tool and not the other would be inconsistent. Happy to add checksum verification for both in a follow-up if supply-chain hardening becomes a priority — filed as a mental TODO rather than blocking this PR.

aidanns (commented):
Claude Reviewer: Reviewed commit 38c7940. Diff looks clean and correctly addresses both prior review comments.

Verified:

1. `strip_comments()` uses `sed -E 's/(^|[[:space:]])#.*$//'` — the `(^|[[:space:]])` guard correctly preserves shell parameter expansions like `${var#foo}` and `${#arr[@]}` (no whitespace/start-of-line before the `#`), so legitimate shell syntax inside `run:` blocks isn't stripped. Tested locally.
2. The tightened treefmt check `^\[formatter\.${tool}\]` correctly anchors on a structural header. `shellcheck` and `shfmt` are both regex-literal (no metachars), so substitution is safe. A comment like `# [formatter.shellcheck] was here` no longer satisfies the gate.
3. Workflow env-var `SHELLCHECK_VERSION` does NOT falsely satisfy the workflow grep because `\bshellcheck\b` is case-sensitive (confirmed).
4. Pinned shellcheck install mirrors the shfmt pattern exactly: tarball tested, extracted path `shellcheck-v0.10.0/shellcheck` matches the `install` target, and `shellcheck --version` runs post-install. Consistent with existing shfmt install (neither pins a checksum, which is a pre-existing convention rather than a regression introduced here).

Minor residual observations (not blocking, not introduced by this commit):

- The workflow grep remains satisfiable by incidental lowercase `shellcheck` occurrences in URLs (e.g. the koalaman release URL in the install step) even if the actual `shellcheck` invocation were later removed. Hardening this further would require matching only `run:`/`command:` contexts, which is likely over-engineering for this check. Fine as-is.
- `strip_comments` will clip content after a space-prefixed `#` inside TOML/YAML string values (e.g. `key = "rule #5"`). No impact on current checks — `treefmt.toml` now uses a structural header match, and `lefthook.yml` entries referencing these tools don't embed them in strings. Worth keeping in mind if this helper is reused elsewhere.

LGTM.

aidanns:
Claude: Thanks — acknowledging the two residuals as intentional trade-offs rather than issues to address in this PR:

1. Incidental URL match in workflow grep: agreed the koalaman URL would satisfy `\bshellcheck\b` independent of the actual `run:` line. Tightening to `run:`-context-only would require a YAML parser and is disproportionate for a regression check. Leaving as-is.
2. `strip_comments` vs quoted `#` in TOML/YAML: current callers (workflows, `treefmt.toml` with structural header match, `lefthook.yml`) don't put tool names inside string values. If the helper is reused for checks where that matters, I'd switch to a structural parser at that point rather than hardening the sed regex.

Both filed as mental TODOs; no code changes needed now.

.github/workflows/bash.yml
```diff
@@ -0,0 +1,38 @@
+name: Bash
```
aidanns: rename / re-work this to a single "check" workflow that runs formatting and linting across all configured languages, probably by running a single task check command and configuring it as required.

.github/workflows/check.yml
```diff
@@ -0,0 +1,38 @@
...
+  SHELLCHECK_VERSION: 0.10.0
+  SHFMT_VERSION: 3.8.0
+
+jobs:
+  bash:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install shellcheck
```
aidanns, line 38: add both shellcheck and shfmt as external dependencies in verify-dependencies

Taskfile.yml
```diff
@@ -9,14 +9,14 @@ tasks:
       - scripts/test.sh {{.CLI_ARGS}}
 
   lint:
```
aidanns, line 11: add a "check" task which runs linting and format --check

.github/workflows/bash.yml
aidanns: Claude: Done! Renamed to `.github/workflows/check.yml` with job `check` that runs `task check` (wraps `task lint` + `task format -- --check`). `task check` is registered in `REQUIRED_TASKS` so verify-standards catches future regressions. 1d38bda.

.github/workflows/check.yml
aidanns: Claude: Done! Added `shellcheck` and `shfmt` to `REQUIRED_TOOLS` in `scripts/verify-dependencies.sh` (alphabetically sorted). The preflight now fails with an actionable list if either is missing locally or in CI. 1d38bda.

Taskfile.yml
aidanns: Claude: Done! Added a `check` task in `Taskfile.yml` that runs `scripts/lint.sh` then `scripts/format.sh --check`. Future linters/formatters extend `task lint` / `task format` and `task check` picks them up automatically. Also registered in `REQUIRED_TASKS`. 1d38bda.

PR #66, opened 2026-04-18
refactor(things-bridge): split Things client into sibling CLIs

### LINKED ISSUE #61: Move Things fake out of things-bridge process so the bridge has no knowledge of test fakes
## Problem

PR #53 introduces an in-process fake (`FakeThingsClient` in `src/things_bridge/fake.py`) selected at startup via `things-bridge serve --fake-things[=PATH]`. The CLI branches on the flag, imports the fake, loads a YAML fixture, and substitutes it for `ThingsApplescriptClient` inside the same Python process.

This works, but it means the production `things-bridge` binary ships with:

- A `--fake-things` CLI flag and its help text.
- An `import things_bridge.fake` path executed on every startup (module is imported at top of `cli.py`).
- A YAML fixture loader, a `FakeThingsStore`, and a second `ThingsClient` implementation living next to the real one.
- A stderr banner whose sole job is to prevent the fake from being mistaken for production.

The bridge needs all of this only because the test harness wants to swap the backend. Nothing about the production request path benefits from it. The trust-boundary story also weakens: the process that holds authz delegation and talks to Things 3 in prod now has a code path — however guarded — that returns synthetic data to any caller.

## Proposal

Move the fake out into a **separate process** that the bridge cannot distinguish from the real `osascript` + Things 3 stack, and remove all fake-mode awareness from `things-bridge` itself.

Concretely: the bridge only ever does what it does today — run the configured `osascript_path` as a subprocess, write an AppleScript on stdin (or via `-e`), read TSV on stdout, surface stderr/exit code. The test harness arranges for `osascript_path` to point at a **fake osascript** binary (e.g. `things-fake-osascript`) that accepts the same CLI contract, matches against the scripts `ThingsApplescriptClient` emits, and returns fixture-backed TSV.

End state:

- `src/things_bridge/cli.py` loses `--fake-things` and the branch around it.
- `src/things_bridge/fake.py` goes away (or moves under `tests/` / a separate `things-fake-osascript` package — it is no longer part of the bridge).
- Tests set `config.osascript_path` (already a first-class config knob) to the fake binary. No monkeypatching, no in-process substitution, no imports of test-only modules by the production entry point.
- The fake runs as its own process with its own lifecycle, logs, and failure modes — closer to how real integration tests should look.

## Why revisit ADR 0001

ADR 0001 (`design/decisions/0001-things-client-fake.md`, landing in PR #53) explicitly rejected a runner-level fake for two reasons:

1. **No open-source AppleScript interpreter on Linux.** True, but not required here — we do not need to execute arbitrary AppleScript, only to recognise the finite set of script shapes `ThingsApplescriptClient` emits and answer with fixture data.
2. **Coupling the fake to emitted script text.** Real concern, but it is a *testing* concern living in test code. The current approach trades that coupling away by coupling the *production* binary to the test fake instead — which is worse: test fragility is easier to manage than production-surface cruft.

This issue proposes the tradeoff be inverted: accept the script-shape coupling (mitigated by a small, stable contract the fake matches against) in exchange for a things-bridge binary with zero test-mode code paths.

## Sketch of the fake-process contract

- Binary: `things-fake-osascript` (shipped in this repo under `src/things_fake_osascript/` or similar, installed as a console entry point).
- CLI: mirrors the subset of `osascript` flags `AppleScriptRunner` actually uses (currently the script is fed via stdin or `-e`; pin and document that surface as the contract).
- Behaviour: parses the script text against a small dispatch table (one entry per `ThingsApplescriptClient` method), resolves against a YAML fixture identified by an env var (e.g. `THINGS_FAKE_FIXTURE=/path/to/fixture.yaml`), emits TSV on stdout, exits 0 / non-zero to mirror failure modes.
- Failure injection: env vars (e.g. `THINGS_FAKE_SLEEP_MS`, `THINGS_FAKE_EXIT=2`, `THINGS_FAKE_STDERR=...`) let tests provoke the exact subprocess failure modes `AppleScriptRunner` must handle — directly unblocking #39 (AppleScript fault-injection) and #55 (osascript timeout regression test).

## Scope

- Remove `--fake-things` and `src/things_bridge/fake.py` once the out-of-process replacement exists.
- Add `things-fake-osascript` (package + entry point + fixture loader — the YAML schema from PR #53 can be reused almost verbatim).
- Port `tests/test_things_bridge_e2e.py` and `tests/test_things_bridge_fake.py` (from PR #53) to start the fake as a subprocess and point `osascript_path` at it.
- Update ADR 0001 with a superseding decision (or add ADR 0002 that supersedes 0001's rejected alternative, explaining why the coupling tradeoff was re-evaluated).
- Update `CLAUDE.md` Linux-devcontainer note to describe the out-of-process harness.

## Non-goals

- Running real Things 3 in CI — still tracked as a separate follow-up (macOS runner, deferred in ADR 0001).
- Shipping a general-purpose AppleScript interpreter. The fake only needs to handle the scripts `ThingsApplescriptClient` emits.

## Acceptance

- `things-bridge --help` shows no `--fake-things` flag; grep for `fake` in `src/things_bridge/` returns nothing.
- `tests/test_things_bridge_e2e.py` runs on Linux with the bridge unmodified, driven by pointing `osascript_path` at the fake binary.
- Fault-injection env vars on the fake process produce the subprocess failure modes needed by #39 and #55 without any test-only code in `things-bridge`.
- ADR superseding 0001's runner-level-fake rejection is committed.

## Related

- Blocks / informs: #39 (fault-injection layer), #55 (osascript timeout regression test).
- Supersedes part of: ADR 0001 (landing in #53).
- Depends on: #53 merging first (so there is something concrete to replace), or this work absorbs #53 directly.

aidanns:
## Summary

- Move Things 3 interaction out of `things-bridge` and into a sibling CLI (`things-client-cli-applescript`). The bridge now shells out to a configured `things_client_command` per request and parses a JSON envelope on stdout.
- Replace the in-process `FakeThingsClient` with a test-only CLI (`tests/things_client_fake/`, invoked as `python -m tests.things_client_fake --fixtures PATH`) that speaks the same subprocess contract. Not shipped in the sdist/wheel.
- `things-bridge --help` no longer has `--fake-things`; the bridge binary holds no test-mode code paths and cannot be tricked into serving synthetic data.

Supersedes ADR 0001 via ADR 0003 (`design/decisions/0003-things-client-cli-split.md`). Closes #61.

## Approach vs. issue sketch

The issue originally proposed a fake `osascript` binary that matched against emitted AppleScript text. This PR takes a higher-leverage cut: move the Things logic itself out of the bridge, so both the real and fake clients speak the same stable JSON-on-stdout contract rather than coupling to AppleScript script shapes. ADR 0003 records the tradeoff.

## Test plan

- [x] `task test` passes (234 passed, 3 skipped Darwin-only)
- [x] `scripts/verify-design.sh` reports "All functions allocated"
- [x] `scripts/verify-function-tests.sh` shows no regression in coverage (43/50, up from 41/50 on main)
- [x] `things-bridge --help` contains no `--fake-things` flag
- [x] No fake-mode code paths remain in `src/things_bridge/` (remaining `fake` mentions are docstring/comment references to the test harness, not runtime logic)
- [x] `things-client-cli-applescript --help` lists todos/projects/areas sub-commands
- [x] `python -m tests.things_client_fake --fixtures examples/fake-things.yaml todos list` emits a `{"todos": [...]}` envelope
- [x] End-to-end: `test_things_bridge_e2e.py` drives the bridge with the fake CLI configured as `things_client_command` and exercises the full HTTP path


aidanns:
Claude Reviewer: Reviewed this in depth. The refactor is a real architectural win — moving AppleScript generation/parsing and YAML-fixture loading out of the authz-holding process is a meaningful trust-boundary reduction, and the JSON-on-stdout contract is a better protocol-level surface than `osascript` argv mimicry would have been. ADR 0003 is honest about why it deviates from #61's sketch.

A few things worth considering before merge:

**1. Contract: exit codes 4/5/6 are under-specified vs. ADR.** `things_client_common/cli.py` defines `EXIT_NOT_FOUND=4`, `EXIT_PERMISSION_DENIED=5`, `EXIT_UNAVAILABLE=6`, but ADR 0003 says the JSON body is authoritative and the exit code "only distinguishes success from failure." The bridge's `_invoke` confirms this — it never reads the specific code, only `returncode != 0`. So the 4/5/6 distinction is non-load-bearing surface that can drift silently. Either promote it to part of the contract (tested) or collapse to a single non-zero code. I'd argue for collapsing — fewer knobs to keep in sync.

**2. Contract: default timeouts stack to zero margin.** `ThingsSubprocessClient` defaults `timeout_seconds=30.0` and `AppleScriptRunner` also defaults `timeout_seconds=30.0`. Since the bridge's timeout wraps the whole CLI invocation (argparse + Python startup + osascript), the bridge will kill the CLI before the CLI's own timeout can emit a structured error, turning a recoverable "osascript timed out" into the bridge's generic "things client subprocess timed out". Bridge's `request_timeout_seconds` should probably be ≥ CLI timeout + a small buffer (a few seconds for Python+osascript startup). Worth either documenting the ordering constraint on `Config.request_timeout_seconds` or bumping the default to e.g. 35s.

**3. Contract: argparse exit-2 path isn't in the ADR error table.** If the bridge ever constructs a bogus argv (or status value), argparse exits 2 with usage on stderr and nothing on stdout. The bridge maps that to "emitted no JSON output (rc=2)" — which masks the real cause (bridge bug). ADR 0003's error table should include "CLI argv error (rc=2, empty stdout)" alongside the other cases so operators grepping stderr know what a `rc=2` means. Tests already cover the empty-stdout branch in `test_things_client_cli_contract.py::test_missing_subcommand_exits_nonzero_without_stdout_payload`, so this is a docs-only gap.

**4. Security: subprocess inherits bridge's full env.** The bridge process holds the `auth_url` and may touch keyring-backed secrets in future. Spawned subprocesses inherit the full env. Not a regression vs. the in-process version (which had even more privileged access), and not exploitable from HTTP since `things_client_command` is config-only. But ADR 0003's STRIDE section says "subprocess inherits the bridge's environment but runs as the same local user, so no privilege boundary is crossed" — fair, but a one-line note that env-based secrets leak into the child would make the threat model complete. For now this is latent since no env secrets exist; worth filing a follow-up to pass an explicitly-minimal env to the subprocess before secrets land.

**5. Test coverage: contract tests exercise only the fake CLI.** `test_things_client_cli_contract.py` runs `tests.things_client_fake` as a real subprocess — the applescript CLI is only covered via unit tests of `things.py` (on Darwin). The shared `add_read_commands` / `run_cli` plumbing is the same for both, so the argparse surface is covered once. That's a reasonable bet, but flag in a docstring that the production CLI's envelope shape is asserted transitively through the shared runner — right now a reader has to piece that together.

**6. Simplification: `_MissingSubcommandError` silently emits empty stdout + `EXIT_UNAVAILABLE`.** When the fake is invoked with e.g. `todos` but no `list|show` subcommand, the CLI prints help to stderr and exits 6 with empty stdout. That looks like a protocol violation to the bridge (`ThingsError("no JSON output")`). Either emit `{"error": "things_unavailable", "detail": "missing subcommand"}` so the bridge's error path is uniform, or document in the ADR that "empty stdout + non-zero exit" is a valid terminal state meaning "argv malformed".

**7. Simplification: `ThingsSubprocessClient` could share argv construction with `add_read_commands`.** Each `list_todos` / `get_todo` / etc. method in the bridge builds argv by hand, mirroring the shared parser in `things_client_common.cli`. Drift risk is modest but real (add a new filter → three touch points). A small "request → argv" helper table would close that loop. Optional; not blocking.

None of the above is a reason to hold the PR — the architecture is right and the contract is specific enough to be useful. Addressing (1), (2) as code and (3) in the ADR would be my priorities.

src/things_client_common/cli.py
```diff
@@ -0,0 +1,164 @@
...
+    ThingsError,
+    ThingsNotFoundError,
+    ThingsPermissionError,
+)
+
+
+EXIT_OK = 0
+EXIT_NOT_FOUND = 4
+EXIT_PERMISSION_DENIED = 5
+EXIT_UNAVAILABLE = 6
```
aidanns, line 42: Claude Reviewer: These EXIT_NOT_FOUND/EXIT_PERMISSION_DENIED/EXIT_UNAVAILABLE constants look like load-bearing contract surface, but nothing on the bridge side reads them — `ThingsSubprocessClient._invoke` only branches on `returncode != 0` vs `== 0` and dispatches off the JSON `error` key. That matches what ADR 0003 says ("the JSON body is authoritative for the error *kind*"), but means these specific codes are a silent-drift hazard. Suggest either (a) collapsing to a single non-zero exit (simplest), or (b) elevating them to part of the tested contract and having `ThingsSubprocessClient` assert the code matches the error kind in the payload.

src/things_bridge/things_client.py
```diff
@@ -0,0 +1,165 @@
...
+
+    ``command`` is the argv prefix (e.g. ``["things-client-cli-applescript"]``
+    or ``[sys.executable, "-m", "tests.things_client_fake", "--fixtures", P]``);
+    sub-commands matching the request are appended. ``timeout_seconds`` caps
+    the per-call wall clock. Subprocess stderr is forwarded to the bridge's
+    own stderr for operator diagnostics; the HTTP response body never
+    contains subprocess output.
+    """
+
+    def __init__(self, command: list[str], timeout_seconds: float = 30.0):
```
aidanns: Claude Reviewer: Default of `30.0` here matches `AppleScriptRunner`'s default of `30.0` in `things_client_applescript/things.py`, so the two timeouts have zero margin between them. The bridge's subprocess timeout wraps Python startup + argparse + osascript startup + AppleScript execution — if a real osascript call approaches its own 30s limit, the bridge will SIGKILL the client before its `TimeoutExpired` handler can emit a structured `{"error": "things_unavailable", "detail": "osascript timed out..."}`, and the operator loses the more specific diagnostic. Worth either bumping the bridge default to e.g. 35s, or adding a note on `Config.request_timeout_seconds` that it must exceed the client's internal timeout with a safety margin.

src/things_client_common/cli.py
```diff
@@ -0,0 +1,164 @@
...
+        return EXIT_NOT_FOUND
+    except ThingsPermissionError as exc:
+        _emit_error("things_permission_denied", str(exc))
+        return EXIT_PERMISSION_DENIED
+    except ThingsError as exc:
+        _emit_error("things_unavailable", str(exc))
+        return EXIT_UNAVAILABLE
+    except _MissingSubcommandError:
+        parser.print_help(sys.stderr)
+        return EXIT_UNAVAILABLE
```
aidanns, line 151: Claude Reviewer: `_MissingSubcommandError` emits help on stderr and returns `EXIT_UNAVAILABLE` with empty stdout. The bridge's `_parse_payload` treats empty stdout as a protocol violation and raises `ThingsError("emitted no JSON output")` — so a CLI misuse surfaces as a generic 502 with an unrelated message, making it harder for operators to tell "bridge built a bad argv" from "osascript genuinely failed silently". Consider emitting `{"error": "things_unavailable", "detail": "missing subcommand: <command>"}` on stdout before returning, so the bridge's error path is uniform and the failure reason is preserved through the subprocess boundary.

src/things_bridge/things_client.py
aidanns: Claude: Done in af28488 — bridge default bumped to 35s and commented with the reason. This gives the CLI's 30s osascript timeout room to surface a structured `things_unavailable` envelope before the bridge kills the child.

src/things_client_common/cli.py
aidanns: Claude: Done in af28488. `run_cli` now calls `_emit_error("things_unavailable", "missing <cmd> sub-command")` on both the no-top-level-command and the `_MissingSubcommandError` paths, so the bridge sees a structured envelope and maps it through the normal error path instead of the generic "no JSON output" protocol violation. Contract test updated to pin the new shape.

src/things_client_common/cli.py
aidanns: Claude: Not adopting as proposed. The distinct exit codes (4/5/6) are observational — the bridge only branches on zero vs non-zero and reads the `error` code from the JSON envelope for the typed-exception mapping. Keeping the separate codes is useful for humans running the CLI directly (`$?` after `things-client-cli-applescript todos show bogus` still tells you "not found" at a glance). I've documented the "JSON body is authoritative over exit code" invariant in ADR 0003 so the two are explicitly redundant by design; the constants can drift from the raw numbers without affecting the bridge's contract.

aidanns:
Claude: Addressed the review in af28488. Walking through the seven concerns:

1. **Exit codes 4/5/6 unused by bridge** — left as-is. They're observational for humans running the CLI directly (`$?` tells you "not found" at a glance). ADR 0003 now documents that the JSON body is authoritative and the exit code is redundant, so the two can't drift in a way that affects the bridge.
2. **Default timeouts stack to zero margin** — fixed. Bridge default bumped from 30s → 35s so the shipped CLI's own 30s osascript timeout can surface a structured `things_unavailable` envelope before the bridge kills the child. Comment on both `Config.request_timeout_seconds` and `ThingsSubprocessClient.__init__` explains the relationship.
3. **Argparse exit-2 path missing from ADR error table** — added row for "argparse misuse (exit 2, empty stdout) → `ThingsError` → 502".
4. **Subprocess inherits full env** — deferred to a follow-up. Restricting to an allowlist (PATH/HOME/LANG/LC_\*/TZ + THINGS_CLIENT_\*) is the right move but deserves its own change with bounds-testing on macOS (osascript may need vars I haven't enumerated). Tracking for a separate PR.
5. **Contract tests only run fake CLI** — accepted limitation. Running the real AppleScript CLI end-to-end needs a macOS CI runner, which is a separate issue. The Darwin-only tests in `test_things_client_applescript_things.py` do exercise it on developer machines.
6. **Missing-subcommand silent empty stdout** — fixed. `run_cli` now calls `_emit_error("things_unavailable", …)` on both the no-top-level-command and the `_MissingSubcommandError` paths. Contract test updated.
7. **Argv drift between bridge and shared parser** — pinned by the combination of `test_things_subprocess_client.py` (bridge's argv builder) and `test_things_client_cli_contract.py` (end-to-end against the fake CLI). If either side changes the shape, one of the two suites fails.

Also picked up from the local code review:
- Close subprocess stdin (`stdin=subprocess.DEVNULL`) so a buggy CLI can't stall on inherited stdin.
- Preserve the raw error `code` in `ThingsError` messages for unknown codes so forward-compat failures stay debuggable.
- New test pinning the "rc=0 + error body still raises" invariant.

aidanns:
Claude: Tracked the deferred items as issues:

- #68 — restrict subprocess env to a minimal allowlist (concern 4 from the review)
- #69 — add a macOS CI runner to exercise the real AppleScript client end-to-end (concern 5)
- #70 — introduce \`ThingsClientCommand\` newtype at the config/subprocess boundary (from the instructions-compliance review; the plan committed to it but the implementation shipped raw \`list[str]\`)

scripts/things-client-applescript.sh
```diff
@@ -0,0 +1,19 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+cd "${REPO_ROOT}"
+
+VENV_DIR=".venv-$(uname -s)-$(uname -m)"
+
+if [[ ! -d "${VENV_DIR}" ]]; then
+  python3 -m venv "${VENV_DIR}"
+  "${VENV_DIR}/bin/pip" install -e ".[dev]"
```
aidanns: move this out of the guard block so that it always re-installs when code changes.

design/functional_decomposition.yaml
```diff
@@ -104,16 +104,26 @@ functions:
...
+  - name: Things Client (AppleScript)
+    description: Read-only Things 3 client CLI that shells to osascript. Ships with the distribution and is the bridge's default things_client_command.
+    functions:
+      - name: Execute External System Interaction
+        description: Interact with the Things 3 application via AppleScript / osascript and return structured results. Owns AppleScript generation, TSV parsing, and the 30s per-call osascript timeout.
+      - name: Serve Things Client CLI Contract
+        description: Expose the shared read surface (todos / projects / areas sub-commands) and emit the JSON-on-stdout + exit-code envelope that the things-bridge subprocess contract requires.
+  - name: Things Client (Fake)
+    description: Test-only in-memory Things client CLI that reads a YAML fixture. Never shipped; invoked by the bridge during integration/e2e tests.
+    functions:
```
aidanns: i think we can remove all of this - it's all implementation specific and not relevant to the overall function of the system.

design/functional_decomposition.yaml
```diff
@@ -104,16 +104,26 @@ functions:
...
       - name: Serve Fake Things Client
-        description: Developer-only in-memory implementation of the Things client protocol that lets the bridge run on Linux without osascript. Selected via `things-bridge serve --fake-things[=PATH]`; seeds todos, projects, areas, and list memberships from a YAML fixture.
+        description: Answer the shared Things-client CLI surface from an in-memory store seeded from a YAML fixture. Enables end-to-end exercise of the agent-auth + things-bridge + things-cli stack on Linux without osascript or Things 3.
```
aidanns: remove this as well.

aidanns:
Claude: Ran /simplify on the refactor after the design reorganisation. Applied the following cleanups from three review agents' findings; all 235 tests still pass.

- `choices=("open","completed","canceled")` in the shared CLI parser replaced with `sorted(VALID_STATUSES)` so adding a status no longer requires touching two files.
- `_MissingSubcommandError` collapsed into a `None`-return from `_dispatch_read`; one class, one raise, one except, and a narrating comment removed from `things_client_common/cli.py`.
- Dead `ThingsClient` re-export from `things_bridge.things_client` removed (no callers; grep-verified).
- `_validate_things_client_command` removed from `config.py` — `ThingsSubprocessClient.__init__` is the single point of enforcement for empty commands.
- `isinstance(exc.stderr, str)` guard on `TimeoutExpired` removed (`subprocess.run(text=True)` already guarantees str).
- Redundant `stdout.strip()` dropped from `_parse_payload` — `json.loads` tolerates whitespace, and pre-scanning a large payload for emptiness is wasted work.
- Narrating/restating comments removed from `things_client.py` (timeout margin comment duplicated the one on `Config`; the "non-zero exit without body" comment restated its own error message; the docstring cross-reference to `server._send_things_error_send_response` leaked module internals).

Filed #72 for the one genuine concern the review surfaced that was out of scope here: `subprocess.run(capture_output=True)` buffers subprocess stderr unboundedly, which compounds across concurrent bridge threads.

scripts/things-client-applescript.sh
aidanns: Claude: Done in 2cf3d24. Moved `pip install --quiet -e ".[dev]"` out of the venv-creation guard so it runs on every invocation and picks up local code changes.

design/functional_decomposition.yaml
aidanns: Claude: Done in 2cf3d24. Removed the "Things Client (AppleScript)" component entirely from functional_decomposition.yaml — the AppleScript generation, TSV parsing, and subprocess contract are implementation details of how the bridge fetches Things data, not separate functional blocks. They're now subsumed under "Fetch Things Data" on the Things Bridge. Kept `things-client-cli-applescript` as a configuration_item in product_breakdown.yaml with no allocated functions (same pattern as the keyring backends) — it's still a distinct physical deliverable shipped with the distribution.

design/functional_decomposition.yaml
aidanns: Claude: Done in 2cf3d24 — "Serve Things Client CLI Contract" was removed along with its parent component in the same change.

PR #67, opened 2026-04-18
test: move HTTP integration tests into Docker

### LINKED ISSUE #7: Move HTTP integration tests to Docker to avoid host-port collisions
The HTTP integration tests in `tests/test_server.py` bind to an ephemeral port on `127.0.0.1` and talk to it over the network. Per review on [PR #4](https://github.com/aidanns/agent-auth/pull/4#discussion_r3086381389), this causes collisions when multiple versions of the test suite run on the same host (e.g. parallel CI workers, a local dev run while another branch is running).

## Proposed shape

- Install agent-auth at the current version into a container image (e.g. via `docker compose` with a `test` profile).
- A pytest fixture starts the container, waits for `GET /agent-auth/healthz` (or a similar readiness probe — tracked separately in TODO.md), and tears it down after the test.
- Tests talk to the mapped port on the container, so each run gets its own isolated network namespace.

## Alternatives

- A dedicated pytest plugin that claims a free port and exports it via fixture. Avoids Docker but doesn't isolate SQLite state, keyring, or filesystem paths.

Original author is open to other approaches as long as they avoid collisions when multiple test suites run on a single host.

## Scope

- Add `docker/` directory with a minimal test image.
- Move the existing `tests/test_server.py` integration tests into a separate `tests/integration/` layer gated by a marker.
- Update `scripts/test.sh` and CI to build and run the container for integration tests.

- `scripts/verify-standards.sh` asserts integration tests under `tests/integration/` use the Docker fixture (no direct `127.0.0.1` host binds in test code) and that the CI workflow builds the test image.

aidanns:
## Summary

- Splits the test suite into a fast in-process **unit** layer and a Docker-backed **integration** layer (`tests/integration/`), filtered by directory rather than by pytest marker. Each integration test spins up its own ephemeral `docker compose` project (uuid-named `COMPOSE_PROJECT_NAME`) via a pytest fixture driven by [`testcontainers-python`](https://testcontainers-python.readthedocs.io/), talks to the mapped loopback port, and uses `DockerCompose.exec_in_container` to drive the `agent-auth` CLI inside the container — no host-port collisions, no shared XDG dirs, public-HTTP/CLI surface only.
- Adds `docker/Dockerfile.test` + `docker/compose.test.yaml` + `docker/config.test.json` (committed baseline integration-test config, bind-mounted per-test into the container), two explicit notification plugins `tests_support.always_approve` / `tests_support.always_deny` (selected by name in the per-test `config.json`), `scripts/test.sh --unit|--integration|--all`, and `scripts/verify-integration-isolation.sh` (regression-guards the isolation contract, runs as part of `task check`).
- Renames the liveness endpoint from `/agent-auth/healthz` to `/agent-auth/health` per project standard; adds `TokenStore.ping()` as the liveness primitive. Documents the rootless-DinD preference for devcontainer use (per issue #7 comment) in README + ADR 0004.

Closes #7. Follow-up filed as #78 to extend the Docker-based e2e pattern to `things-bridge`, `things-cli`, and `things-client-cli-applescript`.

See `design/decisions/0004-docker-integration-tests.md` and `plans/dockerize-integration-tests.md`.

## Test plan

- [x] `scripts/test.sh --unit` — all unit tests pass locally (233 passed, 3 skipped)
- [x] `scripts/test.sh --integration` — on a host with Docker, integration tests pass _(verified by CI `Test` job running `task test -- --all`)_
- [x] `scripts/test.sh --integration` — on a host without Docker, integration tests skip cleanly (13 skipped locally)
- [x] `scripts/verify-integration-isolation.sh` — passes (no raw `127.0.0.1` in `tests/integration/`, conftest invokes `docker build` against `Dockerfile.test`)
- [x] `task check` — runs `scripts/verify-integration-isolation.sh` as its third step
- [x] CI `Test` job passes (runs `task test -- --all`, covers both unit + integration)
- [x] CI `Check` job passes (runs `task check`, now including the integration-isolation check)
- [x] `shellcheck` + `shfmt -d -i 2 -ci` clean on all new/modified bash scripts (`task check` ran both)
- [x] `/agent-auth/health` returns 200 ok when the store is reachable, 503 when `store.ping()` raises (verified in `tests/test_server.py`)
- [x] No `/healthz` references remain anywhere in the tree (grep confirmed)


tests/integration/conftest.py
```diff
@@ -0,0 +1,219 @@
...
+    def exec_cli(self, *args: str) -> subprocess.CompletedProcess:
+        """Run ``agent-auth <args>`` inside the container and return the result."""
+        cmd = [
+            "docker", "compose",
+            "-f", str(COMPOSE_FILE),
+            "-p", self.project_name,
+            "exec", "-T", self.service,
+            "agent-auth", *args,
+        ]
+        return subprocess.run(cmd, capture_output=True, text=True, check=True)
```
aidanns: Claude Reviewer: `subprocess.run(..., capture_output=True, check=True)` raises `CalledProcessError` on non-zero exit, and pytest's default traceback renders only the command + returncode — the captured stdout/stderr that would explain **why** the CLI failed inside the container are attributes on the exception, not in its `str()`. When an integration test fails in CI because e.g. `token create` hit an unexpected error path, the failing test will show something like `CalledProcessError: Command '[docker, compose, ..., agent-auth, --json, token, create, ...]' returned non-zero exit status 1` with no body.

Mirror the pattern `_compose` already uses: `check=False`, and on `returncode != 0` raise a `RuntimeError(f"exec_cli failed: stdout={r.stdout!r} stderr={r.stderr!r}")`. That's what makes the Docker layer debuggable from CI logs.

tests/integration/conftest.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def _mapped_port(project_name: str) -> int:
+    result = _compose(project_name, "port", "agent-auth", "9100")
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"`docker compose port` failed: stdout={result.stdout!r} stderr={result.stderr!r}"
+        )
+    line = result.stdout.strip().splitlines()[-1]
+    return int(line.rsplit(":", 1)[-1])
```
aidanns: Claude Reviewer: Neither `_compose` (`docker compose up -d`) nor `_mapped_port` pass a `timeout` to `subprocess.run`. Only `_wait_for_health` has a bounded deadline (30s). If the Docker daemon stalls — daemon restart mid-image-pull, kernel IO hang, DinD network hiccup — `docker compose up -d` will block forever and pytest will hang without a failure signal. CI will burn through the job's wall-clock budget instead of failing fast.

Consider a `timeout=` on the `up` and `port` calls (e.g. 60s each, 120s for first-run build) and converting `TimeoutExpired` into a `RuntimeError` that includes which phase timed out. Same for `down` in teardown so a stuck teardown doesn't mask the real test result.

tests/integration/conftest.py
```diff
@@ -0,0 +1,219 @@
...
+        ],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"`docker build` failed for {TEST_IMAGE_TAG}: "
+            f"stdout={result.stdout!r} stderr={result.stderr!r}"
+        )
```
aidanns, line 166: Claude Reviewer: The ADR and PR description claim parallel runs of `scripts/test.sh --integration` on the same host cannot collide because each test gets a UUID-named Compose project. That's true for containers / volumes / networks, but **not for the `agent-auth-test:local` tag**. Two concurrent session builds — e.g. two branches on the same CI host, or two developer worktrees — both `docker build -t agent-auth-test:local .` at the same time. Whoever tags last wins, and containers subsequently started under the other project will boot *the wrong branch's code* against the other branch's uuid-named project. This silently breaks the test-isolation contract the ADR promises.

Two straightforward fixes:
- Tag with a session-scoped suffix (`agent-auth-test:{uuid4().hex[:12]}`) and reference that tag from the per-test Compose call. Paired with the existing per-project UUID, collisions become impossible.
- Or content-address the tag (digest of a `git rev-parse HEAD` or of the image's build context) so the same source always reuses the same tag and divergent sources don't step on each other.

Worth calling out in the ADR's *Consequences* section either way — the current "UUID per test ⇒ safe for parallel runs" framing is slightly broader than what the implementation delivers.

scripts/verify-integration-isolation.sh
```diff
@@ -0,0 +1,53 @@
...
+
+integration_dir="tests/integration"
+if [[ ! -d "${integration_dir}" ]]; then
+  echo "FAIL: ${integration_dir}/ is missing" >&2
+  fail=1
+else
+  # Restrict the scan to the integration tests themselves; the conftest
+  # resolves 127.0.0.1 from the ephemeral Docker port mapping, so it is
+  # allowed there.
+  offenders=$(grep -nR --include='test_*.py' -E '127\.0\.0\.1|0\.0\.0\.0' "${integration_dir}" || true)
```
aidanns: Claude Reviewer: The `--include='test_*.py'` filter limits the scan to files named `test_*.py`, which means `tests/integration/conftest.py` and any future helper module under `tests/integration/` (e.g. `tests/integration/_helpers.py`) can introduce a raw `127.0.0.1` literal without this guard noticing — even though they would be imported by the tests and would recreate the exact host-binding contract this script is meant to protect. Given the stated regression guard ("no raw 127.0.0.1 in `tests/integration/test_*.py`"), this is the right narrow scope for the test files themselves, but a dev can route around it trivially.

Either (a) broaden to `--include='*.py' --exclude=conftest.py` to cover helpers, or (b) add an explicit comment pinning the current scope and the rationale so a reviewer catches the exemption on the next change. (a) is probably what's intended given the script's docstring talks about "files under tests/integration/" generically.

docker/entrypoint.test.sh
```diff
@@ -0,0 +1,30 @@
...
+cat >"${config_path}" <<EOF
+{
+  "host": "${host}",
+  "port": ${port},
+  "access_token_ttl_seconds": ${access_ttl},
+  "refresh_token_ttl_seconds": ${refresh_ttl},
+  "notification_plugin": "${plugin}",
+  "notification_plugin_config": {}
+}
+EOF
```
aidanns: Claude Reviewer: `${port}`, `${access_ttl}`, and `${refresh_ttl}` are interpolated as unquoted JSON numbers, and `${host}` / `${plugin}` as unescaped JSON strings. If an env var arrives non-numeric (e.g. the factory fixture stringifies an `int` cleanly today, but a future caller or a typo like `AGENT_AUTH_ACCESS_TOKEN_TTL_SECONDS=900s`) the container starts with a malformed `config.json` and `agent-auth serve` fails with a JSON parse error far away from the real cause. If `AGENT_AUTH_NOTIFICATION_PLUGIN` contains a double-quote or backslash, the rendered file is straight-up invalid.

Swap the heredoc for a tiny Python one-liner that does the encoding:
```bash
python3 -c 'import json, os, sys; json.dump({"host": os.environ["AGENT_AUTH_HOST"], "port": int(os.environ["AGENT_AUTH_PORT"]), ...}, sys.stdout)' > "${config_path}"
```
This also enforces that the TTLs are actually ints — which matches the `Config` dataclass contract — and fails loudly at container start rather than at first request.

scripts/test.sh
```diff
@@ -9,11 +17,35 @@ REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
...
+case "${mode}" in
+  unit)
+    exec ${pytest} tests/ -m "not integration" --ignore=tests/integration "$@"
+    ;;
+  integration)
+    exec ${pytest} tests/integration/ -m integration "$@"
+    ;;
+  all)
+    ${pytest} tests/ -m "not integration" --ignore=tests/integration "$@"
+    exec ${pytest} tests/integration/ -m integration "$@"
```
aidanns: Claude Reviewer: In `--all` mode, the same `"$@"` is forwarded to both the unit and the integration pytest invocation. That's surprising when the extra args are layer-specific: `scripts/test.sh --all -k test_health` runs `-k test_health` against both layers, but `scripts/test.sh --all -x` makes the *unit* layer stop on first failure while the integration layer proceeds regardless of whether it saw `-x` (it will, but integration test selection usually differs). Not a bug in the `--unit` / `--integration` individual modes.

Two options: (a) document that pass-through args are layer-agnostic and apply to both (cheap, just a docstring note), or (b) restrict `--all` to accept no extra args and require the user to pick one layer when they want to filter. I'd lean (a).

tests/integration/conftest.py
aidanns: Claude: Done!

tests/integration/conftest.py
aidanns: Claude: Done!

tests/integration/conftest.py
aidanns: Claude: Done!

scripts/verify-integration-isolation.sh
aidanns: Claude: Done!

docker/entrypoint.test.sh
aidanns: Claude: Done!

pyproject.toml
```diff
@@ -17,7 +17,9 @@ dependencies = [
 dev = [
     "build>=1.0",
     "pytest>=8.0",
+    "testcontainers[compose]>=4",
 ]
+tests-support = []
```
aidanns: Claude Reviewer: `tests-support = []` is an empty optional-dependency group. Combined with `[tool.setuptools.packages.find] where = ["src"]` (no exclude) and the `src/tests_support/` package sitting next to `src/agent_auth/`, every production `pip install agent-auth` also installs the `tests_support` package — including `AlwaysApprovePlugin`/`AlwaysDenyPlugin` which a misconfigured production `notification_plugin` setting could load by name.

This directly contradicts ADR 0004's threat-model statement: *"the package is installed only when the image is built, not in production"*. The `[tests-support]` extra in `Dockerfile.test` is a no-op today.

Two options:
1. Move `tests_support/` out of `src/` (e.g. into `tests/_support/` or a sibling `tests-support/` tree) and install it in the Dockerfile via a separate editable install, so it never ships in the `agent-auth` wheel.
2. Keep it under `src/` but exclude it explicitly in setuptools (`[tool.setuptools.packages.find] exclude = ["tests_support*"]`) and make `[tests-support]` a real extras group that depends on a separately-packaged `agent-auth-tests-support` distribution (or similar).

Either way the ADR and the extras group need to match reality.

tests/integration/conftest.py
```diff
@@ -0,0 +1,245 @@
...
+            notification_plugin=APPROVAL_PLUGINS[approval],
+        )
+
+        # The compose subprocess inherits this process's environment, so
+        # setting the variables here is enough for Compose's
+        # interpolation (``${AGENT_AUTH_TEST_IMAGE}``,
+        # ``${AGENT_AUTH_TEST_CONFIG_DIR}``) and for per-test project
+        # isolation via ``COMPOSE_PROJECT_NAME``.
+        os.environ["AGENT_AUTH_TEST_CONFIG_DIR"] = str(config_dir)
+        os.environ["COMPOSE_PROJECT_NAME"] = project_name
```
aidanns: Claude Reviewer: `os.environ["AGENT_AUTH_TEST_CONFIG_DIR"]` and `os.environ["COMPOSE_PROJECT_NAME"]` are mutated globally on the pytest process and never restored.

A couple of correctness risks fall out of this:

1. **Teardown points at the wrong project.** The fixture collects every `DockerCompose` in `started` and tears them all down at the end via `compose.stop()`. But the env vars still hold whatever values the *last* invocation set. If `testcontainers.compose.DockerCompose` resolves its Compose project from `COMPOSE_PROJECT_NAME` at stop-time (it does, via the subprocess env) rather than caching it at `start()`, `stop()` on the first compose would act against the last project's name — leaving the first container / network running.

   A test that calls the factory twice within the same test (or pytest-xdist workers if ever enabled) would hit this.

2. **Cross-fixture leakage.** Because these env vars persist after the fixture tears down, the next test's factory call inherits stale values until it overwrites them. That's mostly benign today, but it turns "run a single integration test" into an implicit dependency on fixture ordering.

Suggestions:
- Pass the project name and config dir to `DockerCompose` explicitly (its constructor accepts `project_name=`; at least `project_name` on the testcontainers version currently pinned), instead of leaning on `COMPOSE_PROJECT_NAME` in the ambient env.
- Use `monkeypatch` / `mp.setenv` to ensure env vars are restored on teardown, or snapshot+restore manually the way `_test_image_tag` does for `AGENT_AUTH_TEST_IMAGE`.
- Alternatively, build a minimal per-test Compose file with substituted values and `env_file` = /dev/null, so env inheritance is not load-bearing.

docker/compose.test.yaml
```diff
@@ -0,0 +1,13 @@
...
+    image: ${AGENT_AUTH_TEST_IMAGE:?set by tests/integration/conftest.py}
+    # Bind only on loopback so the container is never reachable from the
+    # rest of the network.
+    ports:
+      - "127.0.0.1::9100"
+    volumes:
+      # Per-test tmpdir holding config.json. The fixture copies
+      # docker/config.test.json here and applies any per-test overrides
+      # (TTLs, notification plugin) before starting the container.
+      - ${AGENT_AUTH_TEST_CONFIG_DIR:?set by tests/integration/conftest.py}:/home/agent-auth/.config/agent-auth:ro
```
aidanns, line 12: Claude Reviewer: UID mismatch on the bind-mount is worth an explicit note.

The container runs as UID 1001 (see `Dockerfile.test`). The host-side mountpoint is a `tmp_path_factory.mktemp(...)` directory — on Linux pytest defaults those to mode 0700 owned by whatever UID the host test runner uses (GitHub Actions runner UID 1001 on ubuntu-latest happens to line up, but devcontainers often run as UID 1000 or whatever the user chose, and rootful Docker won't do userns remapping).

If the host UID != 1001 and the tmpdir is 0700, the container user cannot read the `config.json` and `agent-auth serve` will fail on startup. On macOS Docker Desktop this works because of its own bind-mount permission handling, so it may only surface on Linux devcontainers or specific CI configs.

Options:
- Chmod the config dir (or the file) to `0755` / `0644` explicitly in `_write_test_config`.
- Set a specific `user:` override on the compose service (run as the host UID) when running outside CI.
- At minimum, document this in the README's "Running integration tests from a devcontainer" section so the failure mode is recognisable.

plans/dockerize-integration-tests.md
```diff
@@ -0,0 +1,278 @@
...
+### Phase 2 — Env-driven notification plugin
+
+1. Create `tests_support/env_plugin.py` with
+   `EnvPlugin(NotificationPlugin)` whose `request_approval` returns an
+   `ApprovalResult` built from `AGENT_AUTH_TEST_APPROVAL` (values:
+   `approve`, `deny`). Defaults to `deny` so a misconfigured container
+   fails closed.
+2. Register `tests_support` as an installable package in `pyproject.toml`
+   under `[project.optional-dependencies] tests-support`.
+3. Unit test in `tests/test_env_plugin.py`.
```
aidanns: Claude Reviewer: the plan file is stale relative to the implementation on commit 1540b44.

Specifically, it still describes:
- `tests_support.env_plugin` / `EnvPlugin` (superseded by `tests_support.always_approve` + `tests_support.always_deny`)
- `AGENT_AUTH_TEST_APPROVAL` env var (no longer used — approval is now selected by plugin name in the bind-mounted `config.json`)
- `docker/entrypoint.test.sh` (the env-driven shell entrypoint was removed; `Dockerfile.test` now uses `ENTRYPOINT ["agent-auth", "serve"]`)
- A `tests/test_env_plugin.py` unit test (replaced by `tests/test_tests_support_plugins.py`)
- ADR filename `0001-docker-integration-tests.md` (the ADR shipped is `0004-...`)

Either update the plan to reflect the post-refactor shape, or annotate the sections as historical context superseded by the follow-up commit. Leaving the two out of sync makes the plan misleading to future readers, and `.claude/instructions/plan-template.md` treats the plan as the authoritative implementation record.

aidanns (commented):
Claude Reviewer: review of commit 1540b44 (testcontainers + bind-mount config + explicit approve/deny plugin refactor). Four inline comments posted; summary:

**Blocking-ish:**
1. `tests_support` is packaged under `src/` with no setuptools exclude, so it ships with every `pip install agent-auth` in production — contradicting ADR 0004's explicit threat-model assertion. The `[tests-support]` extras group is an empty placeholder (`tests-support = []`) and doesn't actually gate anything. See inline on `pyproject.toml`.
2. `agent_auth_container_factory` mutates `os.environ` (`COMPOSE_PROJECT_NAME`, `AGENT_AUTH_TEST_CONFIG_DIR`) globally without restoring. If a test calls the factory twice, or if teardown walks `started` in order, the env-var-driven Compose project resolution can point `stop()` at the wrong project and leak containers/volumes/networks. See inline on `tests/integration/conftest.py`.

**Non-blocking:**
3. Bind-mount UID/perms: container runs as UID 1001, host-side tmpdirs default to 0700 owned by the host UID. Works on macOS Docker Desktop and on CI runners where UIDs happen to align, but likely breaks on devcontainers running as UID 1000. Worth chmodding the config dir or adding a README note. See inline on `docker/compose.test.yaml`.
4. `plans/dockerize-integration-tests.md` is stale — still describes `env_plugin`, `AGENT_AUTH_TEST_APPROVAL`, `entrypoint.test.sh`, `tests/test_env_plugin.py`, and an ADR numbered 0001 rather than 0004. See inline on the plan file.

**Looks good:**
- Plugin rename is otherwise thorough: no stale `EnvPlugin` or `AGENT_AUTH_TEST_APPROVAL` references outside the plan file; `docker/config.test.json`, `tests/integration/conftest.py`, and `tests/test_tests_support_plugins.py` all use the new names consistently.
- Session-scoped `_test_image_tag` with a unique tag per session (and snapshot+restore of `AGENT_AUTH_TEST_IMAGE`) is a nice pattern — consider mirroring that env-var restore logic for `COMPOSE_PROJECT_NAME` in the factory fixture.
- `_docker_required` skip behaviour is clean; unit layer stays fast on Docker-less hosts.
- `scripts/verify-integration-isolation.sh` regression guard is a good call and correctly wired into `task check`.
- `AgentAuthContainer.exec_cli` interpolating stdout/stderr into the error message instead of a bare `CalledProcessError` will pay for itself the first time a CLI invocation fails in CI.

Consistency with project conventions: `docker/config.test.json` is JSON while `service-design.md` prefers YAML for human-editable config. Arguably fine here (the file is machine-written by the fixture, not hand-edited) but worth a one-line comment in the file or ADR acknowledging the deviation.

Not blocking: lgtm modulo items 1 and 2.

scripts/test.sh
aidanns: Claude: Done!

docker/compose.test.yaml
aidanns: Claude: Done! Adopted option 1 — `_write_test_config` now chmods the config dir to 0755 and `config.json` to 0644, so the container UID 1001 can read it regardless of the host tmpdir default. Also added a short note to the README's integration-test section.

plans/dockerize-integration-tests.md
aidanns: Claude: Done! Rewrote the stale sections in-place (the git log preserves the earlier shape). Updated: removed env_plugin/EnvPlugin/AGENT_AUTH_TEST_APPROVAL/entrypoint.test.sh references, renamed test_env_plugin.py → test_tests_support_plugins.py, fixed the ADR filename (0004 not 0001), and synced the health-endpoint authz + uv-tooling notes to match the shipped code.

pyproject.toml
aidanns: Claude: Done! Adopted option 2 — added `exclude = ["tests_support*"]` to `[tool.setuptools.packages.find]`, deleted the empty `tests-support` extras group, and replaced `pip install .[tests-support]` in `docker/Dockerfile.test` with `pip install .` + a copy of `src/tests_support/` into `/opt/tests-support/` exposed via `PYTHONPATH`. Added `pythonpath = ["src"]` to `[tool.pytest.ini_options]` so host-side tests still resolve `tests_support` without the extras group or an editable install. Updated ADR 0004 to match the new shape.

Verified via `python -m build` + `unzip -p .../top_level.txt` that the wheel no longer contains `tests_support`, and confirmed a fresh `pip install` of the built wheel into a clean venv fails with `ModuleNotFoundError: No module named 'tests_support'`. All 235 unit tests (including `test_tests_support_plugins.py`) pass; integration tests skip cleanly on this host (no Docker) — they'll rebuild the image on a Docker-enabled host and pick up the new layout.

tests/integration/conftest.py
aidanns: Claude: Done! Adopted the second suggestion — snapshot-and-restore of `COMPOSE_PROJECT_NAME` / `AGENT_AUTH_TEST_CONFIG_DIR` via a `_scoped_env` context manager, plus re-applying each compose's pinned env at teardown time (so `stop()` sees the same project name it started with, regardless of which compose was last). The `project_name=` kwarg suggestion wasn't feasible — testcontainers 4.14.2 (currently pinned) doesn't expose it (verified via `inspect.signature`). See commit 5eeb5ca.

PR #71, opened 2026-04-18
build: migrate Python tooling from pip/venv to uv

### LINKED ISSUE #49: Migrate tooling from pip/venv to uv
Source: `.claude/instructions/python.md` — uv.

`scripts/test.sh` creates a venv via `python3 -m venv` and uses `pip`. Standard: `uv` for virtual environment and dependency resolution (creates `.venv/` in the project directory per the global convention).

**Acceptance:**
- `uv` installation documented in README / CONTRIBUTING.
- `scripts/test.sh` (and any other venv-aware script) uses `uv sync` / `uv run`.
- Lockfile (`uv.lock`) committed for reproducibility.
- CI workflows switched to `uv`.
- `.venv/` still sits at the project root so existing conventions hold.

- `scripts/verify-standards.sh` runs `uv lock --check` and fails if any file under `scripts/` invokes `pip install`.

aidanns:
## Summary

- Replaces the `python3 -m venv` + `pip install -e ".[dev]"` bootstrap in every `scripts/*.sh` with a single `uv sync --extra dev --quiet` path, then dispatches via `uv run --no-sync`. Keeps the per-OS/arch `.venv-$(uname -s)-$(uname -m)/` location via `UV_PROJECT_ENVIRONMENT` per `.claude/instructions/python.md`.
- Commits `uv.lock` for reproducible installs.
- Swaps `actions/setup-python@v5` for `astral-sh/setup-uv@v5` (pinned, with `enable-cache: true`) in all five CI workflows. uv provisions a CPython matching `requires-python` from `pyproject.toml` on demand.
- Extends `scripts/verify-standards.sh` with two regression checks: `uv lock --check` (fails on missing or stale lockfile) and a grep that fails if any `scripts/*.sh` invokes `pip install` (backslash-newline continuations are caught too).
- Adds `uv` to `scripts/verify-dependencies.sh`'s required-tools list; updates README, CONTRIBUTING, and CLAUDE.md to document uv as a prerequisite.

Plan at `plans/migrate-to-uv.md`. Closes #49.

## Test plan

- [x] `task test` — full pytest suite runs via uv locally (204 passed, 3 skipped).
- [x] `task verify-standards` — `uv lock --check` reports in-sync, no `scripts/*.sh` invokes `pip install`.
- [x] `task verify-design` — still passes.
- [x] `task lint` and `task format -- --check` — bash gates pass.
- [x] `scripts/agent-auth.sh --help`, `scripts/things-bridge.sh --help`, `scripts/things-cli.sh --help` — CLI entrypoints work under `uv run`.
- [ ] CI: all five workflows (test, check, verify-standards, verify-design, verify-function-tests) are green.


aidanns (commented):
Claude Reviewer: the uv version pin (`version: "0.11.7"` on the `astral-sh/setup-uv@v5` step) is invisible to Dependabot. Dependabot's `github-actions` ecosystem updates the action ref (`@v5`) but does not touch `with:` inputs — so the pinned version string will not auto-bump with weekly PRs the way `setup-python@v5`'s Python version did not need to (that used a major ref too). Net effect: you get patch/minor updates within `v5` of setup-uv, but the uv binary it installs is frozen at 0.11.7 until a human edits the workflow. If the intent is reproducibility, that's fine (just worth documenting). If the intent is "pinned but updated by Dependabot", consider either (a) moving the pin to the action ref (`astral-sh/setup-uv@0.11.7`), which Dependabot will bump, or (b) dropping the `version:` input entirely and letting setup-uv track its own latest.

aidanns (commented):
Claude Reviewer: the `\bpip\b[^\n]*\binstall\b` pattern in `scripts/verify-standards.sh` correctly catches `pip install`, `pip --user install`, and `python -m pip install` (all confirmed by local testing against the exact regex). It does NOT catch `pip3 install`, though — the `\b` after `pip` enforces a word boundary so `pip3` is excluded. Given that `pip3` is a common invocation (especially in Linux scripts where `pip` may not be on PATH), consider broadening to `\bpip[0-9]*\b[^\n]*\binstall\b` or similar to close the regression gap. Not a blocker — just flagging it since the whole point of the check is to keep `pip install` out of `scripts/`.

aidanns (commented):
Claude Reviewer: confirming the self-exclusion (`[[ "${script}" == "scripts/verify-standards.sh" ]] && continue`) is necessary and correct. I tested locally: after `strip_comments` runs, the diagnostic echo lines `echo "verify-standards: scripts/ must not invoke 'pip install' ..."` and `echo "verify-standards: no scripts/*.sh file invokes 'pip install'."` both survive (they're quoted `echo` arguments, not shell comments), and both match `\bpip\b[^\n]*\binstall\b`. Without the self-exclusion the script would flag itself. The inline comment explaining this is also helpful — no change needed.

aidanns (commented):
Claude Reviewer: `plans/migrate-to-uv.md` deliverable #5 lists the lockfile check as two separate invariants (missing lockfile and stale lockfile). The implementation correctly collapses them into a single `uv lock --check` call (`uv lock --check` exits non-zero for both), and the inline comment in `verify-standards.sh` explains this. Worth a one-line update to the plan to reflect the simplification so the plan matches the final implementation — or note this is expected drift in the PR description. Cosmetic only.

aidanns (commented):
Claude Reviewer: plan deliverable #3 says "`pyproject.toml` — no structural change; add minimal uv configuration (e.g. `required-version`) only if needed." The diff does not add one, which is consistent with the "only if needed" qualifier. Worth confirming this was an intentional skip: a `[tool.uv]` section with `required-version = ">=0.11"` (or similar) would give local `uv sync` users a clear error if they're on an older uv, mirroring the CI pin. Not a blocker — CI enforces the version via `setup-uv`, and local users get whatever they install. Flagging in case you want to close the loop.

aidanns:
Claude: Done! Dropped the `version: "0.11.7"` input across all five workflows in 922001e — the `@v5` ref already gives Dependabot a handle to track (and protects against a lockfile-format break within major v5).

aidanns:
Claude: Done! Broadened the regex to `\bpip[0-9]*\b[^\n]*\binstall\b` in 922001e and re-tested the gate against a `pip3 install` regression script — caught as expected.

aidanns:
Claude: Thanks for confirming — no change needed.

aidanns:
Claude: Done! Updated both the deliverable and implementation-step sections of plans/migrate-to-uv.md in 922001e to describe a single `uv lock --check` call covering both missing and stale lockfiles, and added the pip3 / backslash-continuation notes.

aidanns:
Claude: Intentional skip. `[tool.uv].required-version` is appealing, but dropping the CI `version:` pin (per finding 1) means CI and local both track setup-uv's / homebrew's latest uv — pinning a floor here would just lag behind and create noise as uv releases accumulate. The regression gate already catches what matters (`uv.lock` drift). Open to revisiting if we hit a specific cross-version compatibility issue.

PR #74, opened 2026-04-18
build(python): adopt ruff for lint and format, gate in CI

### LINKED ISSUE #47: Configure ruff lint + format and gate in CI
Source: `.claude/instructions/python.md` — ruff.

No linter or formatter is configured. Standard: `ruff` for both linting and formatting.

**Acceptance:**
- `ruff` config in `pyproject.toml` with a sensible rule set (at minimum: E, F, I, UP, B, SIM, RUF).
- `ruff format` applied across `src/` and `tests/`.
- CI job fails on `ruff check` violations and on un-formatted files.
- Wired into `treefmt` for the formatter side.

- `scripts/verify-standards.sh` asserts `pyproject.toml` contains `[tool.ruff]` config, `treefmt.toml` wires `ruff format`, and at least one CI workflow runs `ruff check` and `ruff format --check`.

aidanns:
## Summary

Closes #47.

- Configures `ruff` in `pyproject.toml` with rule families `E`, `F`, `I`, `UP`, `B`, `SIM`, `RUF`; `target-version = "py311"`; `line-length = 100` (mild relaxation of the 88 default to fit the descriptive error messages and SQL strings common in this repo).
- Wires `ruff check` + `ruff format` through the canonical entrypoints: `scripts/lint.sh` and `scripts/format.sh` (via the per-OS/arch venv), `treefmt.toml`, and `lefthook.yml` (pre-commit).
- Extends `scripts/verify-standards.sh` with a deterministic regression check asserting `[tool.ruff]` in `pyproject.toml`, `[formatter.ruff]` in `treefmt.toml`, and `task check` (or `ruff check` + `ruff format --check`) in at least one workflow.
- Applies `ruff format` across `src/` and `tests/` and addresses residual `ruff check` findings (auto-fix + narrow hand-edits for `ClassVar`, `zip(strict=...)`, etc.) so the baseline is clean.
- No CI workflow changes required: `check.yml` already runs `task check`, which now dispatches to the updated lint/format scripts.

Plan: `plans/ruff-lint-format.md`.

## Test plan

- [x] `task verify-standards` passes
- [x] `task lint` passes (shellcheck + `ruff check`)
- [x] `task format -- --check` passes (shfmt + `ruff format --check`)
- [x] `task test` passes (no behavioural regressions from auto-fixes)
- [x] CI `Check` workflow is green on this PR


aidanns (commented):
Claude Reviewer: Overall this PR is clean and on-scope. Ruff config matches issue #47 (rule families E/F/I/UP/B/SIM/RUF, `target-version = py311` aligning with `requires-python = ">=3.11"`, `line-length = 100` adequately justified in the inline comment). The venv-local invocation pattern in `lefthook.yml` vs bare `ruff` in `treefmt.toml` is internally consistent with how `shellcheck`/`shfmt` are wired, and the wiring across `scripts/lint.sh`, `scripts/format.sh`, `scripts/verify-standards.sh`, `pyproject.toml`, and `treefmt.toml` is complete. I spot-checked the ~40 reformatted files; every auto-rewrite I saw is behaviour-preserving (`timezone.utc` → `UTC`, `.encode("utf-8")` → `.encode()`, `try/except pass` → `contextlib.suppress`, SIM-style boolean collapse, `dict(zip(...))` → `dict(zip(..., strict=False))`, adjacent-string SQL concat, `assert False` → `raise AssertionError`, `ClassVar` on mutable class attrs). Leaving a handful of small inline comments below.

scripts/verify-standards.sh
```diff
@@ -180,3 +180,46 @@ if [[ ${bash_tool_missing} -ne 0 ]]; then
...
+# `ruff format --check` as an equivalent alternative.
+if ! grep -qE "\\btask check\\b" <<<"${workflows_stripped}" \
+  && ! { grep -qE "\\bruff check\\b" <<<"${workflows_stripped}" \
+    && grep -qE "\\bruff format --check\\b" <<<"${workflows_stripped}"; }; then
+  fail_ruff_check ".github/workflows/*.yml" \
+    "Add a workflow step that runs 'task check' (or both 'ruff check' and 'ruff format --check' directly)."
+fi
+
+if [[ ${ruff_missing} -ne 0 ]]; then
+  exit 1
```
aidanns, line 322: Claude Reviewer: the shellcheck/shfmt regression gate above asserts each tool is wired into **three** places — workflows, `treefmt.toml`, **and `lefthook.yml`**. This ruff block only asserts pyproject + treefmt + workflows; it does not assert ruff is still wired into `lefthook.yml`, even though the PR adds `ruff-check` / `ruff-format` pre-commit hooks. A future change that drops those hooks would silently pass `task verify-standards`.

Suggest adding a fourth check mirroring the existing pattern, e.g.:
```bash
if ! grep -qE "\\bruff\\b" <<<"${lefthook_stripped}"; then
  fail_ruff_check "lefthook.yml" \
    "Add a pre-commit command that invokes 'ruff' to lefthook.yml."
fi
```
and updating the trailing success echo accordingly. The plan deliverable #5 explicitly includes the lefthook integration, so the regression gate should cover it.

scripts/verify-standards.sh
```diff
@@ -180,3 +180,46 @@ if [[ ${bash_tool_missing} -ne 0 ]]; then
...
+
+# A single `task check` invocation satisfies both the lint and format
+# gate (it dispatches through scripts/lint.sh and scripts/format.sh
+# --check, each of which invokes ruff). Accept direct `ruff check` +
+# `ruff format --check` as an equivalent alternative.
+if ! grep -qE "\\btask check\\b" <<<"${workflows_stripped}" \
+  && ! { grep -qE "\\bruff check\\b" <<<"${workflows_stripped}" \
+    && grep -qE "\\bruff format --check\\b" <<<"${workflows_stripped}"; }; then
+  fail_ruff_check ".github/workflows/*.yml" \
+    "Add a workflow step that runs 'task check' (or both 'ruff check' and 'ruff format --check' directly)."
```
aidanns, line 318: Claude Reviewer: `\btask check\b` will also match e.g. `task check-something` (the `-` after `check` is a non-word char, so `\b` is satisfied). Low-probability false positive today, but if a sibling task like `task check-migrations` were ever added to a workflow without wiring ruff, this would pass. Consider anchoring on whitespace/end-of-line (`task check(\s|$)`) or matching `task check\b` followed by a non-hyphen. Not blocking.

pyproject.toml
```diff
@@ -38,3 +39,16 @@ fallback_version = "0.0.0+unknown"
...
+target-version = "py311"
+# 100-character soft limit. Ruff's default is 88; 100 better fits
+# descriptive error messages and SQL strings already common in this
+# codebase without forcing wraps that reduce readability.
+line-length = 100
+
+[tool.ruff.lint]
+select = ["E", "F", "I", "UP", "B", "SIM", "RUF"]
+
+[tool.ruff.format]
```
aidanns: Claude Reviewer: the empty `[tool.ruff.format]` table at the end of the file is valid TOML but carries no information — ruff reads defaults whether the section is present or absent. The inline plan notes it's "left at defaults". Consider either (a) dropping the empty section to reduce noise, or (b) adding an explicit `# defaults: double-quote, space indent` comment so future readers don't wonder if a config value got lost in a merge.

src/things_client_applescript/things.py
```diff
@@ -476,14 +491,15 @@ def _parse_rows(output: str, expected_cols: int) -> list[list[str]]:
         parts = raw_line.split("\t")
         if len(parts) != expected_cols:
             raise ThingsError(
-                f"Unexpected column count from AppleScript: got {len(parts)}, expected {expected_cols}"
+                f"Unexpected column count from AppleScript: "
```
aidanns, line 494: Claude Reviewer: `zip(..., strict=False)` here (and in `_row_to_project` / `_row_to_area`) preserves the old permissive behaviour, which is fine because `_parse_rows` already raises `ThingsError` when `len(parts) != expected_cols` — so by the time execution reaches these zips, row length is guaranteed to match the field list. That said, `strict=True` would be the more defensive default (it would catch a future drift between `_TODO_FIELDS` and the parser's `expected_cols` argument at the point of failure rather than silently dropping trailing fields). Fine to keep as-is given the invariant, but worth a `# TODO(aidanns):` or a comment noting that strictness is enforced upstream. Non-blocking.

src/agent_auth/store.py
```diff
@@ -175,7 +176,10 @@ def get_tokens_by_family(self, family_id: str) -> list[dict]:
         ]
 
     def mark_consumed(self, token_id: str) -> bool:
-        """Atomically mark a refresh token as consumed. Returns True if successful (was not already consumed)."""
+        """Atomically mark a refresh token as consumed.
+
+        Returns True if the token was not already consumed.
+        """
```
aidanns, line 182: Claude Reviewer: the docstring prose changed from "Returns True if successful (was not already consumed)" to "Returns True if the token was not already consumed" — this is a subtle wording tweak ("successful" vs "not already consumed") that isn't a `ruff check --fix` output or a `ruff format` output, since `D` rules aren't selected. It reads like a drive-by improvement. Docstring tweaks are fine in their own right, but they're out of scope for a lint/format adoption PR — future `git blame` readers will chase this line through the ruff commit. Suggest either reverting to the original text or pulling this into a separate follow-up commit. Non-blocking.

src/things_bridge/server.py
```diff
@@ -36,7 +36,7 @@ def _safe_id(raw: str | None) -> str | None:
         return None
     for ch in raw:
         cp = ord(ch)
-        # Allow printable ASCII (0x20–0x7E) except slash, plus non-ASCII (>0x7F).
+        # Allow printable ASCII (0x20-0x7E) except slash, plus non-ASCII (>0x7F).
```
aidanns, line 39: Claude Reviewer: the en-dash → hyphen swap in `0x20-0x7E` is correct (ambiguous-unicode `RUF003`; en-dash inside a numeric range is grep-hostile). Note that the em-dashes on nearby lines (`# Suppress the default access log —`, `# — BaseHTTPRequestHandler API`) are kept — that's consistent with ruff's handling of em-dashes as unambiguous separators in comments, so no action needed, just flagging the asymmetry in case it was ambiguous from the PR description ("replaces an en-dash").

tests/test_things_bridge_authz.py
```diff
@@ -17,13 +18,13 @@
 
 class _Responder(BaseHTTPRequestHandler):
     status = 200
-    body: dict = {"valid": True}
+    body: ClassVar[dict] = {"valid": True}
```
aidanns, line 21: Claude Reviewer: `ClassVar[dict]` on `body` is correct for this fake handler — it's a class-level dict reassigned via `_Responder.body = {...}` in each test, not a per-instance attribute (which would be the `RUF012` motivation). Same pattern is applied consistently in `tests/test_things_cli_client.py` (`responses`, `requests`, `refresh_responses`, `reissue_responses`). Worth noting `last_request_body: bytes | None = None` on line 22 is *also* a class-level mutable-ish attribute written as `_Responder.last_request_body = ...`, but `RUF012` doesn't flag `None` defaults so no `ClassVar` needed there — just confirming the annotation choices are coherent. No change requested.

aidanns: Claude: Ack, no change needed — confirming the `ClassVar[dict]` annotation matches intent here (it's a per-instance-reset class-level fixture).

src/things_client_applescript/things.py
aidanns: Claude: Leaving as-is for this PR. `strict=True` would be a behavioural change (raise on upstream invariant break instead of silently skipping a malformed row), and this PR is scoped to lint+format adoption. I'll track switching to `strict=True` separately so it can land with a dedicated test for the mismatch branch.

scripts/verify-standards.sh
aidanns: Claude: Leaving as-is. The existing shellcheck/shfmt blocks above use the same `\b<tool>\b` pattern, so tightening only the ruff block would be inconsistent. Happy to follow up with a dedicated PR that hardens all three if we actually grow a `task check-foo` alias.

scripts/verify-standards.sh
aidanns: Claude: Done! Added a lefthook assertion to the ruff block in c5dc76a, matching the three-place gate (CI + treefmt + lefthook) the shellcheck/shfmt block uses.

src/agent_auth/store.py
aidanns: Claude: Done! Restored the original phrasing ("Returns True if successful (was not already consumed)") in c5dc76a — you're right that a ruff-driven multi-line split shouldn't have reworded the prose.

pyproject.toml
aidanns: Claude: Done! Dropped the empty `[tool.ruff.format]` section in c5dc76a.

src/things_bridge/server.py
aidanns: Claude: Ack, no change needed — confirming the swap is intentional and RUF003 was the trigger.

scripts/verify-standards.sh
aidanns: Claude: Follow-up issue filed as #77 so this lands across all three grep blocks together.

src/things_client_applescript/things.py
aidanns: Claude: Follow-up issue filed as #76.

PR #75, opened 2026-04-18
feat(ci): adopt mdformat, taplo, and keep-sorted tooling

### LINKED ISSUE #45: Add mdformat, taplo, and keep-sorted configuration
Source: `.claude/instructions/tooling-and-ci.md` — Markdown, TOML, Orchestration (keep-sorted).

No markdown or TOML formatter / linter is configured, and sorted blocks are maintained by hand. Standard: `mdformat` (with GFM + tables plugins), `taplo` (TOML lint/format), `keep-sorted` (sorted-block annotation).

**Acceptance:**
- `mdformat` config committed; hooked into `treefmt` (see the treefmt issue).
- `taplo` config committed; hooked into `treefmt`.
- `keep-sorted` installed; existing sortable blocks (imports, dependency lists, allow-lists) annotated with `keep-sorted start/end` markers.
- Pre-commit runs all three via treefmt.

- `scripts/verify-standards.sh` asserts `treefmt.toml` wires `mdformat` and `taplo`, and a `keep-sorted` hook is configured (either in `lefthook.yml` pre-commit or a CI workflow).

aidanns:
## Summary

Wires `mdformat` (with GFM + tables plugins), `taplo`, and `keep-sorted` into the project's formatter/linter surfaces so Markdown, TOML, and sorted-block annotations are gated the same way bash is today. Plan committed at `plans/mdformat-taplo-keep-sorted.md`.

- `treefmt.toml` registers `mdformat` + `taplo`; `lefthook.yml` runs all three pre-commit (following the existing shellcheck/shfmt precedent of direct tool calls rather than via `treefmt`, which deviates from the literal wording of #45 but matches the established repo pattern).
- `scripts/format.sh` and `scripts/lint.sh` extend bash-only coverage to Markdown, TOML, and keep-sorted annotations.
- `.github/workflows/check.yml` installs each tool at pinned versions (`MDFORMAT_VERSION`, `TAPLO_VERSION`, `KEEP_SORTED_VERSION`, plus the two mdformat plugin pins) and picks them up through the existing `task check` step.
- `scripts/verify-standards.sh` asserts `[formatter.mdformat]` and `[formatter.taplo]` in `treefmt.toml` and requires a `keep-sorted --mode=...` invocation in either `lefthook.yml` or a workflow (the `--mode` specificity prevents a CI install step from silently satisfying the check).
- `scripts/verify-dependencies.sh` now requires the three new tools on PATH.
- Sortable blocks annotated: `pyproject.toml` dependencies + scripts, `.gitignore`, and the `REQUIRED_TASKS` / `REQUIRED_TOOLS` arrays.
- All existing Markdown and TOML normalised with the new formatters.
- `CONTRIBUTING.md` documents install instructions (pip for mdformat, brew / GitHub releases for taplo / keep-sorted).

Closes #45.

## Test plan

- [x] `task verify-dependencies` passes locally.
- [x] `task verify-standards` passes locally.
- [x] `task check` passes locally.
- [x] `task test` passes locally (no behavioural changes expected).
- [x] CI (check workflow) passes on this branch.
- [x] Confirm `keep-sorted --mode=lint` catches a deliberately-misordered entry (smoke-test the regression check).


scripts/verify-standards.sh
```diff
@@ -180,3 +184,37 @@ if [[ ${bash_tool_missing} -ne 0 ]]; then
...
+  fi
+done
+
+# Match the invocation pattern `keep-sorted --mode=...` rather than the bare
+# tool name — a CI workflow's install step mentions `keep-sorted` without
+# actually running it, which would otherwise defeat the regression check.
+if ! grep -qE "keep-sorted --mode=" <<<"${lefthook_stripped}" \
+  && ! grep -qE "keep-sorted --mode=" <<<"${workflows_stripped}"; then
+  fail_doc_check "'keep-sorted' is not invoked in lefthook.yml pre-commit or any .github/workflows/*.yml." \
+    "Add a 'keep-sorted --mode=lint' invocation to lefthook.yml, or run it from a workflow."
```
aidanns: Claude Reviewer: The regression check only scans `lefthook.yml` and `.github/workflows/*.yml` for `keep-sorted --mode=`, but the actual production invocation lives in `scripts/lint.sh` (run in CI via `task check`). If someone removes the lefthook entry but keeps `scripts/lint.sh`, keep-sorted is still enforced in CI, yet this check fails — the error message "not invoked in lefthook.yml pre-commit or any .github/workflows/*.yml" would be inaccurate. Consider either (a) also grepping `scripts/lint.sh` (or any script referenced from the workflow), or (b) narrowing the header comment on line 23-26 and the error message on line 212 to claim only what is actually checked ("wired into lefthook.yml or a CI workflow directly").

lefthook.yml
```diff
@@ -1,13 +1,22 @@
...
 # Install with `task install-hooks` (wraps `lefthook install`).
-# shfmt formatting options live in .editorconfig.
+# shfmt formatting options live in .editorconfig;
+# mdformat reads .mdformat.toml; taplo reads taplo.toml.
 
 pre-commit:
   parallel: true
   commands:
+    keep-sorted:
+      run: keep-sorted --mode=lint {staged_files}
```
aidanns, line 14: Claude Reviewer: The `keep-sorted` hook has no `glob` filter, unlike the sibling `mdformat`, `shellcheck`, `shfmt`, and `taplo` hooks. `{staged_files}` expands to *all* staged files, so keep-sorted is invoked on every staged path (binary, image, anything). It works today because the repo is all text, but `keep-sorted --mode=lint` with zero filenames errors out (`FTL error="must pass one or more filenames"`) — lefthook normally skips commands with empty substitution, but the absent `glob` makes the behaviour subtly different from the other commands. Suggest either adding `glob: "**"` (or a narrower glob covering only annotated file types) for consistency, or dropping a short comment explaining why keep-sorted is intentionally unfiltered.

scripts/format.sh
```diff
@@ -23,28 +27,55 @@ for arg in "$@"; do
...
-mapfile -t shell_files <<<"${shell_files_raw}"
+format_group() {
+  local label="$1"
+  local files_raw="$2"
+  shift 2
+  if [[ -z "${files_raw}" ]]; then
+    echo "task format: no ${label} files tracked; skipping."
+    return 0
+  fi
+  mapfile -t files <<<"${files_raw}"
```
aidanns, line 70: Claude Reviewer: `mapfile -t files` inside `format_group` is not declared `local`, so `files` leaks into the caller's scope. Each subsequent call to `format_group` will inherit the previous call's array (and mapfile will overwrite it, so behaviourally it is fine today), but the leak is surprising and one refactor away from a subtle bug. Add `local files` alongside the existing `local label`, `local files_raw` declarations.

scripts/verify-standards.sh
aidanns: Claude: Done — extended the check to also grep `scripts/*.sh`, so removing the `keep-sorted --mode=lint` invocation from `scripts/lint.sh` (the real CI path) now fails `task verify-standards`.

lefthook.yml
aidanns: Claude: Done — confirmed `keep-sorted` errors with `must pass one or more filenames` on zero args. Added `glob: "{*.toml,*.sh,*.yml,*.yaml,*.md,*.py,.gitignore}"` covering every currently-annotated file type plus common future ones. CI still scans every tracked file via `scripts/lint.sh`, so annotations in uncovered types remain caught there.

scripts/format.sh
aidanns: Claude: Done!

PR #81, opened 2026-04-19
ci: extract shared workflow setup into a composite action

### LINKED ISSUE #79: ci: extract shared workflow setup into a composite action
## Problem

All five workflows under `.github/workflows/` duplicate the same ~40-line setup block — only the final `task <target>` step differs. The shared prefix is:

- `actions/checkout@v6`
- `actions/setup-python@v6` (3.12)
- Install `d2`
- Install `systems-engineering`
- Install `go-task` (arduino/setup-task@v2)
- Install `shellcheck` (pinned via `SHELLCHECK_VERSION` env var)
- Install `shfmt` (pinned via `SHFMT_VERSION` env var)
- Run `task verify-dependencies`

Affected files:
- `.github/workflows/check.yml`
- `.github/workflows/test.yml`
- `.github/workflows/verify-design.yml`
- `.github/workflows/verify-function-tests.yml`
- `.github/workflows/verify-standards.yml`

Every toolchain change (e.g. bumping \`SHELLCHECK_VERSION\`, upgrading \`setup-python\`, adding a new tool) has to be made in five places and kept in sync by hand. This has already caused minor drift (e.g. \`check.yml\` prints \`shellcheck --version\` / \`shfmt --version\` after install; the others don't).

## Expected behaviour

Each workflow should declare its trigger, then reduce to:

\`\`\`yaml
steps:
  - uses: actions/checkout@v6
  - uses: ./.github/actions/setup-toolchain
  - run: task <workflow-specific-target>
\`\`\`

All tool installation, version pinning, and \`verify-dependencies\` verification should live in one place.

## Proposed approach

Extract a **composite action** at \`.github/actions/setup-toolchain/action.yml\` containing the shared steps. Composite actions are the right fit here because the reuse is step-level (inside an existing job), not job-level — a reusable workflow would force each caller into a separate job with extra checkout/artifact plumbing.

Composite action should:
- Own the \`SHELLCHECK_VERSION\` and \`SHFMT_VERSION\` pins (either as action-level defaults or as inputs with defaults).
- Expose Python version as an input (defaulting to \`3.12\`) so future workflows can override if needed.
- Run \`task verify-dependencies\` as its final step so every caller gets the same preflight check.
- Normalise the minor drift between workflows (consistently print \`--version\` for installed tools, or consistently omit it).

## Scope

- Add \`.github/actions/setup-toolchain/action.yml\`.
- Update all five workflows to use it.
- Remove per-workflow \`env:\` blocks that only existed to feed the duplicated install steps.

## Out of scope

- Cross-repo sharing (e.g. publishing the action to a central \`aidanns/actions\` repo). Can be considered later if other repos grow the same pattern.
- Converting any workflow into a reusable workflow (\`on: workflow_call\`). The reuse boundary here is steps, not jobs.

aidanns:
## Summary

- Adds `.github/actions/setup-toolchain/action.yml` — a composite action owning all tool installs (uv, d2, systems-engineering, go-task, shellcheck, shfmt, mdformat, taplo, keep-sorted) and `task verify-dependencies`
- Reduces all five workflows to: checkout → setup-toolchain → workflow-specific task
- Version pins are now inputs with defaults, so a single edit updates all workflows; mdformat uses `uv tool install` for consistency with the project's uv adoption

Closes #79

## Test plan

- [x] All five CI workflows pass on this PR
- [x] Verify each workflow runs its correct task (`task check`, `task test -- --all`, `task verify-design`, `task verify-function-tests`, `task verify-standards`)

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -26,6 +26,9 @@ inputs:
   taplo-version:
     description: taplo version to install
     default: "0.10.0"
+  systems-engineering-ref:
+    description: Commit SHA or ref of aidanns/systems-engineering to install
+    default: "d8e664a5ad20c071d4ec1cd62377597f9404b6e9"
   github-token:
     description: GitHub token to avoid rate limits when installing go-task
     required: false
@@ -44,8 +47,10 @@ runs:
 
     - name: Install systems-engineering
       shell: bash
+      env:
+        SYSTEMS_ENGINEERING_REF: ${{ inputs.systems-engineering-ref }}
       run: |
-        curl -fsSL https://raw.githubusercontent.com/aidanns/systems-engineering/main/install.sh | bash
+        curl -fsSL "https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh" | bash
         echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
     - name: Install go-task
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -1,0 +1,114 @@
+name: Setup Toolchain
+description: Install all CI tooling and verify dependencies
+
+inputs:
+  go-task-version:
+    description: go-task version to install
+    default: "3.50.0"
+  keep-sorted-version:
+    description: keep-sorted version to install
+    default: "0.8.0"
+  mdformat-version:
+    description: mdformat version to install
+    default: "0.7.22"
+  mdformat-gfm-version:
+    description: mdformat-gfm version to install
+    default: "1.0.0"
+  mdformat-tables-version:
+    description: mdformat-tables version to install
+    default: "1.0.0"
+  shellcheck-version:
+    description: shellcheck version to install
+    default: "0.10.0"
+  shfmt-version:
+    description: shfmt version to install
+    default: "3.8.0"
+  taplo-version:
+    description: taplo version to install
+    default: "0.10.0"
+  github-token:
+    description: GitHub token to avoid rate limits when installing go-task
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
+      with:
+        enable-cache: true
+
+    - name: Install d2
+      shell: bash
+      run: curl -fsSL https://d2lang.com/install.sh | sh -s --
+
+    - name: Install systems-engineering
+      shell: bash
+      run: |
+        curl -fsSL https://raw.githubusercontent.com/aidanns/systems-engineering/main/install.sh | bash
+        echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+    - name: Install go-task
+      uses: arduino/setup-task@v2
+      with:
+        version: ${{ inputs.go-task-version }}
+        repo-token: ${{ inputs.github-token }}
+
+    - name: Install shellcheck
+      shell: bash
+      env:
+        SHELLCHECK_VERSION: ${{ inputs.shellcheck-version }}
+      run: |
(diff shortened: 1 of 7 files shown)


aidanns (commented):
Claude Reviewer: Overall, this is a clean refactor — the DRY improvement is clear and the composite action is well-structured. I have a few security and correctness concerns worth addressing before merge.

**Security: curl-piped scripts lack integrity verification**

Both the `d2` and `systems-engineering` install steps use the `curl | sh` pattern with no checksum or signature verification:

```yaml
run: curl -fsSL https://d2lang.com/install.sh | sh -s --
```

```yaml
run: curl -fsSL https://raw.githubusercontent.com/aidanns/systems-engineering/main/install.sh | bash
```

If either upstream URL is compromised or redirected, arbitrary code runs with the CI job's permissions. For d2, prefer downloading a pinned release tarball and verifying its SHA256 (the d2 GitHub releases page publishes checksums). For the systems-engineering script, pin to a specific commit SHA rather than `main`:
```
https://raw.githubusercontent.com/aidanns/systems-engineering/<COMMIT_SHA>/install.sh
```
This is a private repo so the risk is lower, but it also means a force-push to `main` silently changes what gets installed in CI.

**Security: systems-engineering installed from an unpinned branch**

The `systems-engineering` step fetches from `main` with no input to override the ref and no default SHA pinned. Any push to `main` in that repo changes what every CI run installs, with no diff in this repo. Add a `systems-engineering-ref` input (defaulting to a pinned commit SHA) so updates are explicit and auditable.

**Correctness: `actions/checkout@v6` does not exist**

All five workflows reference `actions/checkout@v6` but the latest released major version is v4. This PR does not change that line, but since the composite action refactor is touching these files, it is worth flagging: if v6 is intentional as a future-pinning placeholder, it will fail at runtime. The safe default is `actions/checkout@v4`.

**Maintainability: uv version is unpinned**

The `astral-sh/setup-uv@v5` step does not set a `uv-version` input. The uv project moves quickly; an unexpected uv upgrade between CI runs can silently change tool resolution behaviour. Consider adding a `uv-version` input to the composite action and threading it through to `setup-uv`:
```yaml
inputs:
  uv-version:
    description: uv version to install
    default: "0.6.14"  # pin to known-good version
```

**Minor: `install.sh` temp file left behind in systems-engineering step**

The old workflow downloaded to a named file before running it; the new composite action pipes directly (good), but as a side note, the old approach left an `install.sh` file in the working directory. The new piped form avoids that — worth noting this is an improvement that snuck in via the refactor.

.github/actions/setup-toolchain/action.yml
```diff
@@ -0,0 +1,114 @@
...
+    description: mdformat version to install
+    default: "0.7.22"
+  mdformat-gfm-version:
+    description: mdformat-gfm version to install
+    default: "1.0.0"
+  mdformat-tables-version:
+    description: mdformat-tables version to install
+    default: "1.0.0"
+  shellcheck-version:
+    description: shellcheck version to install
```
aidanns, line 21: Claude Reviewer: This step fetches and executes a shell script from `main` with no integrity check. Pin to a specific commit SHA (e.g. `https://raw.githubusercontent.com/aidanns/systems-engineering/<SHA>/install.sh`) and add a `systems-engineering-ref` input so future updates are explicit and auditable in diffs.

aidanns (commented):
Claude Reviewer: This step fetches and executes a shell script from `main` with no integrity check. Pin to a specific commit SHA (e.g. `/aidanns/systems-engineering/<SHA>/install.sh`) and add a `systems-engineering-ref` input so updates are visible in this repo's diff. Also consider downloading first and verifying a checksum before executing.

.github/actions/setup-toolchain/action.yml
aidanns: Claude: Done! Added a `systems-engineering-ref` input defaulting to the current HEAD SHA (`d8e664a5`). Future bumps will be explicit diffs.

PR #82, opened 2026-04-19
ci: add pip-audit dependency vulnerability scanning

### LINKED ISSUE #50: Add pip-audit dependency vulnerability scanning in CI
Source: `.claude/instructions/python.md` — pip-audit.

No dependency vulnerability scanning is configured. Standard: `pip-audit` (or `safety`) in CI.

**Acceptance:**
- CI workflow runs `pip-audit` (or equivalent against the `uv.lock` once #uv-migration lands) on every PR and on a schedule against `main`.
- Known-acceptable advisories documented in an ignore list with rationale.
- Workflow fails on new advisories.

- `scripts/verify-standards.sh` asserts at least one `.github/workflows/*.yml` file invokes `pip-audit` (or the documented equivalent).

aidanns:
Closes #50.

## Summary

- Adds `.github/workflows/security.yml` that runs `pip-audit` on every PR, push to `main`, and daily at 06:00 UTC
- Adds `pip-audit>=2.7` as a dev dependency (resolved into `uv.lock`)
- Adds `[tool.pip-audit]` section to `pyproject.toml` as the documented mechanism for recording known-acceptable advisories with rationale (PYSEC and GHSA examples included)
- Adds `task pip-audit` target for local discoverability
- Updates `verify-standards.sh` to assert a workflow invokes `pip-audit`

## Test plan

- [x] `bash scripts/verify-standards.sh` passes
- [x] `shellcheck scripts/verify-standards.sh` passes
- [x] CI checks pass on this PR

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -23,4 +23,4 @@ jobs:
       - name: Run pip-audit
         # --extra dev intentionally covers dev deps (e.g. testcontainers) —
         # a vulnerable dev dep can still be exploited in CI.
-        run: uv run --extra dev pip-audit --strict
+        run: uv run --extra dev pip-audit --strict --skip-editable
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -21,6 +21,4 @@ jobs:
           enable-cache: true
 
       - name: Run pip-audit
-        # Audit all packages resolved from uv.lock (production + dev deps).
-        # pip-audit fails the step on any unignored advisory.
-        run: uv run --extra dev pip-audit
+        run: uv run --with pip-audit pip-audit
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -21,6 +21,10 @@ jobs:
           enable-cache: true
 
       - name: Run pip-audit
-        # --extra dev intentionally covers dev deps (e.g. testcontainers) —
-        # a vulnerable dev dep can still be exploited in CI.
-        run: uv run --extra dev pip-audit --strict --skip-editable
+        # Export all third-party deps (including dev, excluding the local
+        # project package) then audit them. --extra dev intentionally covers
+        # dev deps — a vulnerable dev dep can still be exploited in CI.
+        run: |
+          uv export --extra dev --no-emit-project --format requirements-txt --no-hashes \
+            > /tmp/requirements.txt
+          uvx pip-audit --strict -r /tmp/requirements.txt
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,26 @@
+name: Security
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    # Run daily at 06:00 UTC to catch newly-published advisories on main.
+    - cron: "0 6 * * *"
+
+jobs:
+  pip-audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+
+      - name: Run pip-audit
+        # Audit all packages resolved from uv.lock (production + dev deps).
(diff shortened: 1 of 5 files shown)


aidanns (commented):
Claude Reviewer: Overall this is a solid, well-structured PR. Here are my findings across the five focus areas:

---

**1. Correctness of the pip-audit invocation**

The invocation `uv run --extra dev pip-audit` is correct and will pick up `[tool.pip-audit]` configuration from `pyproject.toml` automatically (pip-audit reads that section natively). No issues here.

---

**2. Security coverage completeness**

The workflow audits the `--extra dev` dependency set. This means **dev-only dependencies** (e.g. `testcontainers`, `build`) are scanned in addition to production ones. That is actually good — a vulnerable dev dep can still be exploited in CI — but it is worth confirming this is intentional rather than accidental. A comment in the workflow step noting this would help future readers understand the scope.

---

**3. The ignore list mechanism in pyproject.toml**

The `ignore-vuln = []` entry and the accompanying comment are helpful. One gap: pip-audit's `--ignore-vuln` flag accepts identifiers from multiple advisory databases (OSV, PyPA, GHSA). The example comment only shows the `PYSEC-` prefix. It would reduce friction for future contributors to show that GHSA identifiers are also accepted, e.g.:

```toml
# Examples of acceptable ignore entries:
#   "PYSEC-2024-12345"  # Not reachable via our usage of <package>
#   "GHSA-xxxx-yyyy-zzzz"  # Not reachable via our usage of <package>
```

This is minor but could save debugging time when the first real advisory is encountered.

---

**4. verify-standards.sh robustness**

The new check reuses `${workflows_stripped}` (which strips comments before grepping), consistent with the existing pattern. One concern: the check only verifies that the literal string `pip-audit` appears somewhere in a workflow file. It does not verify it runs in a job that is gated on PRs. If someone were to comment out the invocation inside a YAML comment block, `strip_comments` would correctly exclude it — but if someone wraps the step in a condition like `if: false`, the check would still pass. This is a known limitation of the text-grep approach used throughout this file, so it is consistent with the existing standard rather than a regression introduced by this PR.

A more significant issue: the check uses `<<<<"${workflows_stripped}"` but `workflows_stripped` is populated by `find .github/workflows -name '*.yml'`. The new workflow file is named `security.yml` with a `.yml` extension, so it will be included. However, if someone creates a `security.yaml` (with `.yaml` extension), it would be missed. Looking at line 160 of verify-standards.sh, only `*.yml` files are collected into `workflows_stripped`. Line 113 checks for both `.yml` and `.yaml` for existence, but the content collection on line 160 only covers `.yml`. This pre-existing inconsistency means a `.yaml` workflow containing a `pip-audit` invocation would satisfy the existence check but not the content check — the opposite of the problem here, but worth noting as a latent bug.

---

**5. Missing edge cases**

- **No `--strict` or `--require-hashes` flag**: pip-audit by default only checks for known CVEs, not hash integrity. For a project that pins everything in `uv.lock` (which it does), passing `--require-hashes` would provide supply-chain integrity checking in addition to CVE scanning. This is an enhancement, not a bug.

- **No `--output` / SARIF upload**: The workflow currently prints results to stdout only. For GitHub's Security tab integration, uploading a SARIF report via `github/codeql-action/upload-sarif` would make vulnerabilities visible in the PR diff view. pip-audit supports `--format sarif`. This is an enhancement worth a follow-up issue.

- **Scheduled run notification**: The daily `schedule:` trigger will fail silently on `main` if a new advisory is published. There is no mechanism to notify the team (e.g. Slack, GitHub issue creation). This is out of scope for this PR but worth tracking.

- **No pinning of `astral-sh/setup-uv@v5`**: The other workflows appear to use the same floating tag, so this is consistent, but using a SHA pin for third-party actions is generally recommended for supply-chain security in security-sensitive workflows.

aidanns (commented):
Claude: Done! Added a GHSA example alongside the PYSEC one in the ignore list comment.

aidanns:
Claude: Done! Added a comment to the workflow step explaining that dev deps are intentionally in scope.

PR #89, opened 2026-04-19
test: extend Docker per-test integration pattern to things-* services

### LINKED ISSUE #78: Add Docker-based e2e tests for things-bridge, things-cli, and things-client-cli-applescript
## Summary

`agent-auth` now has Docker-based integration tests (see ADR
`design/decisions/0004-docker-integration-tests.md` and PR #67) that
spin up a per-test container via `testcontainers.compose.DockerCompose`,
bind-mount a test `config.json`, and exercise the service strictly
through its public HTTP + CLI surface.

The other services in this repo still rely on in-process tests or host
shelling out to `osascript`, which have the same host-port / shared-state
collision risk that motivated the `agent-auth` migration, and they do
not catch regressions in the compose / CLI / config wiring.

## What to migrate

- `things-bridge` — HTTP bridge; already has a `--fake-things` mode that
  avoids AppleScript, so it can run in Linux containers end-to-end.
- `things-cli` — client for `things-bridge`; should run paired with a
  dockerised `things-bridge` and talk to it over HTTP.
- `things-client-cli-applescript` — AppleScript-dependent, so a true
  container-based e2e requires a host path; the in-repo layer can at
  least gain a Dockerised contract test against a recorded-fixture
  backend.

## Definition of done

- Each service gets a `docker/Dockerfile.test` + `docker/compose.test.yaml`
  (co-located with the existing `agent-auth` ones) and a pytest
  `conftest.py` that uses `testcontainers-python` to drive the Compose
  lifecycle.
- Per-test config is supplied via bind-mounted file, not env-var
  materialisation in the Dockerfile entrypoint (mirrors the agent-auth
  approach).
- `scripts/verify-integration-isolation.sh` (or a per-service analogue)
  prevents regressions back to raw `127.0.0.1` references in the test
  modules.
- Existing in-process integration tests for those services are retired
  (kept as unit tests where they still pull their weight).
- ADRs extended (or new ADRs added) describing the migration per
  service.

## References

- PR introducing the pattern for agent-auth: #67
- ADR: `design/decisions/0004-docker-integration-tests.md`
- Issue that motivated the agent-auth migration: #7

aidanns:
## Summary

- Migrates `things-bridge`, `things-cli`, and `things-client-cli-applescript` integration suites onto the per-test Docker Compose pattern established for `agent-auth` in #67 / ADR 0004.
- Single shared `docker/Dockerfile.test` runs every console-script; one shared `docker/docker-compose.yaml` declares the `agent-auth` + `things-bridge` pair on an internal network with only host-mapped loopback ports. Each per-test fixture renders the compose template with `{{ VAR }}` placeholder substitution in Python before docker compose ever sees it — the file is fully self-contained per test (no env-var inheritance).
- Inlines the bridge's baseline config under a compose `configs:` block rather than a separate file (no more `docker/config.test.things-bridge.yaml`).
- Adds `GET /things-bridge/health`, authenticated under a `things-bridge:health` scope to mirror `GET /agent-auth/health`. The Compose readiness probe accepts 401/403 as a positive "server is up" signal.
- Retires `tests/test_things_bridge_e2e.py` and `tests/test_things_client_cli_contract.py`; error-mapping coverage retained at the unit layer in `tests/test_things_bridge_server.py`.
- ADR 0005 documents the decision; `scripts/verify-integration-isolation.sh` extended to enforce the new layout.

Closes #78.

## Test plan

- [x] `task check` passes
- [x] `uv run pytest tests/ --ignore=tests/integration` (unit suite) passes
- [x] `uv run pytest tests/integration/things_bridge/` passes against Docker (CI run 24621963241)
- [x] `uv run pytest tests/integration/things_cli/` passes against Docker (CI run 24621963241)
- [x] `uv run pytest tests/integration/things_client_applescript/` passes against Docker (CI run 24621963241)
- [x] `bash scripts/verify-integration-isolation.sh` passes


tests/integration/_support.py
```diff
@@ -0,0 +1,147 @@
...
+
+def wait_until_server_ready(
+    health_url: str,
+    *,
+    accept_status: tuple[int, ...] = (401, 403),
+) -> None:
+    """Block until ``health_url`` answers, treating ``accept_status`` as up.
+
+    The agent-auth health endpoint requires an ``agent-auth:health``
+    token, so an unauthenticated probe returns ``401``. The fixture
```
aidanns: Claude Reviewer: Stale comment after the bridge-health-auth change in this same commit. The docstring still claims "The things-bridge health endpoint is unauthenticated, so any 2xx response is up." — but as of this commit `/things-bridge/health` requires the `things-bridge:health` scope, and the bridge fixture explicitly probes it with `accept_status=(401, 403)`. Suggest updating to: both endpoints require a scope, so an unauthenticated probe gets 401 (or 403 if the scope check ran), which the fixtures treat as a positive "server is up" signal.

tests/integration/_support.py
```diff
@@ -0,0 +1,147 @@
...
+import urllib.error
+import urllib.request
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+DOCKER_DIR = REPO_ROOT / "docker"
+DOCKERFILE = DOCKER_DIR / "Dockerfile.test"
+COMPOSE_TEMPLATE = DOCKER_DIR / "docker-compose.yaml"
+COMPOSE_FILE_NAME = "docker-compose.yaml"
+_PLACEHOLDER_PATTERN = re.compile(r"\{\{\s*([A-Z0-9_]+)\s*\}\}")
```
aidanns: Claude Reviewer: Minor robustness gap: `[A-Z0-9_]+` only catches uppercase placeholders, so the leftover-detection guard silently misses any future placeholder written as `{{ project_name }}` or `{{ MyVar }}` (typo'd casing). The first guard ("placeholder not found") only fires when a *caller* passes the wrong key; a typo in the *template* itself would slip through both checks. Consider broadening to `[A-Za-z][A-Za-z0-9_]*` (or document the uppercase convention explicitly in the template comment up top).

src/things_bridge/server.py
```diff
@@ -71,14 +72,14 @@ def _extract_bearer(self) -> str | None:
             return None
         return header[7:].strip() or None
 
-    def _validate(self, token: str, description: str) -> bool:
+    def _validate(self, token: str, description: str, scope: str = READ_SCOPE) -> bool:
```
aidanns: Claude Reviewer: Default scope is convenient for the read endpoints (which dominate call sites), but it makes the call-site grep harder — every `_validate(token, "...")` invocation now silently means `things:read`, and the only place that does *not* mean read (the new health route) is the only place that passes the kwarg. If you ever add a third scope, it's easy to forget the `scope=` kwarg and silently get a `things:read` validation. Consider making `scope` a required positional param so every call site states its intent explicitly. Optional / judgment call.

aidanns:
Claude Reviewer: Independent review of commit 6910515.

**Overall**: solid refactor; the per-test-rendered compose pattern is a clear improvement over env-var inheritance, the inline `configs:` block correctly retires the standalone `config.test.things-bridge.yaml`, and the bridge health auth change is consistent with `agent-auth/health`. The new health-endpoint tests cover the three meaningful states (no token, wrong scope, right scope).

**Spot checks**:
- Placeholder substitution: `template.replace()` ordering is safe in practice — values are paths/UUIDs/image tags with no embedded `{{ }}` markup.
- Both KeyError checks in `render_compose_file` are useful (one catches caller typos, one catches forgotten substitutions). Not redundant.
- Probe accepting 401/403 is a reasonable indirection — the path-handler returns 401 *before* hitting authz when no bearer is present, so the probe verifies the HTTP server is up but doesn't reach into authz reachability. Same as the old unauthenticated probe — net no regression.
- `description="things-bridge health check"` only surfaces in the JIT prompt (not the audit log), and only if a future policy applied prompt-tier to `things-bridge:health`. Operationally fine.
- Compose `configs:` with `content:` requires compose v2.23.1+ (Nov 2023); GitHub `ubuntu-latest` runners ship v2.27+. Safe.

**Inline comments** filed for: stale docstring in `_support.py`, regex tightness, default-scope ergonomics in `_validate`.

**Splitting**: I'd lean toward saying the templating + auth changes *could* have been split into two commits — they're logically independent (the only coupling is that the inline `configs:` content references the same `auth_url`/`port` regardless of auth scope). Not a blocker; the commit message already calls out the two changes.

Recommendation: merge after addressing the stale docstring; the other two are optional polish.

tests/integration/_support.py
aidanns: Claude: Done in 2c7b690 — docstring now reads that both endpoints require a scope and the unauthenticated probe gets 401/403 as the up signal.

tests/integration/_support.py
aidanns: Claude: Done in d86ec9c — broadened to `[A-Za-z][A-Za-z0-9_]*`, so a typo'd-case placeholder in the template now trips the leftover guard.

src/things_bridge/server.py
aidanns: Claude: Done in d86ec9c — `scope` is now a required positional and every call site (read endpoints + the new health route) states it explicitly.

PR #92, opened 2026-04-19
docs: add commit-signing section to CONTRIBUTING.md; gate in verify-standards

### LINKED ISSUE #13: Add CONTRIBUTING.md with dev setup, test, and release procedures
Source: `.claude/instructions/release-and-hygiene.md` — Required project files.

The repo does not yet have a `CONTRIBUTING.md`. Standard requires it to document dev setup, testing, release cutting, and signing procedures.

**Acceptance:**
- `CONTRIBUTING.md` added at repo root covering: dev environment setup, running tests, release process (ties in with #release-automation), and commit signing.

- `scripts/verify-standards.sh` asserts `CONTRIBUTING.md` exists and contains dev-setup, testing, release, and commit-signing sections.

aidanns:
## Summary

- Adds `## Commit signing` section to `CONTRIBUTING.md` covering GPG and SSH signing configuration, fulfilling the requirement in `.claude/instructions/release-and-hygiene.md`
- Extends `scripts/verify-standards.sh` to assert `CONTRIBUTING.md` exists and contains dev-setup, testing, release, and commit-signing sections
- Uses bash 3.2-compatible parallel arrays (not `declare -A`) so the script works on macOS system bash

Closes #13

## Test plan

- [x] `bash scripts/verify-standards.sh` passes
- [x] Temporarily rename `CONTRIBUTING.md` and verify the check fails with a clear error
- [x] Temporarily remove `## Commit signing` and verify the check fails identifying that section

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,9 @@ Use conventional commit messages (`feat:`, `fix:`, `docs:`, `chore:`,
 is `main`; feature branches follow `aidanns/<feature-name>`. See the
 project [CLAUDE.md](CLAUDE.md) for the full working agreement.
 
+See also: [Commit signing](#commit-signing) — commit message format and GPG/SSH
+signing are sibling requirements, not the same thing.
+
 ## Release process
 
 `task release` is the release entrypoint. The underlying automation
@@ -80,7 +83,10 @@ releases that would skip the standard checks.
 
 ## Commit signing
 
-Commits to `main` must be signed. Configure GPG or SSH signing in git:
+Commits to `main` must be signed. Enforcement via GitHub branch protection is
+not yet wired up (tracked in
+[#93](https://github.com/aidanns/agent-auth/issues/93)); until then this is an
+honour-system requirement. Configure GPG or SSH signing in git:
 
 ```bash
 # GPG (recommended)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,3 +77,23 @@ project [CLAUDE.md](CLAUDE.md) for the full working agreement.
 [#18](https://github.com/aidanns/agent-auth/issues/18) and is not yet
 implemented — running the task today exits non-zero to prevent manual
 releases that would skip the standard checks.
+
+## Commit signing
+
+Commits to `main` must be signed. Configure GPG or SSH signing in git:
+
+```bash
+# GPG (recommended)
+git config --global commit.gpgsign true
+git config --global user.signingkey <YOUR_KEY_ID>
+
+# SSH (alternative — set these instead of the GPG options above)
+git config --global gpg.format ssh
+git config --global user.signingkey ~/.ssh/id_ed25519.pub
+```
+
+Verify signing is working:
+
+```bash
+git log --show-signature -1
+```
--- a/scripts/verify-standards.sh
+++ b/scripts/verify-standards.sh
@@ -29,6 +29,9 @@
 #      `pip install` to bootstrap a venv.
 #   6. ruff is configured in pyproject.toml, wired into treefmt and
 #      lefthook, and gated in CI per .claude/instructions/python.md.
+#   7. CONTRIBUTING.md exists at repo root and contains dev-setup, testing,
+#      release, and commit-signing sections per
+#      .claude/instructions/release-and-hygiene.md.
 
 set -euo pipefail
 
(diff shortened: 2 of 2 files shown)


scripts/verify-standards.sh
```diff
@@ -323,3 +326,50 @@ if [[ ${ruff_missing} -ne 0 ]]; then
...
+  # Parallel arrays (bash 3.2-compatible): section_names[i] pairs with section_patterns[i].
+  section_names=(
+    dev-setup
+    testing
+    release
+    commit-signing
+  )
+  section_patterns=(
+    "## Dev setup|## Development environment setup|## Getting started|## Setup"
+    "## Running tasks|## Testing|## Running tests"
```
aidanns, line 368: Claude Reviewer: The `grep -E` pattern matching is case-sensitive, which makes the check fragile. For example, `## Dev Setup` (capital S) would not match `## Dev setup` in the pattern. Consider using `grep -qiE` (add `-i`) for case-insensitive matching, or document that section headings must use the exact casing in the patterns.

This is particularly relevant for the `dev-setup` pattern which lists four alternatives (`## Dev setup|## Development environment setup|## Getting started|## Setup`) — if a future contributor writes `## Dev Setup` the check silently fails to match.

scripts/verify-standards.sh
```diff
@@ -323,3 +326,50 @@ if [[ ${ruff_missing} -ne 0 ]]; then
...
+  echo "  $2" >&2
+  contributing_missing=1
+}
+
+if [[ ! -f CONTRIBUTING.md ]]; then
+  fail_contributing_check \
+    "CONTRIBUTING.md is missing from the repo root." \
+    "Add CONTRIBUTING.md covering dev setup, testing, release, and commit signing."
+else
+  contributing_content="$(cat CONTRIBUTING.md)"
```
aidanns: Claude Reviewer: The `cat CONTRIBUTING.md` into a shell variable and then passing it via herestring (`<<<`) is unnecessary indirection. Since the script has already `cd`'d to `REPO_ROOT`, you can grep the file directly:

```bash
if ! grep -qE "${section_patterns[${i}]}" CONTRIBUTING.md; then
```

This avoids slurping the whole file into memory and eliminates the `contributing_content` variable entirely. It also avoids a subtle correctness risk: if CONTRIBUTING.md contains a trailing newline followed by null bytes or very unusual content, herestring expansion can behave unexpectedly in some bash versions.

CONTRIBUTING.md
```diff
@@ -77,3 +77,23 @@ project [CLAUDE.md](CLAUDE.md) for the full working agreement.
...
+# SSH (alternative — set these instead of the GPG options above)
+git config --global gpg.format ssh
+git config --global user.signingkey ~/.ssh/id_ed25519.pub
+```
+
+Verify signing is working:
+
+```bash
+git log --show-signature -1
+```
```
aidanns, line 105: Claude Reviewer: The section heading "## Commit signing" is checked by `verify-standards.sh`, but the existing CONTRIBUTING.md already has a "## Commit conventions" section. These are different sections covering different topics (signing vs. message format), which is correct — but it is worth confirming there is no risk of the new `## Commit signing` section being confused with or conceptually merged into `## Commit conventions` by a future editor who removes one as a duplicate.

The two sections could benefit from a brief cross-reference: e.g. add a line at the end of `## Commit conventions` like "See also: [Commit signing](#commit-signing)" to make it clear they are siblings, not duplicates.

CONTRIBUTING.md
```diff
@@ -77,3 +77,23 @@ project [CLAUDE.md](CLAUDE.md) for the full working agreement.
 [#18](https://github.com/aidanns/agent-auth/issues/18) and is not yet
 implemented — running the task today exits non-zero to prevent manual
 releases that would skip the standard checks.
+
+## Commit signing
```
aidanns, line 84: Claude Reviewer: The section states "Commits to `main` must be signed" but does not describe how this is enforced. If enforcement is via GitHub branch protection rules, that should be noted so contributors know the check isn't purely advisory. If it is not yet enforced by branch protection (i.e., this is documentation-only for now), calling that out explicitly avoids confusion when unsigned commits are accepted without error.

Either add a sentence like "This is enforced by branch protection on `main`" (if true), or "Enforcement via branch protection is tracked in issue #N" (if not yet wired up).

scripts/verify-standards.sh
```diff
@@ -323,3 +326,50 @@ if [[ ${ruff_missing} -ne 0 ]]; then
...
+
+  # Parallel arrays (bash 3.2-compatible): section_names[i] pairs with section_patterns[i].
+  section_names=(
+    dev-setup
+    testing
+    release
+    commit-signing
+  )
+  section_patterns=(
+    "## Dev setup|## Development environment setup|## Getting started|## Setup"
```
aidanns, line 364: Claude Reviewer: The `section_patterns` array uses pipe (`|`) as ERE alternation inside double-quoted strings, which works correctly with `grep -E`. However, the pattern for `testing` is:

```
"## Running tasks|## Testing|## Running tests"
```

The current CONTRIBUTING.md uses `## Running tasks` — which matches fine today. But this pattern is named `testing` while the primary match is `## Running tasks`, which is ambiguous: `Running tasks` covers the task catalogue (a superset of testing), not testing specifically. A future refactor that splits the file into a separate `## Testing` subsection and renames `## Running tasks` would cause the pattern to match via the wrong alternative.

Consider narrowing the intent: if the standard requires a section about how to run tests, the pattern should prefer `## Testing` or `## Running tests` as primaries. If it intentionally accepts `## Running tasks` because this project collapses both into one section, add a comment to that effect in the script.

scripts/verify-standards.sh
aidanns: Claude: Done! Changed `grep -qE` to `grep -qiE` for case-insensitive matching.

scripts/verify-standards.sh
aidanns: Claude: Done! Removed the `contributing_content` variable and grep now reads the file directly.

CONTRIBUTING.md
aidanns: Claude: Done! Added a "See also" cross-reference from `## Commit conventions` to `## Commit signing` to make clear they are sibling requirements.

CONTRIBUTING.md
aidanns: Claude: Done! Added a note that enforcement via branch protection is not yet wired up, with a link to the new tracking issue #93.

scripts/verify-standards.sh
aidanns: Claude: Done! Added an inline comment explaining that `## Running tasks` is intentionally accepted because this project collapses testing and the task catalogue into one section, and that the broader alternatives remain in case the two are split later.

Issue #23, opened 2026-04-17
Select and record cybersecurity standard in SECURITY.md


aidanns:
Source: `.claude/instructions/design.md` — Cybersecurity standard.

No cybersecurity standard has been selected or recorded. Standard requires a project-level choice (e.g. ISM, NIST SP 800-53) recorded in `SECURITY.md`, against which implementation plans verify compliance.

**Acceptance:**
- Cybersecurity standard chosen and recorded in `SECURITY.md` with rationale.
- Relevant controls enumerated (or linked) so plan review can walk them.
- Depends on #SECURITY.md-creation (tracked separately).

**Deterministic regression check:**
- `scripts/verify-standards.sh` asserts `SECURITY.md` contains a `Cybersecurity standard` section that names the chosen standard.

aidanns:
Claude: Addressed in #96 (commit 4bb02a3).

`SECURITY.md` now has a `## Cybersecurity standard` section adopting **NIST SP 800-53 Revision 5**, with rationale and the in-scope control families (AC, AU, IA, SC, SI) mapped to their implementing code paths, plus a detailed controls table.

`scripts/verify-standards.sh` asserts the section is present and passes.

PR #96, opened 2026-04-19
feat: add release artifacts — CHANGELOG, LICENSE, SECURITY, install.sh, release automation

### LINKED ISSUE #14: Add CHANGELOG.md using Keep-a-Changelog format
Source: `.claude/instructions/release-and-hygiene.md` — Required project files.

No `CHANGELOG.md` exists. Standard requires Keep-a-Changelog formatting paired with semantic versioning, and updates on every user-facing PR.

**Acceptance:**
- `CHANGELOG.md` added at repo root following the Keep-a-Changelog template.
- Past user-visible changes (agent-auth server + CLI in #4, things-bridge + things-cli in #9) back-filled under an initial `Unreleased` / `0.1.0` entry.
- `CONTRIBUTING.md` flags the CHANGELOG update as a required step on user-facing PRs.

- `scripts/verify-standards.sh` runs a keep-a-changelog structural check on `CHANGELOG.md` and asserts the `## [Unreleased]` section is present.

### LINKED ISSUE #15: Add LICENSE.md (MIT) and link from README
Source: `.claude/instructions/release-and-hygiene.md` — Required project files.

No license file is present. Standard default is MIT, with a link from the README's "License" section.

**Acceptance:**
- `LICENSE.md` added at repo root with MIT text and the appropriate copyright line.
- README gains a `## License` section linking to `LICENSE.md`.

- `scripts/verify-standards.sh` asserts `LICENSE.md` exists at the repo root and `README.md` links to it from a `## License` section.

### LINKED ISSUE #16: Add SECURITY.md with threat model and vulnerability reporting
Source: `.claude/instructions/release-and-hygiene.md` — Required project files; `.claude/instructions/plan-template.md` — Threat model.

No `SECURITY.md` exists. Standard requires it to document trust boundaries, threat model (STRIDE / attack-tree), key handling, revocation flow, audit surface, and vulnerability reporting.

**Acceptance:**
- `SECURITY.md` added covering all of the above for the agent-auth server, things-bridge, and things-cli.
- Threat model captures the plugin-in-process trust concern (tracked in #6) and the 127.0.0.1-only bind posture.
- Vulnerability-reporting contact documented.
- Linked from README's Security section.

This issue is a prerequisite for recording the chosen cybersecurity standard (tracked separately).

- `scripts/verify-standards.sh` asserts `SECURITY.md` exists and contains the required sections (trust boundaries, threat model, key handling, revocation flow, audit surface, vulnerability reporting).

### LINKED ISSUE #17: Add install.sh and document curl | bash install idiom
Source: `.claude/instructions/release-and-hygiene.md` — Release process.

agent-auth ships user-facing binaries (`agent-auth`, `things-bridge`, `things-cli`). Standard requires an `install.sh` at the repo root and documentation of the `curl -fsSL <url> | bash` idiom in the README.

**Acceptance:**
- `install.sh` at repo root handles a clean install of the three CLIs and their Python deps into a user-owned venv.
- README Installation section documents the one-liner.
- Script follows user global bash conventions (`#!/usr/bin/env bash`, `set -euo pipefail`, description comment surrounded by blank lines).

- `scripts/verify-standards.sh` asserts `install.sh` exists at the repo root, is executable, and passes `shellcheck` (dispatches to the check from #51).

### LINKED ISSUE #18: Automate release process as a task-runner target
Source: `.claude/instructions/release-and-hygiene.md` — Release process.

No release automation exists. Standard requires a task-runner target that handles version bumping, tagging, and posting the GitHub release, with the procedure documented in `CONTRIBUTING.md`.

**Acceptance:**
- Task-runner target (depends on adopting `go-task` — tracked separately) wires up: validate clean tree → bump/tag (`vX.Y.Z`) → push tag → create GitHub release from CHANGELOG entry.
- `CONTRIBUTING.md` documents invocation.
- Version still flows from VCS tags via `setuptools-scm` (already in place).

- `scripts/verify-standards.sh` asserts the release task is defined in the task runner and invokable (the task itself embeds the clean-tree / tag / release verification steps).

### LINKED ISSUE #19: Populate GitHub repository About metadata (topics, homepage)
Source: `.claude/instructions/release-and-hygiene.md` — Repository metadata.

Repository topics are currently null and homepage is empty. Standard requires populating description, homepage, and topics with a one-line description matching the README summary.

**Acceptance:**
- Repository "About" description matches README one-liner.
- Topics set (e.g. `authorization`, `ai-agents`, `macos`, `applescript`, `cli`).
- Homepage field populated (README / docs URL, if any).

- `scripts/verify-standards.sh` calls `gh api repos/{owner}/{repo}` (or `gh repo view --json`) and asserts the description, homepage, and topics fields are non-empty.

aidanns:
## Summary

- **CHANGELOG.md** (closes #14): Keep-a-Changelog format. v0.1.0 entries live under `[Unreleased]` until the first `task release` cuts the tag (nothing has been released yet)
- **LICENSE.md** (closes #15): MIT license; README gains `## License` section linking to it
- **SECURITY.md** (closes #16, #23): Full STRIDE threat model with NIST SP 800-30 Rev 1 risk matrix, trust boundaries, key handling, revocation flow, audit surface, vulnerability reporting via GitHub private advisories, and NIST SP 800-53 Rev 5 as the chosen cybersecurity standard
- **install.sh** (closes #17): `curl | bash` installer via `uv tool install` from the latest tagged release; README Installation section gains the one-liner
- **scripts/release.sh** (closes #18): Implemented — validates clean tree + branch sync, extracts CHANGELOG section, prompts, creates signed tag, pushes, creates GitHub release via `gh`. **Auto-derives the next version from Conventional Commits since the last `v*` tag** (`<type>!:` / `BREAKING CHANGE:` / `BREAKING-CHANGE:` → major, `feat:` → minor, `fix:` → patch; pass an explicit version to override). Walks with `--first-parent` (stable across squash- vs merge-commit PR policies). Reverts with BREAKING-CHANGE footers are handled (strip from the `This reverts commit` marker). While the current tag is in the `v0.*` range, major bumps are demoted to minor per SemVer 2.0.0 §4; pass `task release -- 1.0.0` to graduate.
- **scripts/lib/semver.sh**: pure `compute_bump` / `apply_bump` helpers extracted for testability
- **tests/test_release_semver.py**: 29 unit-test cases pinning down the bump logic (happy path, regex asymmetry, multi-line bodies, 0.x demotion, reverts, invalid-range failure, plus shellcheck/shfmt gates on the lib)
- **GitHub repo About** (closes #19): description, homepage, and topics populated via `gh repo edit`
- **CONTRIBUTING.md**: release section documents CHANGELOG update requirement, auto-version resolution (including the pre-v1.0.0 exception), and step-by-step procedure
- **scripts/verify-standards.sh**: new checks for all of the above
- **design/product_breakdown.yaml**: `install.sh` added as a deliverable

## Test plan

- [x] `task test` — full suite passes (237 passed, 3 skipped) including the 29 new `test_release_semver.py` cases
- [x] `task check` — lint + format + integration-isolation clean
- [x] `bash scripts/verify-standards.sh` passes all 14 checks
- [x] GitHub repo About metadata verified via `gh repo view --json description,homepageUrl,repositoryTopics`
- [x] CHANGELOG.md contains `## [Unreleased]` section with all 0.1.0 entries (no tag cut yet)
- [x] LICENSE.md present; README `## License` section links to it
- [x] SECURITY.md present with all 7 required sections and no author email (reports go through GitHub private advisories)
- [x] install.sh present and executable (`-x`)
- [x] `scripts/release.sh` auto-version end-to-end (fixture repos in `test_release_semver.py`): docs-only → `none`, +fix → `patch`, +feat → `minor`, +feat then +fix → `minor`, `feat!:` → `major`, `BREAKING CHANGE:` footer → `major`, `BREAKING-CHANGE:` footer → `major`, pure revert of BREAKING → `none`
- [x] `scripts/release.sh` pre-v1.0.0 demotion: v0.1.0 + `feat!:` → `v0.2.0`; v0.9.5 + `feat!:` → `v0.10.0`; v1.2.3 + `feat!:` → `v2.0.0` (no demotion)
- [x] `scripts/release.sh` regex is lowercase-only: `Feat!:` / `FIX:` at v0.9.5 → `none`
- [x] `scripts/release.sh` fails loudly on invalid range instead of silently printing `none`
- [x] `scripts/release.sh` with no args in this repo exits with "no 'vX.Y.Z' tag found" guidance
- [x] `scripts/release.sh 1.2` fails format validation with MAJOR.MINOR.PATCH error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,31 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.1.0] - 2026-04-19
+
+### Added
+
+- **agent-auth server and CLI** — HTTP validation server (`agent-auth serve`) with full
+  token lifecycle management: create, list, modify, revoke, rotate. HMAC-SHA256 signed
+  tokens with AES-256-GCM field encryption and signing key held in the system keyring.
+  Three-tier scope model (allow / prompt / deny), JIT approval via pluggable notification
+  plugin, token families with refresh-token reuse detection, and audit logging.
+- **things-bridge** — HTTP bridge server (`things-bridge serve`) that delegates token
+  validation to agent-auth and exposes read-only Things 3 endpoints under
+  `/things-bridge/`. The bridge contains no Things 3 logic; it shells out to a configured
+  Things-client CLI per request.
+- **things-client-cli-applescript** — Standalone read-only CLI that talks to Things 3 via
+  `osascript` on macOS. Emits JSON on stdout; usable independently of things-bridge for
+  local debugging.
+- **things-cli** — Thin HTTP client for things-bridge that auto-refreshes/reissues tokens
+  via agent-auth. Stores credentials in the system keyring (falls back to a
+  `~/.config/things-cli/credentials.yaml` file when no keyring backend is available).
+
+[Unreleased]: https://github.com/aidanns/agent-auth/compare/v0.1.0...HEAD
+[0.1.0]: https://github.com/aidanns/agent-auth/releases/tag/v0.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`task release` auto-derives the next version.** Run `task release` with no
   argument and the script walks Conventional Commits since the last `v*` tag to
   pick a major / minor / patch bump (BREAKING → major, `feat:` → minor,
-  `fix:` → patch). Pass `task release -- X.Y.Z` to override.
+  `fix:` → patch). Pass `task release -- X.Y.Z` to override. While the current
+  tag is in the `0.x` range the API is not considered stable (SemVer 2.0.0 §4),
+  so a detected major bump is demoted to a minor bump; pass an explicit
+  `task release -- 1.0.0` to graduate.
 
 ## [0.1.0] - 2026-04-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,18 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
-
-- **`task release` auto-derives the next version.** Run `task release` with no
-  argument and the script walks Conventional Commits since the last `v*` tag to
-  pick a major / minor / patch bump (BREAKING → major, `feat:` → minor,
-  `fix:` → patch). Pass `task release -- X.Y.Z` to override. While the current
-  tag is in the `0.x` range the API is not considered stable (SemVer 2.0.0 §4),
-  so a detected major bump is demoted to a minor bump; pass an explicit
-  `task release -- 1.0.0` to graduate.
-
-## [0.1.0] - 2026-04-19
-
 ### Added
 
 - **agent-auth server and CLI** — HTTP validation server (`agent-auth serve`) with full
@@ -37,5 +25,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   via agent-auth. Stores credentials in the system keyring (falls back to a
   `~/.config/things-cli/credentials.yaml` file when no keyring backend is available).
 
-[0.1.0]: https://github.com/aidanns/agent-auth/releases/tag/v0.1.0
-[unreleased]: https://github.com/aidanns/agent-auth/compare/v0.1.0...HEAD
+### Changed
+
+- **`task release` auto-derives the next version.** Run `task release` with no
+  argument and the script walks Conventional Commits since the last `v*` tag to
+  pick a major / minor / patch bump (BREAKING → major, `feat:` → minor,
+  `fix:` → patch). Pass `task release -- X.Y.Z` to override. While the current
+  tag is in the `0.x` range the API is not considered stable (SemVer 2.0.0 §4),
+  so a detected major bump is demoted to a minor bump; pass an explicit
+  `task release -- 1.0.0` to graduate.
+
+[unreleased]: https://github.com/aidanns/agent-auth/compare/HEAD
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`task release` auto-derives the next version.** Run `task release` with no
+  argument and the script walks Conventional Commits since the last `v*` tag to
+  pick a major / minor / patch bump (BREAKING → major, `feat:` → minor,
+  `fix:` → patch). Pass `task release -- X.Y.Z` to override.
+
 ## [0.1.0] - 2026-04-19
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,5 +27,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   via agent-auth. Stores credentials in the system keyring (falls back to a
   `~/.config/things-cli/credentials.yaml` file when no keyring backend is available).
 
-[Unreleased]: https://github.com/aidanns/agent-auth/compare/v0.1.0...HEAD
 [0.1.0]: https://github.com/aidanns/agent-auth/releases/tag/v0.1.0
+[unreleased]: https://github.com/aidanns/agent-auth/compare/v0.1.0...HEAD
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,22 +85,35 @@ For every user-facing PR, update `CHANGELOG.md` before merging:
 
 ### Cutting a release
 
-`task release` is the release entrypoint. It delegates to `scripts/release.sh`
-which performs these steps automatically:
+`task release` is the release entrypoint. It delegates to `scripts/release.sh`,
+which:
 
-1. Validates the working tree is clean and local `main` matches `origin/main`.
-2. Checks that `CHANGELOG.md` contains a `## [X.Y.Z]` section with content.
-3. Prompts for confirmation, then creates a signed git tag (`vX.Y.Z`).
-4. Pushes the tag to `origin`.
-5. Creates a GitHub release from the CHANGELOG entry for that version.
+1. Resolves the target version — either from the argument you pass, or by
+   deriving it from Conventional Commits since the last `v*` tag (see below).
+2. Validates the working tree is clean and local `main` matches `origin/main`.
+3. Checks that `CHANGELOG.md` contains a `## [X.Y.Z]` section with content for
+   the resolved version.
+4. Prompts for confirmation, then creates a signed git tag (`vX.Y.Z`).
+5. Pushes the tag to `origin`.
+6. Creates a GitHub release from the CHANGELOG entry for that version.
+
+#### Version resolution
+
+| Invocation              | Behaviour                                                                                                                                                                                                                                                                                                                                |
+| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `task release`          | Auto-detect. Finds the latest `vX.Y.Z` tag, walks commits since that tag, and applies the largest SemVer bump implied by their Conventional Commit types: any `<type>!:` subject or `BREAKING CHANGE:` footer → **major**; any `feat:` → **minor**; any `fix:` → **patch**. Other types alone (docs, chore, refactor, ...) → no release. |
+| `task release -- 1.2.3` | Explicit override — use this for the very first release (before any `v*` tag exists) or to force a non-default bump (e.g. `1.0.0` graduation).                                                                                                                                                                                           |
 
 To cut a release:
 
-1. Move the entries under `## [Unreleased]` into a new `## [X.Y.Z] - YYYY-MM-DD`
-   section in `CHANGELOG.md`, replacing `X.Y.Z` with the new version and
-   `YYYY-MM-DD` with today's date. Leave `## [Unreleased]` empty above it.
-2. Commit and push: `git commit -m "chore: prepare release vX.Y.Z"`.
-3. Run `task release -- X.Y.Z` (e.g. `task release -- 1.2.3`).
+1. Move the entries under `## [Unreleased]` in `CHANGELOG.md` into a new
+   `## [X.Y.Z] - YYYY-MM-DD` section. If you don't know the version yet, run
+   `task release` once — it will print the auto-detected version (e.g.
+   `Auto-detected minor bump from commits since v0.1.0: v0.2.0`) then exit
+   asking you to update the CHANGELOG.
+2. Leave a fresh empty `## [Unreleased]` section above the new version.
+3. Commit and push: `git commit -m "chore: prepare release vX.Y.Z"`.
+4. Run `task release` (auto-detect) or `task release -- X.Y.Z` (explicit).
 
 The version string embedded in the distributed package is derived from the git
 tag at build time via `setuptools-scm`; no other version file needs updating.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,29 +2,36 @@
 
 ## Dev setup
 
-1. Install [uv](https://docs.astral.sh/uv/) — the project's canonical
-   Python package and environment manager (`brew install uv` on macOS,
-   `curl -LsSf https://astral.sh/uv/install.sh | sh` elsewhere).
-2. Install [go-task](https://taskfile.dev/installation/) — the project's
-   canonical task runner (`brew install go-task` on macOS,
-   `sh -c "$(curl -fsSL https://taskfile.dev/install.sh)" -- -d -b "$HOME/.local/bin"`
-   elsewhere).
-3. Install [shellcheck](https://www.shellcheck.net/) and
-   [shfmt](https://github.com/mvdan/sh) — required by `task lint` and
-   `task format` (and gated in CI). On macOS: `brew install shellcheck shfmt`. On Debian/Ubuntu: `apt-get install shellcheck` and download
-   `shfmt` from its [GitHub releases](https://github.com/mvdan/sh/releases).
-4. Install [mdformat](https://mdformat.readthedocs.io/) with its GFM and
-   tables plugins — required by `task format` for Markdown. Install as a
-   uv-managed tool: `uv tool install mdformat --with mdformat-gfm --with mdformat-tables`.
-5. Install [taplo](https://taplo.tamasfe.dev/) — required by
-   `task format` for TOML. On macOS: `brew install taplo`. Elsewhere:
-   download from [GitHub releases](https://github.com/tamasfe/taplo/releases).
-6. Install [keep-sorted](https://github.com/google/keep-sorted) — required
-   by `task lint` to verify annotated sorted blocks stay sorted. On macOS
-   / Linux: `go install github.com/google/keep-sorted@latest`, or download
-   from [GitHub releases](https://github.com/google/keep-sorted/releases).
-7. Clone the repo and `cd` into it.
-8. Run `task --list` to see every repeatable operation.
+01. Install [uv](https://docs.astral.sh/uv/) — the project's canonical
+    Python package and environment manager (`brew install uv` on macOS,
+    `curl -LsSf https://astral.sh/uv/install.sh | sh` elsewhere).
+02. Install [go-task](https://taskfile.dev/installation/) — the project's
+    canonical task runner (`brew install go-task` on macOS,
+    `sh -c "$(curl -fsSL https://taskfile.dev/install.sh)" -- -d -b "$HOME/.local/bin"`
+    elsewhere).
+03. Install [shellcheck](https://www.shellcheck.net/) and
+    [shfmt](https://github.com/mvdan/sh) — required by `task lint` and
+    `task format` (and gated in CI). On macOS: `brew install shellcheck shfmt`. On Debian/Ubuntu: `apt-get install shellcheck` and download
+    `shfmt` from its [GitHub releases](https://github.com/mvdan/sh/releases).
+04. Install [mdformat](https://mdformat.readthedocs.io/) with its GFM and
+    tables plugins — required by `task format` for Markdown. Install as a
+    uv-managed tool: `uv tool install mdformat --with mdformat-gfm --with mdformat-tables`.
+05. Install [taplo](https://taplo.tamasfe.dev/) — required by
+    `task format` for TOML. On macOS: `brew install taplo`. Elsewhere:
+    download from [GitHub releases](https://github.com/tamasfe/taplo/releases).
+06. Install [keep-sorted](https://github.com/google/keep-sorted) — required
+    by `task lint` to verify annotated sorted blocks stay sorted. On macOS
+    / Linux: `go install github.com/google/keep-sorted@latest`, or download
+    from [GitHub releases](https://github.com/google/keep-sorted/releases).
+07. Install [lefthook](https://lefthook.dev/) — runs the pre-commit
+    checks configured in `lefthook.yml` (shellcheck, shfmt, ruff,
+    mdformat, taplo, keep-sorted). On macOS: `brew install lefthook`.
+    Elsewhere: `go install github.com/evilmartians/lefthook@latest`.
+08. Clone the repo and `cd` into it.
+09. Run `task install-hooks` to install the pre-commit hook shim.
+    `task verify-standards` gates on this being installed so that
+    configured pre-commit checks actually fire on every commit.
+10. Run `task --list` to see every repeatable operation.
 
 The first time you run any venv-backed task (e.g. `task test`,
 `task build`, or a service task like `task agent-auth -- serve`), the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,10 +99,16 @@ which:
 
 #### Version resolution
 
-| Invocation              | Behaviour                                                                                                                                                                                                                                                                                                                                |
-| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `task release`          | Auto-detect. Finds the latest `vX.Y.Z` tag, walks commits since that tag, and applies the largest SemVer bump implied by their Conventional Commit types: any `<type>!:` subject or `BREAKING CHANGE:` footer → **major**; any `feat:` → **minor**; any `fix:` → **patch**. Other types alone (docs, chore, refactor, ...) → no release. |
-| `task release -- 1.2.3` | Explicit override — use this for the very first release (before any `v*` tag exists) or to force a non-default bump (e.g. `1.0.0` graduation).                                                                                                                                                                                           |
+| Invocation              | Behaviour                                                                                                                                                                                                                                                                                                                                                     |
+| ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `task release`          | Auto-detect. Finds the latest `vX.Y.Z` tag, walks commits since that tag, and applies the largest SemVer bump implied by their Conventional Commit types: any `<type>!:` subject or `BREAKING CHANGE:` / `BREAKING-CHANGE:` footer → **major**; any `feat:` → **minor**; any `fix:` → **patch**. Other types alone (docs, chore, refactor, ...) → no release. |
+| `task release -- 1.2.3` | Explicit override — use this for the very first release (before any `v*` tag exists) or to force a non-default bump (e.g. `1.0.0` graduation).                                                                                                                                                                                                                |
+
+While the current tag is in the `0.x` range, the public API is not
+considered stable (SemVer 2.0.0 §4). A BREAKING change that would
+normally map to a **major** bump is demoted to a **minor** bump until
+`v1.0.0` ships. Force the graduation to `1.0.0` with an explicit
+`task release -- 1.0.0` when the API is ready to stabilise.
 
 To cut a release:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,11 +75,35 @@ signing are sibling requirements, not the same thing.
 
 ## Release process
 
-`task release` is the release entrypoint. The underlying automation
-(version bump, tag, GitHub release, publish) is tracked in
-[#18](https://github.com/aidanns/agent-auth/issues/18) and is not yet
-implemented — running the task today exits non-zero to prevent manual
-releases that would skip the standard checks.
+### Before releasing
+
+For every user-facing PR, update `CHANGELOG.md` before merging:
+
+1. Add a bullet under `## [Unreleased]` describing the user-visible change.
+2. Keep the format consistent with the existing entries (present-tense action,
+   linked to relevant issues/PRs where helpful).
+
+### Cutting a release
+
+`task release` is the release entrypoint. It delegates to `scripts/release.sh`
+which performs these steps automatically:
+
+1. Validates the working tree is clean and local `main` matches `origin/main`.
+2. Checks that `CHANGELOG.md` contains a `## [X.Y.Z]` section with content.
+3. Prompts for confirmation, then creates a signed git tag (`vX.Y.Z`).
+4. Pushes the tag to `origin`.
+5. Creates a GitHub release from the CHANGELOG entry for that version.
+
+To cut a release:
+
+1. Move the entries under `## [Unreleased]` into a new `## [X.Y.Z] - YYYY-MM-DD`
+   section in `CHANGELOG.md`, replacing `X.Y.Z` with the new version and
+   `YYYY-MM-DD` with today's date. Leave `## [Unreleased]` empty above it.
+2. Commit and push: `git commit -m "chore: prepare release vX.Y.Z"`.
+3. Run `task release -- X.Y.Z` (e.g. `task release -- 1.2.3`).
+
+The version string embedded in the distributed package is derived from the git
+tag at build time via `setuptools-scm`; no other version file needs updating.
 
 ## Commit signing
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -159,8 +159,9 @@ specific component of the current implementation:
 
 - **AC — Access Control**: the three-tier scope model (`allow`/`prompt`/`deny`)
   and the per-family scope set enforced in `src/agent_auth/scopes.py` and the
-  `validate_token_scope` function in
-  [`design/functional_decomposition.yaml`](design/functional_decomposition.yaml).
+  [Check Scope Authorization](design/functional_decomposition.yaml#L28) and
+  [Resolve Access Tier](design/functional_decomposition.yaml#L30) leaf
+  functions.
 - **AU — Audit and Accountability**: the append-only audit log in
   `src/agent_auth/audit.py`, fed by every token lifecycle event and
   authorization decision.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,179 @@
+# Security
+
+## Trust boundaries
+
+agent-auth is a **local, single-user** authorization system. All components
+bind to `127.0.0.1` by default; they are not designed for multi-tenant or
+network-exposed deployment.
+
+```
+┌──────────────────────┐            ┌─────────────────────────────────────────────┐
+│  Devcontainer        │            │  Host machine                               │
+│                      │            │                                             │
+│  things-cli ──────HTTP────────────▶  things-bridge ──────────────▶ Things 3     │
+│              │       │            │    │                                        │
+│              │       │            │    │ HTTP (validate, approve)               │
+│              │       │            │    ▼                                        │
+│              └────HTTP────────────▶  agent-auth                                  │
+│                      │            │    ├─ tokens.db (SQLite + AES-256-GCM)      │
+│                      │            │    └─ signing key (system keyring)          │
+└──────────────────────┘            └─────────────────────────────────────────────┘
+```
+
+Trust boundary decisions:
+
+- **agent-auth** is the trust root. Only it holds the signing key and token
+  store. All other components validate tokens by calling agent-auth's HTTP API;
+  they never access the store or key directly.
+- **things-bridge** trusts agent-auth's validation response and the configured
+  Things-client CLI's stdout. It does not trust the bearer token it receives
+  from things-cli — it always re-validates with agent-auth before acting.
+- **things-cli** trusts agent-auth for token issuance and refresh, and
+  things-bridge for data responses.
+- **things-client-cli-applescript** runs on the host with the user's macOS
+  Automation permission. It receives argv from things-bridge and emits JSON on
+  stdout; its only trust assumption is that the invoking process (things-bridge)
+  is legitimate. No authentication between bridge and client CLI.
+- **Notification plugin** (JIT approval): currently loaded in-process in the
+  agent-auth server via `importlib.import_module`, which means a malicious
+  plugin would run inside the process that holds the signing and encryption keys.
+  Migration to an out-of-process plugin boundary is tracked in
+  [#6](https://github.com/aidanns/agent-auth/issues/6).
+
+## Threat model
+
+The following STRIDE analysis covers the agent-auth server, things-bridge, and
+things-cli. Threats are rated High / Medium / Low by impact × likelihood.
+
+### Spoofing
+
+| Threat | Rating | Mitigation |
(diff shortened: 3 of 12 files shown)


install.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Install agent-auth, things-bridge, things-cli, and things-client-cli-applescript
+# from the agent-auth GitHub repository into a uv-managed tool environment.
+
+set -euo pipefail
+
+REPO="aidanns/agent-auth"
+INSTALL_SOURCE="git+https://github.com/${REPO}.git"
```
aidanns: Claude Reviewer: `INSTALL_SOURCE` is `git+https://github.com/.git` with no tag, so `uv tool install --force` always installs whatever is on the HEAD of `main` — not the version that was just tagged. After `task release -- 0.1.0` creates `v0.1.0`, a user running the one-liner will still get the current HEAD rather than the released version.

Consider appending the latest tag at release time, e.g. by having `release.sh` update `install.sh` to set `INSTALL_SOURCE="git+https://github.com/${REPO}.git@v${VERSION}"`, or by making the installer accept an optional version argument:

```bash
VERSION="${1:-$(gh api repos/aidanns/agent-auth/releases/latest --jq .tag_name)}"
INSTALL_SOURCE="git+https://github.com/${REPO}.git@${VERSION}"
```

Without this, the installer and the release process are effectively decoupled.

scripts/release.sh
```diff
@@ -1,19 +1,122 @@
...
-exit 1
+VERSION="${1}"
+TAG="v${VERSION}"
+
+if [[ ! "${VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+  echo "release: version must be in MAJOR.MINOR.PATCH form (got '${VERSION}')" >&2
+  exit 1
+fi
+
+for cmd in git gh uv; do
```
aidanns: Claude Reviewer: `uv` is listed as a required dependency here but is never invoked in the script. The script only uses `git` and `gh`. This will cause `task release` to fail on machines where `uv` is not on PATH but `git` and `gh` are. Either remove `uv` from this list, or add the `uv` invocation you intended (e.g. `uv build` + `uv publish` to produce and upload a wheel).

scripts/verify-standards.sh
```diff
@@ -384,3 +384,167 @@ if [[ ${contributing_missing} -ne 0 ]]; then
...
+  echo "  $2" >&2
+  security_missing=1
+}
+
+if [[ ! -f SECURITY.md ]]; then
+  fail_security_check \
+    "SECURITY.md is missing from the repo root." \
+    "Add SECURITY.md covering trust boundaries, threat model, key handling, revocation flow, audit surface, and vulnerability reporting."
+else
+  security_section_names=(
```
aidanns: Claude Reviewer: `security_section_names` carries a `# keep-sorted start/end` marker but `security_section_patterns` does not. If a future `keep-sorted` run re-orders the names array, the patterns array will stay in the old order and silently misalign — the wrong pattern is checked against the wrong section name, potentially letting a malformed SECURITY.md pass the check.

Either remove the `keep-sorted` markers from `security_section_names`, or add matching markers to `security_section_patterns` so they are sorted in lockstep. The simplest fix is to drop the `keep-sorted` markers entirely from this block, since the names are checked by index not by value.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+AES-256-GCM encryption key), a local SQLite token store, and a JIT approval
+surface. NIST SP 800-53 Rev 5 provides clear control families for these
+responsibilities (AC, AU, IA, SC, SI) with well-defined implementation guidance,
+and its control catalog is machine-readable for automated compliance checking.
+
+### Control families relevant to this project
+
+| Family | ID | Selected controls |
+|--------|----|-------------------|
+| Access Control | AC | AC-2 (Account Management), AC-3 (Access Enforcement), AC-6 (Least Privilege), AC-17 (Remote Access) |
```
aidanns: Claude Reviewer: SC-8 (Transmission Confidentiality and Integrity) is listed as a selected NIST control, but in the devcontainer deployment scenario shown in the trust boundary diagram, traffic between `things-cli` and the host services (`things-bridge`, `agent-auth`) travels over a real network interface — not loopback — and is currently plain HTTP with no TLS. SC-8 is therefore selected but not implemented for that traffic path.

This should be noted here, either by explicitly marking SC-8 as "partially implemented — loopback only; devcontainer traffic is unencrypted" or by adding a gap entry to the threat model under Information Disclosure. Silently listing a control as "selected" when it has a known implementation gap risks misleading future audits.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+| Oversized request body exhausting agent-auth | Medium | Request bodies are capped at 1 MiB. |
+| Rapid token-creation filling `tokens.db` | Low | No rate limiting currently; the server is local-only so the attack requires code execution on the host. Rate limiting is a future hardening step. |
+
+### Elevation of privilege
+
+| Threat | Rating | Mitigation |
+|--------|--------|------------|
+| AI agent escalates from `allow`-tier to `prompt`-tier scope without JIT approval | High | Scope tier is validated server-side on every request; the agent cannot self-approve `prompt`-tier requests. |
+| Malicious notification plugin runs in-process | High | Tracked in [#6](https://github.com/aidanns/agent-auth/issues/6). Current mitigation: only install plugins from trusted sources under your user account. |
+| things-cli constructs arbitrary argv passed to things-client CLI | Medium | things-bridge constructs the argv from validated, schema-matched request parameters, not from raw client input. |
```
aidanns: Claude Reviewer: The threat description says "things-cli constructs arbitrary argv passed to things-client CLI", but `things-cli` does not construct any argv — it only sends validated HTTP parameters to `things-bridge`. It is `things-bridge` that constructs the argv passed to the `things-client-cli-applescript` subprocess. The mitigation text is correct ("things-bridge constructs the argv from validated, schema-matched request parameters"), but the threat label misidentifies the actor.

Suggest renaming to something like: "Untrusted parameters from things-cli flow through things-bridge into things-client CLI argv".

install.sh
aidanns: Claude: Done! install.sh now queries the GitHub releases API to resolve the latest tag and installs `git+<url>@<tag>`. Falls back to HEAD if no release has been cut yet.

scripts/release.sh
aidanns: Claude: Done! Removed `uv` from the dependency check loop — the release script never invokes it.

scripts/verify-standards.sh
aidanns: Claude: Done! Removed the `# keep-sorted` markers from `security_section_names`. The array must stay order-locked to the parallel `security_section_patterns` array and cannot be auto-sorted.

SECURITY.md
aidanns: Claude: Done! Added a ¹ footnote to the SC-8 row in the NIST control table documenting the known gap: plain-HTTP over a real network interface in the devcontainer scenario, with a note that TLS is a future hardening step.

SECURITY.md
aidanns: Claude: Done! Fixed the threat label: `things-bridge constructs arbitrary argv passed to things-client CLI`.

Taskfile.yml
```diff
@@ -62,9 +62,9 @@ tasks:
       - scripts/install-hooks.sh
 
   release:
-    desc: Cut a release (version bump, tag, GitHub release, publish).
+    desc: "Cut a release (version bump, tag, GitHub release, publish). Usage: task release -- X.Y.Z"
```
aidanns: what determines the version to be used in a release?

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+- **Notification plugin** (JIT approval): currently loaded in-process in the
+  agent-auth server via `importlib.import_module`, which means a malicious
+  plugin would run inside the process that holds the signing and encryption keys.
+  Migration to an out-of-process plugin boundary is tracked in
+  [#6](https://github.com/aidanns/agent-auth/issues/6).
+
+## Threat model
+
+The following STRIDE analysis covers the agent-auth server, things-bridge, and
+things-cli. Threats are rated High / Medium / Low by impact × likelihood.
```
aidanns: explicitly state which risk assessment framework is being used here.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Spoofing
+
+| Threat | Rating | Mitigation |
```
aidanns: explicitly state the impact and liklihood for each threat, and whether the mitigation is mitigating the liklihood, or the impact.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+|--------|--------|------------|
+| Direct modification of `tokens.db` | High | Scope and HMAC-signature fields are AES-256-GCM encrypted at rest; modification without the key produces authentication-tag failures on read. |
+| Replay of a revoked token | High | Revocation writes `revoked_at` to the token record; validation checks `revoked_at IS NULL` before accepting. Reuse of a refresh token triggers family-wide revocation. |
+| Tampering with the signing key in the system keyring | High | Requires OS-level access to the keyring (macOS Keychain or libsecret). If the key is replaced, all previously issued tokens become invalid on next validation. |
+
+### Repudiation
+
+| Threat | Rating | Mitigation |
+|--------|--------|------------|
+| Agent denies performing a privileged operation | Medium | All token operations and authorization decisions are written to the audit log before the response is sent. |
```
aidanns: which audit log? specify a service here.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+  Automation permission. It receives argv from things-bridge and emits JSON on
+  stdout; its only trust assumption is that the invoking process (things-bridge)
+  is legitimate. No authentication between bridge and client CLI.
+- **Notification plugin** (JIT approval): currently loaded in-process in the
+  agent-auth server via `importlib.import_module`, which means a malicious
+  plugin would run inside the process that holds the signing and encryption keys.
+  Migration to an out-of-process plugin boundary is tracked in
+  [#6](https://github.com/aidanns/agent-auth/issues/6).
+
+## Threat model
```
aidanns, line 43: for every mitigation listed here, provide a link to the function in the functional decomposition that relates to the mitigation (to show that the system functions if implemented are secure), or to the github issue that tracks that mitigation for future implementation.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+| Threat | Rating | Mitigation |
+|--------|--------|------------|
+| Oversized request body exhausting agent-auth | Medium | Request bodies are capped at 1 MiB. |
+| Rapid token-creation filling `tokens.db` | Low | No rate limiting currently; the server is local-only so the attack requires code execution on the host. Rate limiting is a future hardening step. |
+
+### Elevation of privilege
+
+| Threat | Rating | Mitigation |
+|--------|--------|------------|
+| AI agent escalates from `allow`-tier to `prompt`-tier scope without JIT approval | High | Scope tier is validated server-side on every request; the agent cannot self-approve `prompt`-tier requests. |
```
aidanns: what does it mean to escalate from allow to prompt tiers?

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+3. **Token expiry**: access tokens carry an expiry timestamp; validation rejects
+   expired tokens with `401 Unauthorized`. things-cli retries with the refresh
+   token on `token_expired` responses.
+4. **Key rotation**: replacing the signing key in the keyring invalidates all
+   previously issued tokens on the next validation call, acting as a
+   revocation-of-last-resort.
+
+## Audit surface
+
+The following events are written to the audit log (stderr + structured log):
```
aidanns: add a github issue for consolidating on a single structure log audit format for all services.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+
+This is a personal project with no published releases or user base beyond the
+author. If you find a security issue:
+
+1. **Do not open a public GitHub issue.** Use
+   [GitHub private vulnerability reporting](https://github.com/aidanns/agent-auth/security/advisories/new)
+   to disclose the issue confidentially.
+2. Alternatively, email **aidanns@gmail.com** with the subject line
+   `[agent-auth] Security vulnerability`.
+3. Expect acknowledgment within 7 days and a fix or mitigation within 30 days.
```
aidanns: remove these.

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+|--------|----|-------------------|
+| Access Control | AC | AC-2 (Account Management), AC-3 (Access Enforcement), AC-6 (Least Privilege), AC-17 (Remote Access) |
+| Audit and Accountability | AU | AU-2 (Event Logging), AU-3 (Content of Audit Records), AU-9 (Protection of Audit Information), AU-12 (Audit Record Generation) |
+| Identification and Authentication | IA | IA-5 (Authenticator Management), IA-9 (Service Identification and Authentication) |
+| System and Communications Protection | SC | SC-8 (Transmission Confidentiality and Integrity), SC-28 (Protection of Information at Rest) |
+| System and Information Integrity | SI | SI-10 (Information Input Validation), SI-12 (Information Management and Retention) |
+
+Implementation plans for new features should verify compliance against the
+controls above that are in scope for the change. Control applicability
+assessments live in `design/decisions/` ADRs.
```
aidanns: is this actually true?

SECURITY.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Control families relevant to this project
+
+| Family | ID | Selected controls |
+|--------|----|-------------------|
+| Access Control | AC | AC-2 (Account Management), AC-3 (Access Enforcement), AC-6 (Least Privilege), AC-17 (Remote Access) |
+| Audit and Accountability | AU | AU-2 (Event Logging), AU-3 (Content of Audit Records), AU-9 (Protection of Audit Information), AU-12 (Audit Record Generation) |
+| Identification and Authentication | IA | IA-5 (Authenticator Management), IA-9 (Service Identification and Authentication) |
+| System and Communications Protection | SC | SC-8 (Transmission Confidentiality and Integrity), SC-28 (Protection of Information at Rest) |
+| System and Information Integrity | SI | SI-10 (Information Input Validation), SI-12 (Information Management and Retention) |
```
aidanns: these statements feel high level and/or throw-away - do we have enough detail here to ensure that we are appropriately representing our implementation maturity against these control families?

scripts/verify-standards.sh
```diff
@@ -449,15 +449,13 @@ if [[ ! -f SECURITY.md ]]; then
     "Add SECURITY.md covering trust boundaries, threat model, key handling, revocation flow, audit surface, and vulnerability reporting."
 else
   security_section_names=(
-    # keep-sorted start
```
aidanns: why were these removed? seems logical to have them here

install.sh
```diff
@@ -21,7 +21,21 @@ EOF
   exit 1
 fi
 
-echo "Installing agent-auth from ${INSTALL_SOURCE} ..."
+# Resolve the latest release tag so users get the tagged version, not HEAD.
```
aidanns: check the structure of https://github.com/aidanns/systems-engineering/blob/main/install.sh and replicate that in this file

SECURITY.md
```diff
@@ -152,13 +152,20 @@ and its control catalog is machine-readable for automated compliance checking.
...
+| Audit and Accountability             | AU  | AU-2 (Event Logging), AU-3 (Content of Audit Records), AU-9 (Protection of Audit Information), AU-12 (Audit Record Generation) |
+| Identification and Authentication    | IA  | IA-5 (Authenticator Management), IA-9 (Service Identification and Authentication)                                              |
+| System and Communications Protection | SC  | SC-8 (Transmission Confidentiality and Integrity) ¹, SC-28 (Protection of Information at Rest)                                 |
+| System and Information Integrity     | SI  | SI-10 (Information Input Validation), SI-12 (Information Management and Retention)                                             |
+
+¹ **SC-8 known gap**: In the devcontainer deployment scenario (see trust boundary
+diagram), `things-cli` communicates with `things-bridge` and `agent-auth` over a
+plain HTTP loopback connection that crosses a real network interface. SC-8 is
+not currently satisfied for this path. Mitigation: TLS between devcontainer and
```
aidanns: ensure we have a github issue for this - always reference the github issues if we claim that there is future worked tracked.

Taskfile.yml
aidanns: Claude: Done! The version is determined entirely by the argument you pass: `task release -- X.Y.Z`. The script validates it's in MAJOR.MINOR.PATCH form, checks that a matching `## [X.Y.Z]` section exists in CHANGELOG.md, then creates and pushes the tag `vX.Y.Z`. Updated the task desc to make this explicit.

SECURITY.md
aidanns: Claude: Done! The threat model intro now explicitly names **NIST SP 800-30 Rev 1** as the risk assessment framework, with a description of how Impact and Likelihood are rated independently and combined into an overall Rating.

SECURITY.md
aidanns: Claude: Done! All STRIDE tables now have separate Impact and Likelihood columns. Each mitigation row notes what it targets (e.g. 'Targets: **likelihood**' or 'Targets: **impact**') so it's clear which dimension the mitigation addresses.

SECURITY.md
aidanns: Claude: Done! Changed to 'agent-auth's audit log' throughout.

SECURITY.md
aidanns: Claude: Done! Every mitigation now links to its implementing function in `design/functional_decomposition.yaml` (e.g. `[Verify Token Signature](design/functional_decomposition.yaml#L24)`) or to the GitHub issue tracking the future implementation.

SECURITY.md
aidanns: Claude: Done! Added a plain-English explanation before the EoP table: `allow` tier = immediately permitted if the token holds the scope; `prompt` tier = held pending real-time human approval even if the token holds the scope. The threat row was also rephrased to: 'AI agent invokes a `prompt`-tier scope without triggering JIT approval'.

SECURITY.md
aidanns: Claude: Done! Created [#100](https://github.com/aidanns/agent-auth/issues/100) — 'Consolidate structured audit log format across all services'. Referenced from the audit surface section.

SECURITY.md
aidanns: Claude: Done! Removed the acknowledgment and fix-window SLA lines from the vulnerability reporting section.

SECURITY.md
aidanns: Claude: Good catch — not yet true. Changed to: 'Control applicability assessments for new features *should* be documented in `design/decisions/` ADRs at the time the feature is implemented.' (future intent, not current state).

SECURITY.md
aidanns: Claude: Done! Expanded the NIST control families table with a Status column showing Implemented / Partial / Planned for each family, with specific gap references (e.g. 'AU-9 lacks cryptographic integrity ([#103](https://github.com/aidanns/agent-auth/issues/103))') and concrete implementation details per control.

scripts/verify-standards.sh
aidanns: Claude: Restructured to a single `security_sections` array of `"name|pattern"` combined entries with `# keep-sorted` markers. This keeps entries sortable without the parallel-array misalignment risk (a keep-sorted run can't accidentally sort names independently of patterns). The entries are still alphabetically ordered and the check logic splits on `|` to extract name and pattern.

install.sh
aidanns: Claude: Done! Restructured to follow the systems-engineering install.sh pattern: added `--uninstall` flag with matching instructions, explicit python3 prerequisite check, installation verification step (checks `agent-auth` is on PATH after install), PATH warning (checks uv tool bin dir against $PATH), and uninstall reminder at the end.

SECURITY.md
aidanns: Claude: Done! Created [#101](https://github.com/aidanns/agent-auth/issues/101) — 'Add TLS for devcontainer-to-host HTTP traffic (SC-8 compliance)'. The SC-8 footnote now references it explicitly.

CHANGELOG.md
```diff
@@ -0,0 +1,31 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.1.0] - 2026-04-19
```
aidanns: we havn't done a release yet - everything should be listed as unreleased.

scripts/release.sh
```diff
@@ -1,19 +1,236 @@
...
+# other types (docs, chore, refactor, style, test, ci, build, perf) do not
+# produce a release on their own, matching the Conventional Commits spec.
+compute_bump() {
+  local range="$1"
+  local bump="none"
+  local sha subject body
+  # Regexes live in variables because shellcheck can't parse `:` inside `[[ =~ ]]`
+  # patterns (SC1073). Quoting the RHS of `=~` would turn it into a literal match,
+  # so the variable-indirection trick is the shellcheck-safe way to keep them regex.
+  local breaking_re='^[a-zA-Z]+(\([^)]+\))?!:'
```
aidanns: Claude Reviewer: `breaking_re` uses `[a-zA-Z]+` (case-insensitive, any type name), while `feat_re` and `fix_re` are lowercase-only and anchored to `feat`/`fix`. This means `Feat!: x`, `FIX!: x`, or even `FOO!: x` (an unknown type) all trigger a **major** bump, but `Feat: x` doesn't register at all. Suggest restricting the breaking regex to the same Conventional-Commits type whitelist you accept elsewhere, e.g. `^(feat|fix|docs|style|refactor|perf|test|build|ci|chore|revert)(\([^)]+\))?!:` — or at minimum `^[a-z]+` for consistency with feat/fix. Note `revert!:` is arguable but spec-aligned; a typo-type like `Feats!:` should probably not silently become a major release.

scripts/release.sh
```diff
@@ -1,19 +1,236 @@
...
+    if [[ "${subject}" =~ ${feat_re} ]]; then
+      [[ "${bump}" != "major" ]] && bump="minor"
+      continue
+    fi
+
+    if [[ "${subject}" =~ ${fix_re} ]]; then
+      [[ "${bump}" == "none" ]] && bump="patch"
+      continue
+    fi
+  done < <(git log --format='%H' "${range}")
```
aidanns: Claude Reviewer: Failures of `git log --format='%H' "${range}"` inside the process substitution are silently swallowed — `pipefail` does not propagate through `<(...)`, and `set -e` doesn't fire because the failing command is in a child shell that isn't part of the foreground pipeline. If the range is malformed or the repo is in a weird state, `compute_bump` prints `none` and the caller reports 'no feat/fix/BREAKING commits since LAST_TAG — nothing to release' instead of the real error.

Reproduction:

```bash
compute_bump() {
  local bump=none sha
  while IFS= read -r sha; do bump=patch; done < <(false)
  echo "$bump"
}
compute_bump   # prints "none", exit 0
```

Suggest capturing the git output to a temp variable first and checking the exit code explicitly, e.g. `commits="$(git log --format='%H' "${range}")"` above the loop, then `while … done <<<"${commits}"`.

scripts/release.sh
```diff
@@ -1,19 +1,236 @@
...
+# Apply a SemVer bump ("major"/"minor"/"patch") to vX.Y.Z and print X.Y.Z.
+apply_bump() {
+  local last_tag="$1"
+  local bump="$2"
+  local last_version="${last_tag#v}"
+  local major minor patch
+  IFS='.' read -r major minor patch <<<"${last_version}"
+  case "${bump}" in
+    major)
+      major=$((major + 1))
```
aidanns: Claude Reviewer: While in 0.x the project almost certainly does **not** want a breaking-change commit to graduate to 1.0.0 automatically. SemVer (clause 4) explicitly says 0.y.z is for initial development and 'anything may change at any time'; the community convention implemented by `standard-version`, `release-please`, etc. is to keep BREAKING in 0.x as a **minor** bump (0.1.0 → 0.2.0) and require an explicit `task release -- 1.0.0` to graduate.

With the current logic, a single `feat!:` on the current 0.1.0 codebase silently becomes v1.0.0, which is likely not what the CONTRIBUTING.md reader expects ('use this for the very first release … or to force a … 1.0.0 graduation' implies graduation is manual). Suggest either:

1. Cap pre-1.0 bumps at minor in `apply_bump` (`if [[ $major -eq 0 && $bump == major ]]; then bump=minor; fi`), or
2. Refuse to auto-graduate across the 0.x → 1.x boundary and require `task release -- 1.0.0` explicitly.

Either way, document the chosen behaviour in CONTRIBUTING.md's 'Version resolution' table.

scripts/release.sh
```diff
@@ -1,19 +1,236 @@
...
+  local fix_re='^fix(\([^)]+\))?:'
+
+  while IFS= read -r sha; do
+    [[ -z "${sha}" ]] && continue
+    subject="$(git log -1 --format='%s' "${sha}")"
+    body="$(git log -1 --format='%b' "${sha}")"
+
+    # `<type>!:` subject OR `BREAKING CHANGE:` / `BREAKING-CHANGE:` footer → major.
+    if [[ "${subject}" =~ ${breaking_re} ]] \
+      || grep -qE '^BREAKING[ -]CHANGE:' <<<"${body}"; then
```
aidanns: Claude Reviewer: The walker counts every reachable commit in `LAST_TAG..HEAD`, including merge commits and `revert:` commits. Two cases to think about:

- **Merge commits from PRs.** `git log` (without `--first-parent`) lists every commit on merged branches. If a contributor squashes into `chore: internal cleanup` but one intermediate commit on a feature branch was `feat: …`, the merged history will bump minor based on the unreleased intermediate commit. If you squash-merge by policy you may not care, but if you ever land a merge commit you will. Consider `git log --first-parent --format='%H' "${range}"` (matches how release-please and go-semantic-release walk history).
- **Revert commits.** `revert: feat: X` body often contains `This reverts commit …` followed by the original message; `%b` will then include the original `BREAKING CHANGE:` footer if there was one, triggering a major on a pure revert. Worth at least mentioning in the docstring so the user knows to squash-revert or pass the explicit version in that case.

scripts/release.sh
```diff
@@ -1,19 +1,236 @@
...
+    exit 1
+  fi
+done
+
+# Walk commits in <range> and emit the largest SemVer bump implied by the
+# Conventional Commit subjects/bodies: "major", "minor", "patch", or "none".
+# Only the four Conventional Commit types that map to SemVer levels count;
+# other types (docs, chore, refactor, style, test, ci, build, perf) do not
+# produce a release on their own, matching the Conventional Commits spec.
+compute_bump() {
```
aidanns: Claude Reviewer: `compute_bump` and `apply_bump` are pure functions whose behaviour the rest of this PR's docs commit you to (the CONTRIBUTING.md table is now part of the project contract). Nothing else in the repo has shell-level unit tests — that's consistent — but this is the first non-trivial bit of logic in a `.sh` file that has real edge cases (regex asymmetry, 0.x, merges, reverts, empty range). Recommend either:

1. Extracting `compute_bump`/`apply_bump` into a tiny sourced library (e.g. `scripts/lib/semver.sh`) and adding `tests/test_release_semver.py` that invokes them via `bash -c 'source … && compute_bump …'` against fixture ranges (created with `git -c …` in a tmp repo), or
2. At minimum, add a fast-path smoke test to CI that runs `scripts/release.sh` in `--dry-run` mode (not currently supported) or asserts `compute_bump` outputs for a handful of canonical subjects.

Given how reliant the release process is on this logic correctness, and that a wrong bump produces a tag that can't be un-cut, this seems worth the upfront cost. `release-and-hygiene.md` doesn't explicitly require it, but `testing-standards.md` does emphasise coverage of high-consequence logic.

CONTRIBUTING.md
```diff
@@ -75,11 +75,48 @@ signing are sibling requirements, not the same thing.
...
+6. Creates a GitHub release from the CHANGELOG entry for that version.
+
+#### Version resolution
+
+| Invocation              | Behaviour                                                                                                                                                                                                                                                                                                                                |
+| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `task release`          | Auto-detect. Finds the latest `vX.Y.Z` tag, walks commits since that tag, and applies the largest SemVer bump implied by their Conventional Commit types: any `<type>!:` subject or `BREAKING CHANGE:` footer → **major**; any `feat:` → **minor**; any `fix:` → **patch**. Other types alone (docs, chore, refactor, ...) → no release. |
```
aidanns: Claude Reviewer: The table entry for `task release` says 'any `<type>!:` subject or `BREAKING CHANGE:` footer → **major**'. That's correct for the happy path, but the script also accepts `BREAKING-CHANGE:` (hyphen form per spec) and accepts any type in `<type>!:` not just feat/fix. Consider either:

- Match the docs to the current permissive behaviour: note that both `BREAKING CHANGE:` and `BREAKING-CHANGE:` footers work, and that any `<type>!:` (including `chore!:`) triggers major.
- Or tighten the script (see other review comment on the `breaking_re` asymmetry) and keep the docs as they are.

The doc table is rendered as a 2-column markdown table with very wide second column — it renders OK on GitHub, but on narrower viewports (and in mdformat output) it may wrap oddly. Could be restructured as a bulleted list under the existing steps without loss of information, but this is a nit.

scripts/release.sh
```diff
@@ -1,19 +1,236 @@
...
+  # Regexes live in variables because shellcheck can't parse `:` inside `[[ =~ ]]`
+  # patterns (SC1073). Quoting the RHS of `=~` would turn it into a literal match,
+  # so the variable-indirection trick is the shellcheck-safe way to keep them regex.
+  local breaking_re='^[a-zA-Z]+(\([^)]+\))?!:'
+  local feat_re='^feat(\([^)]+\))?:'
+  local fix_re='^fix(\([^)]+\))?:'
+
+  while IFS= read -r sha; do
+    [[ -z "${sha}" ]] && continue
+    subject="$(git log -1 --format='%s' "${sha}")"
```
aidanns: Claude Reviewer: Minor perf nit — each iteration forks two more `git log -1` processes (one for `%s`, one for `%b`), which for a large range between tags means `N*3` git invocations. You can pull subject+body for the whole range in a single call and parse records with a sentinel, e.g.:

```bash
while IFS= read -r -d '' record; do
  subject="${record%%$'\n'*}"
  body="${record#*$'\n'}"
  …
done < <(git log --format='%s%n%b%x00' "${range}")
```

Not critical for personal release cadence, but worth noting since it compounds with the swallowed-stderr problem in the other comment.

aidanns:
Claude Reviewer: Independent review of commit b30a0dc (auto-derive next release version).

**Scope of this review:** only commit b30a0dc (earlier commits on the PR were already addressed).

**Findings, roughly ranked by severity:**

1. **`breaking_re` is broader than `feat_re`/`fix_re`.** `^[a-zA-Z]+(\(…\))?!:` matches `Feat!:`, `FOO!:`, etc., while feat/fix are lowercase-only. Any typo-type with a `!` silently becomes a major release. (line-level comment on `scripts/release.sh`:56)
2. **0.x BREAKING auto-bumps to 1.0.0.** With `v0.1.0` current, a single `feat!:` auto-resolves to v1.0.0. Community convention is to require explicit graduation. (`scripts/release.sh`:95)
3. **`git log` failure inside `<(…)` is swallowed.** process-sub errors do not propagate through `pipefail`; a bad range surfaces as "nothing to release" instead of the real git error. (`scripts/release.sh`:81)
4. **Merge commits + revert commits.** `git log` without `--first-parent` sees intermediate commits from feature branches; `revert:` commits can carry the original `BREAKING CHANGE:` footer in `%b` and trigger a spurious major. (`scripts/release.sh`:67)
5. **No tests for the new pure functions.** Consistent with the rest of the repo (no shell tests), but this is the first script where nontrivial branching matters for a user-facing contract. A small Python test that shells out to `compute_bump`/`apply_bump` would cover the edge cases above. (`scripts/release.sh`:49)
6. **Docs drift nit.** The CONTRIBUTING.md version-resolution table lists only `BREAKING CHANGE:` (not `BREAKING-CHANGE:`) and implies only feat/fix with `!:` trigger major, but the implementation accepts any type. (`CONTRIBUTING.md`:104)
7. **Perf nit.** `N*3` git invocations per range; can be collapsed to one `git log` with a null-delimited record format. (`scripts/release.sh`:62)

**Verified clean:**
- `shellcheck` passes (v0.9.0).
- `shfmt -d` produces no diff.
- `set -e` / `continue` patterns inside `compute_bump` are safe — the conditional `&&` isn't the function's terminal statement.
- `git describe --match 'v[0-9]*.[0-9]*.[0-9]*'` correctly handles lightweight tags; the subsequent regex re-validation of `LAST_TAG` is belt-and-braces.
- CHANGELOG.md entry under `[Unreleased]` is consistent with the script's user-facing behaviour.
- Clean-tree check runs *after* `compute_bump`, so the docs workflow ("run `task release` once to see the suggested version") works.

Overall direction is good; #1-#3 are worth addressing before merging.

SECURITY.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Cybersecurity standard
+
+This project adopts **NIST SP 800-53 Revision 5** as its reference cybersecurity
+standard. The five control families below are in scope because each maps to a
+specific component of the current implementation:
+
+- **AC — Access Control**: the three-tier scope model (`allow`/`prompt`/`deny`)
+  and the per-family scope set enforced in `src/agent_auth/scopes.py` and the
+  `validate_token_scope` function in
```
aidanns: Claude Reviewer: The `validate_token_scope` function referenced here does not exist in `design/functional_decomposition.yaml`. The functional decomposition names it **Check Scope Authorization** (under `Token Validation`). This is the only occurrence of the string `validate_token_scope` in the repo — grep confirms it appears nowhere else. Please either rename to match the decomposition entry ("Check Scope Authorization") or drop the function reference, since the `src/agent_auth/scopes.py` pointer already covers the AC mapping.

aidanns: Claude: Done! Replaced the stale `validate_token_scope` reference with links to the actual functional-decomposition entries — [Check Scope Authorization](design/functional_decomposition.yaml#L28) and [Resolve Access Tier](design/functional_decomposition.yaml#L30) — which is what the prose was pointing at.

CHANGELOG.md
aidanns: Claude: Done! Moved the 0.1.0 entries back under `## [Unreleased]`. The tag doesn't exist yet so there's no `0.1.0` in the world for the section header to anchor to — the first `task release` will prompt us to rename Unreleased to `[X.Y.Z] - YYYY-MM-DD` just before tagging.

scripts/release.sh
aidanns: Claude: Done! In 48000de — `breaking_re` is now `^[a-z]+(\([^)]+\))?!:`, symmetric with `feat_re` / `fix_re`. Kept the broader `<type>!:` shape rather than the closed whitelist so revert!/chore! etc. still trigger major per spec; regression-guarded by `test_uppercase_subject_does_not_match` and `test_unknown_type_does_not_match_even_with_bang` in tests/test_release_semver.py.

scripts/release.sh
aidanns: Claude: Done! In 48000de — `compute_bump` now writes `git log` output to a tempfile and checks the exit code explicitly; process-substitution is gone. Regression test: `test_invalid_range_returns_failure` asserts a malformed range now exits non-zero with a clear `semver: failed to list commits` message instead of silently printing `none`.

scripts/release.sh
aidanns: Claude: Done! In c6c1925 — `apply_bump` detects `v0.*` last tag and demotes major→minor per SemVer 2.0.0 §4 (tests: `v0.1.0 feat!:` → v0.2.0, `v0.9.5 feat!:` → v0.10.0, `v1.2.3 feat!:` → v2.0.0 unchanged). Graduation is explicit: `task release -- 1.0.0`. Documented in CONTRIBUTING.md's Version Resolution table.

scripts/release.sh
aidanns: Claude: Done! In 48000de — both cases handled. `git log --first-parent` keeps merge-commit subjects only and ignores the merged branch's intermediate commits. Revert commits: `body=${body%%This reverts commit *}` strips the appended reverted-message so a lone `git revert` of a BREAKING commit no longer promotes itself to a major bump (regression-guarded by `test_pure_revert_does_not_re_trigger_breaking`).

scripts/release.sh
aidanns: Claude: Done! In 48000de — `compute_bump` / `apply_bump` extracted to `scripts/lib/semver.sh`, exercised by `tests/test_release_semver.py` (29 cases: happy path, regex asymmetry, multi-line bodies, 0.x demotion, reverts, invalid ranges, and shellcheck/shfmt gates on the lib itself).

CONTRIBUTING.md
aidanns: Claude: Done! In c6c1925 — the table entry now calls out both footer variants (`BREAKING CHANGE:` / `BREAKING-CHANGE:`) and the lowercase-`<type>!:` form is the whitelist the script actually uses. Kept the 2-column table; GitHub renders it fine and mdformat doesn't reflow it.

scripts/release.sh
aidanns: Claude: Done! In 48000de — single `git log --first-parent -z --format='%s%n%b'` call per range, parsed in-bash from a tempfile. No more N×3 git invocations per range. Bundled with the pipefail fix since they share the same parsing path.

Issue #28, opened 2026-04-17
Document stable error taxonomy as public API


aidanns:
Source: `.claude/instructions/service-design.md` — Stable error taxonomy.

The server returns error strings (`invalid_token`, `token_expired`, `family_revoked`, `scope_denied`, `malformed_request`, `refresh_token_reuse_detected`, etc.) without a documented stability contract. Standard treats these as public API.

**Acceptance:**
- `design/DESIGN.md` (or dedicated doc) enumerates every error code returned by each endpoint with its HTTP status, meaning, and stability guarantee.
- Schema-contract tests assert the error-code set remains stable (mirrors the audit-log pinning issue).
- Covers agent-auth server and things-bridge server.

**Deterministic regression check:**
- `scripts/verify-standards.sh` dispatches to the error-code contract test, which asserts the documented error-code set per endpoint matches what the server actually emits.

aidanns:
Claude: Closing — all acceptance criteria are met on `main`.

- `design/error-codes.md` enumerates every error code per endpoint with HTTP status, meaning, and an explicit stability guarantee ("stable within a major version; rename/remove requires a major bump"). Covers both `agent-auth` and `things-bridge`, including health and server-wide codes.
- `tests/test_error_taxonomy.py` exercises every documented code end-to-end via the in-process server fixture style.
- `scripts/verify-standards.sh` dispatches to the test file and fails if any documented error code is missing coverage.
- `design/DESIGN.md` links to `error-codes.md` from the API Versioning Policy section.

Landed primarily in #126; verified against the current `main` while auditing the `1.0/api-stability` queue.

Issue #27, opened 2026-04-17
Adopt URL-versioned API namespace (/v1/)


aidanns:
Source: `.claude/instructions/service-design.md` — API versioning strategy.

Current endpoints are under `/agent-auth/...` and `/things-bridge/...` with no version segment. Standard requires URL-versioned APIs with a documented versioning policy.

**Acceptance:**
- All HTTP endpoints move to `/agent-auth/v1/...` and `/things-bridge/v1/...`.
- Versioning policy documented in `design/DESIGN.md` (when we break compatibility, what constitutes a breaking change, deprecation window).
- Clients (`things-cli`, test harnesses) updated.
- Health and metrics endpoints remain unversioned by convention (document the exception).

**Deterministic regression check:**
- `scripts/verify-standards.sh` asserts every registered HTTP route (outside `/healthz` and `/metrics`) matches `^/(agent-auth|things-bridge)/v\d+/`.

Issue #24, opened 2026-04-17
Migrate configuration format from JSON to YAML


aidanns:
Source: `.claude/instructions/service-design.md` — Human-editable config defaults to YAML; `.claude/instructions/python.md` — PyYAML safe_load.

`src/agent_auth/config.py` currently loads `config.json`. Standard requires YAML for human-editable config (JSON lacks comments, trailing commas, readable grouping). `pyyaml` is already a declared dependency.

**Acceptance:**
- `config.py` reads `config.yaml` via `yaml.safe_load`.
- things-bridge config (`~/.config/things-bridge/config.yaml` — already YAML) remains unchanged.
- things-cli credentials file migrated from JSON to YAML if applicable (README says YAML; confirm implementation matches).
- `config.json` read path dropped (clean break — no compatibility shim).
- Docs (README, CLAUDE.md) updated to reference YAML.

**Deterministic regression check:**
- `scripts/verify-standards.sh` greps `src/` and fails if any config-loading path references `config.json` or calls `json.load` / `json.loads` on a config file.

Issue #20, opened 2026-04-17
Pin audit-log and structured-log schemas with contract tests


aidanns:
Source: `.claude/instructions/release-and-hygiene.md` — Structured output schemas.

The audit-log format in `src/agent_auth/audit.py` and any application log output consumed downstream are public APIs. Today they have implicit contracts only — the docstring on `AuditLogger` says the on-disk format is part of the public surface but there's no schema-pinning test, so a field rename wouldn't fail CI.

**Acceptance:**
- Explicit schema (e.g. JSON Schema or typed dataclass with a test-enforced emit contract) for each `event` kind in `audit.py`.
- Schema-contract tests assert field presence, names, and types for every audit event (validation_allowed, validation_denied, token_refreshed, token_reissued, token_revoked, reissue_denied, and bridge/CLI audit events when those land).
- Breaking schema changes require bumping a documented schema version.

**Deterministic regression check:**
- `scripts/verify-standards.sh` asserts a schema-contract test (e.g. `tests/test_audit_schema.py`) exists and references every documented audit event kind.

PR #126, opened 2026-04-20
feat: migrate config to YAML, version APIs at /v1/, and add error/audit contract tests

aidanns:
## Summary

Addresses four \`1.0/api-stability\` issues together:

- **#24 — Config YAML migration**: \`config.json\` → \`config.yaml\`; loading switched from \`json.load\` to \`yaml.safe_load\`. Integration test fixtures updated to write YAML. Old \`docker/config.test.json\` deleted in favour of \`docker/config.test.yaml\`.
- **#27 — URL-versioned APIs**: All HTTP endpoints on both servers moved to \`/v1/\` namespace (e.g. \`/agent-auth/v1/validate\`, \`/things-bridge/v1/todos\`). Health endpoints (\`/agent-auth/health\`, \`/things-bridge/health\`) stay unversioned per convention. API Versioning Policy added to \`design/DESIGN.md\`. \`src/things_bridge/authz.py\` (the production validation client) also updated.
- **#28 — Stable error taxonomy**: All 17 error codes documented in \`design/error-codes.md\`. Contract tests in \`tests/test_error_taxonomy.py\` exercise every code end-to-end using in-process server fixtures. \`verify-standards.sh\` now fails if any code is missing coverage.
- **#20 — Audit log schema**: Contract tests in \`tests/test_audit_schema.py\` pin the schema for all 11 audit event kinds. \`verify-standards.sh\` now fails if any event kind is missing coverage.

## Test plan

- [x] \`uv run pytest tests/ --ignore=tests/integration\` passes (283 passed, 3 skipped)
- [x] \`uv run pytest tests/test_error_taxonomy.py\` passes (32 tests)
- [x] \`uv run pytest tests/test_audit_schema.py\` passes (14 tests)
- [x] \`uv run pytest tests/test_things_bridge_authz.py\` passes including path assertion
- [x] Pre-commit hooks pass (ruff, shfmt, shellcheck, mdformat, keep-sorted)
- [x] Docker integration tests pass (CI run 24665149982)
- [x] \`scripts/verify-standards.sh\` passes (CI: all non-Test workflows green)

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ config, and images built inside DinD are not visible to the host
 `docker` CLI.
 
 The integration-test fixture chmods the bind-mounted config directory
-to `0755` and `config.json` to `0644` so the container user (UID 1001,
+to `0755` and `config.yaml` to `0644` so the container user (UID 1001,
 see `docker/Dockerfile.test`) can read it regardless of the host
 tmpdir's default mode or the host runner's UID.
 
--- a/design/DESIGN.md
+++ b/design/DESIGN.md
@@ -55,15 +55,15 @@ The subprocess contract is stable and publicly documented so the bridge can adop
 
 Read-only endpoints. Read endpoints require the `things:read` scope; the health endpoint requires `things-bridge:health` (mirroring the agent-auth health pattern — see `GET /agent-auth/health` below).
 
-| Method | Path                                                     | Scope                  | Description                                   |
-| ------ | -------------------------------------------------------- | ---------------------- | --------------------------------------------- |
-| GET    | `/things-bridge/todos?list=&project=&area=&tag=&status=` | `things:read`          | List todos, optionally filtered               |
-| GET    | `/things-bridge/todos/{id}`                              | `things:read`          | Fetch one todo by Things id                   |
-| GET    | `/things-bridge/projects?area=`                          | `things:read`          | List projects, optionally filtered by area id |
-| GET    | `/things-bridge/projects/{id}`                           | `things:read`          | Fetch one project by Things id                |
-| GET    | `/things-bridge/areas`                                   | `things:read`          | List all areas                                |
-| GET    | `/things-bridge/areas/{id}`                              | `things:read`          | Fetch one area by Things id                   |
-| GET    | `/things-bridge/health`                                  | `things-bridge:health` | Liveness / readiness probe                    |
+| Method | Path                                                        | Scope                  | Description                                   |
+| ------ | ----------------------------------------------------------- | ---------------------- | --------------------------------------------- |
+| GET    | `/things-bridge/v1/todos?list=&project=&area=&tag=&status=` | `things:read`          | List todos, optionally filtered               |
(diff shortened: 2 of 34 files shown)


aidanns (commented):
Claude Reviewer: I found a critical correctness bug and two smaller issues.

---

## Critical: `src/things_bridge/authz.py` not updated — production bridge will break

`AgentAuthClient.validate()` (line 42 of `src/things_bridge/authz.py`) still POSTs to the old unversioned path `/agent-auth/validate`. Since the agent-auth server now only registers `/agent-auth/v1/validate`, every token-validation call from a real things-bridge deployment will receive `404 {"error": "not_found"}`. `AgentAuthClient` treats any non-200/401/403 as `AuthzUnavailableError`, so every single request through the bridge will fail with a 502.

This file was not changed in the PR and is therefore not in the diff. The test suite doesn't catch it because `test_things_bridge_authz.py`'s `_Responder` mock accepts any POST path and returns the configured status without inspecting the URL.

**Fix needed in `src/things_bridge/authz.py` line 42:**
```python
# Before:
"/agent-auth/validate",
# After:
"/agent-auth/v1/validate",
```

Also add a path-check assertion in `test_validate_allows_on_success` (or a dedicated test) so this class of regression is caught by CI:
```python
assert sent_path == "/agent-auth/v1/validate"
```

---

## Minor: Stale reference in `design/decisions/0004-docker-integration-tests.md`

Line 75 of the ADR still reads `docker/config.test.json` (which was deleted in this PR). It should be `docker/config.test.yaml`.

---

## Minor: `verify-standards.sh` route-versioning check does not cover things-bridge routes

The new check (around line 705 of `scripts/verify-standards.sh`) greps for `self.path ==`:

```bash
unversioned_routes=$(grep -rn 'self\.path ==' src/agent_auth/server.py src/things_bridge/server.py ...)
```

In `src/things_bridge/server.py`, routes are matched against a local variable `path` (extracted via `urlsplit`), not `self.path`. The grep produces zero matches for things-bridge — so the check silently passes without verifying any things-bridge routes at all.

Currently all things-bridge routes happen to be correctly versioned, so this doesn't cause a CI failure. But the check gives a false sense of security: someone could accidentally add an unversioned things-bridge route and this guard would not catch it.

**Suggested fix:** Extend the grep pattern to also match `path ==` and `path.startswith(` in the things-bridge server, or note the limitation in a comment.

aidanns (commented):
Claude Reviewer: Test comment to verify posting works.

aidanns (commented):
Claude: Done! All three issues addressed in e476fc1:
- Critical: `src/things_bridge/authz.py` updated to POST to `/agent-auth/v1/validate`; `test_things_bridge_authz.py` now asserts the exact path.
- Minor: ADR 0004 reference updated from `config.test.json` to `config.test.yaml`.
- Minor: `verify-standards.sh` route-versioning grep extended to match both `self.path ==` (agent-auth) and bare `path ==`/`path.startswith(` (things-bridge).

Issue #128, opened 2026-04-20
ci: migrate release-please from PAT to GitHub App token


aidanns:
Raised during review of #106/#110/#111/#115 bundle.

\`release-please-action\` needs a \`RELEASE_PLEASE_TOKEN\` secret (PAT or GitHub App token) because tags created with the default \`GITHUB_TOKEN\` do not fire downstream \`on: push: tags\` workflows, which would break the chain into \`release-publish.yml\`. ADR 0014 calls this out as the one long-lived credential in the trust root — weaker than the keyless cosign posture elsewhere.

## Scope

- Create a GitHub App with minimum required permissions (\`contents: write\`, \`pull-requests: write\`) installed on this repository.
- Use \`actions/create-github-app-token\` (or similar) to mint an installation token per workflow run.
- Replace \`secrets.RELEASE_PLEASE_TOKEN\` with the app-generated token in \`.github/workflows/release-please.yml\`.
- Remove the PAT secret and document the App installation in \`CONTRIBUTING.md\`.

## Related

- #106 introduced the autorelease workflow.
- ADR 0014 (Consequences, bullet "Release Please requires a PAT").
- \`CONTRIBUTING.md\` § *Release process → Default path* currently documents the PAT requirement.

aidanns:
Superseded by the semantic-release migration (see `.claude/worktrees/semantic-release-migration` / `.releaserc.mjs` as referenced in CLAUDE.md). The GitHub-App-token workflow landed in #187 is being rolled into the semantic-release direction rather than pursued further as a release-please migration.

Issue #127, opened 2026-04-20
ci: pin release-affecting GitHub Actions to commit SHAs


aidanns:
Raised during review of #106/#110/#111/#115 bundle.

The new release workflows hold `id-token: write` and publish signed artefacts:

- \`.github/workflows/release-please.yml\` — \`googleapis/release-please-action@v4\`
- \`.github/workflows/release-publish.yml\` — \`astral-sh/setup-uv@v5\`, \`anchore/sbom-action@v0\`, \`sigstore/cosign-installer@v3\`
- \`.github/workflows/reuse.yml\` — indirectly via \`./.github/actions/setup-toolchain\`

All are pinned to floating major tags. A compromised action release could siphon the runner's OIDC token or replace the signing binary. The repo's existing convention is floating-major pins across all workflows; hardening the release-adjacent ones (which have write permissions, unlike \`check.yml\`/\`test.yml\`) is the higher-priority subset.

## Scope

- Convert every \`uses: foo/bar@vX\` line in \`release-please.yml\`, \`release-publish.yml\`, \`reuse.yml\`, and \`.github/actions/setup-toolchain/action.yml\` to commit-SHA pins (e.g. \`uses: sigstore/cosign-installer@<sha>\`).
- Add Dependabot entries for each action so the pins auto-refresh into review-gated PRs.
- Document the policy in \`CONTRIBUTING.md\` or a new \`.claude/instructions/tooling-and-ci.md\` section.

## Related

- #106 / #110 / #111 / #115 introduced the workflows without SHA pins.
- ADR 0014 (design/decisions/0014-release-supply-chain.md) can be updated with the pinning policy.

Issue #109, opened 2026-04-19
supply chain: attest SLSA build provenance on releases


aidanns:
## Context

SLSA (slsa.dev) defines four levels of build-provenance guarantees.
Level 3 requires a hosted, hermetic build producing signed,
verifiable provenance — none of which is possible from the local
\`task release\` path tracked in #18. #106 introduces the
Release-Please-driven CI release; this issue lands the SLSA
generator inside it.

## Scope

- Use \`slsa-framework/slsa-github-generator\` in the tag-triggered
  publish workflow (#106) to emit in-toto provenance for each release
  artifact (sdist, wheel).
- Publish provenance alongside artifacts on the GitHub release.
- Document verification (\`slsa-verifier verify-artifact\`) in
  \`CONTRIBUTING.md\` or a new \`SUPPLY_CHAIN.md\`.
- Target SLSA Level 3 for 1.0; accept Level 2 for 0.x.

## Acceptance

- [ ] Each release attaches SLSA provenance.
- [ ] README / SECURITY.md reference the verification command.
- [ ] Target level recorded and achieved.

## Related

- #106 — autorelease workflow (this lands inside it)
- #18 — local release ritual

PR #132, opened 2026-04-20
feat(release): signed SBOMs, autorelease via Release Please, REUSE compliance

### LINKED ISSUE #106: ci: add autorelease workflow (Release Please + signed artifacts)
## Context

#18 tracks a local `task release` target: an operator runs it on their
laptop, which validates, bumps, tags, pushes, and creates the GitHub
release. That closes the floor (release is reproducible and scripted),
but it still requires a human at the wheel and gives us no supply-chain
provenance — the tag, sdist, and GitHub release are all produced on a
developer machine.

An autorelease workflow on GitHub Actions would:

- Drive versioning from the conventional-commit history already in use
  (#14 tracks the `CHANGELOG.md` feeder), so cutting a release becomes
  "merge the release PR" rather than "remember to run `task release`".
- Move artifact production onto an ephemeral runner where we can attach
  provenance (SLSA), signatures (Sigstore/cosign), and an SBOM (SPDX via
  Syft) — none of which are realistic to produce from a laptop.

## Proposal

Adopt **Release Please** (`googleapis/release-please-action`) as the
driver:

1. On every push to `main`, the action parses conventional commits and
   maintains a single open "chore: release vX.Y.Z" PR that updates
   `CHANGELOG.md` and the version file.
2. Merging that PR creates a Git tag and a draft GitHub release.
3. A second workflow, triggered on the tag push, builds distributions
   (sdist + wheel via `uv build`), generates an SBOM, signs artifacts
   with cosign, attests SLSA provenance, and attaches everything to the
   GitHub release.

Alternative driver considered: `semantic-release`. Rejected as too
aggressive for this project — it cuts a release on every qualifying
merge, whereas Release Please batches changes into a reviewable release
PR, which fits the existing review-gated flow and the QM/SIL posture
tracked in #21.

## Relationship to #18

Not a duplicate. Two viable outcomes:

- **Keep both**: `task release` remains the break-glass path (offline,
  emergency hotfix); CI is the default.
- **Supersede #18**: once the CI path is solid, collapse `task release`
  into a much smaller local validator and close #18.

Decide at implementation time.

## Scope

- New workflow `.github/workflows/release-please.yml` running on every
  push to `main`.
- New workflow `.github/workflows/release-publish.yml` triggered on
  tag pushes matching `v*`, producing:
  - sdist + wheel (`uv build`)
  - SBOM (`anchore/sbom-action` → SPDX)
  - Artifact signatures (`sigstore/cosign-installer` + `cosign sign-blob`)
  - SLSA provenance (`slsa-framework/slsa-github-generator`)
- All attached to the GitHub release created by Release Please.
- Minimal `GITHUB_TOKEN` / OIDC permissions on each job (no long-lived
  PATs).
- `CONTRIBUTING.md` updated to document the new flow (what gets
  released, how to opt a commit out, how to cut a manual release if CI
  is down).

## Considerations

- **Tag protection**: once autorelease is the default, the `v*` tag
  namespace should be write-protected to CI only — ties into #93.
- **Required checks on the release PR**: the Release Please PR should
  gate on the standard test/check workflows so a release can't merge
  with a red CI.
- **Prerelease channel**: decide up front whether `v0.x` tags are
  cut automatically or held behind a manual approval step. Pre-1.0 I'd
  lean manual; post-1.0 fully automatic.
- **setuptools-scm**: still the source of truth for the installed
  version at runtime; Release Please just manages the tag.

## Acceptance

- [ ] Release PR workflow auto-opens/updates on pushes to `main`.
- [ ] Merging the release PR produces a signed tag and a GitHub release.
- [ ] Release includes sdist, wheel, SBOM, signatures, SLSA provenance.
- [ ] `CONTRIBUTING.md` documents the flow and the manual-override path.
- [ ] Decision recorded on whether #18 is retained as the break-glass
  path or closed.

### LINKED ISSUE #110: supply chain: sign release artifacts with Sigstore/cosign
## Context

Release artifacts (sdist, wheel, SBOM, provenance) should be signed
so consumers can verify authenticity without the project managing
long-lived keys. Sigstore / cosign provides keyless signing backed
by short-lived OIDC-issued certs and the Rekor transparency log —
ideal for GitHub Actions where \`id-token: write\` already
authenticates the workflow.

Companions #106 (autorelease) and the SLSA and SBOM issues filed in
this batch. Commit signing is already enforced (PR #92).

## Scope

- Install \`sigstore/cosign-installer\` in the tag-triggered publish
  workflow (#106).
- Sign each artifact: \`cosign sign-blob --yes <artifact>\`.
- Attach \`.sig\` / \`.pem\` alongside artifacts on the GitHub
  release.
- Document verification in \`CONTRIBUTING.md\`.

## Acceptance

- [ ] Every release artifact has a Sigstore signature.
- [ ] Verification command documented.
- [ ] \`SECURITY.md\` records the signing posture.

### LINKED ISSUE #111: supply chain: publish SPDX SBOM with every release
## Context

A Software Bill of Materials enumerates the exact dependency closure
baked into a release artifact. \`pip-audit\` already scans runtime
deps (#82), but a scanner is not an artifact — downstream consumers
and compliance frameworks (SLSA, SSDF, EO 14028) want the SBOM
alongside the release so they can run their own queries.

## Scope

- In the tag-triggered publish workflow (#106), generate an SPDX
  SBOM from the built sdist / wheel using \`anchore/sbom-action\`
  (Syft).
- Attach \`<artifact>.spdx.json\` to the GitHub release.
- Sign the SBOM with cosign so it is tamper-evident.

## Acceptance

- [ ] Every release attaches an SPDX SBOM.
- [ ] SBOM is signed (via the cosign issue in this batch).
- [ ] \`SECURITY.md\` records SBOM location and format.

### LINKED ISSUE #115: licensing: adopt the REUSE specification for machine-readable licensing
## Context

A top-level \`LICENSE.md\` (#15) declares the project license but
doesn't tell a tool which file carries which licence. The REUSE
Specification (reuse.software) standardises per-file SPDX headers
and a \`REUSE.toml\` for files that can't carry headers, plus a
\`reuse\` linter CI check. This becomes important as vendored
snippets or differently-licensed assets accrue.

## Scope

- Add SPDX-License-Identifier headers to source files (covered by
  \`reuse annotate\`).
- Add \`REUSE.toml\` for unheaderable files (binary fixtures,
  LICENSE itself).
- CI check: \`reuse lint\` in \`.github/workflows/check.yml\`.
- Render REUSE badge in README.

## Acceptance

- [ ] \`reuse lint\` passes on \`main\`.
- [ ] CI fails a PR that drops a header.
- [ ] Badge rendered in README.

## Related

- #15 — LICENSE.md (precondition for REUSE compliance)

aidanns:
## Summary

Bundles four release-artefact issues into one pipeline — the pieces are tightly coupled so splitting them would thrash the CI surface:

- **#106** — Release Please workflow opens/maintains a single release PR on every push to `main`; merging it pushes a `vX.Y.Z` tag.
- **#110** — Tag-triggered publish workflow signs every release artefact and SBOM with keyless Sigstore cosign (no long-lived keys).
- **#111** — Per-artefact SPDX JSON SBOMs generated via `anchore/sbom-action` (Syft) and signed alongside the artefacts.
- **#115** — REUSE 3.3 compliance: SPDX headers on every tracked file, `REUSE.toml` for unheaderable files, `fsfe/reuse-action` lint in CI, REUSE status badge in README.

Rationale, trade-offs, and alternative-analysis live in [ADR 0015](design/decisions/0015-release-supply-chain.md). Verification recipe for downstream consumers is in [SECURITY.md § Supply-chain artifacts](SECURITY.md#supply-chain-artifacts).

> **Merge note**: `main` was merged in after opening this PR; our ADR was renumbered from 0014 → 0015 because the management-endpoint-auth ADR (#97) landed at 0014 first. SPDX headers were added to files brought in from main so `reuse lint` stays clean.

## Requires before the first release

One-time setup, documented in `CONTRIBUTING.md`:

- Create a fine-grained PAT (or GitHub App token) with `contents: write` + `pull-requests: write` on this repo; save as the **`RELEASE_PLEASE_TOKEN`** repository secret. Tags created with the default `GITHUB_TOKEN` do not fire downstream workflows, which would silently break the release-publish chain.

Migrating this to a GitHub App installation token is tracked in #128.

## Test plan

- [x] `scripts/reuse-lint.sh` passes locally
- [x] `task check` passes (lint + format --check + integration isolation)
- [x] `task test` passes (274 passed, 3 skipped)
- [x] `scripts/verify-standards.sh` passes
- [ ] CI `Check`, `Test`, `REUSE`, `Verify Design`, `Verify Function Tests`, `Verify Standards`, `Security` workflows all green
- [ ] REUSE status badge renders `✓ compliant` on the PR preview
- [ ] Release Please workflow runs on merge and opens a release PR
- [ ] After configuring `RELEASE_PLEASE_TOKEN` and merging the release PR: release-publish workflow runs, attaches sdist + wheel + SBOMs + `.sig.bundle` signatures
- [ ] `cosign verify-blob` recipe from SECURITY.md succeeds against the published artefacts

The last three can only be verified post-merge with a real release. Recommend cutting a test release (e.g. `v0.1.1` or `v0.2.0-rc.1`) immediately after this lands to shake out any CI-only issues, then closing #106/#110/#111 based on that run.

## Follow-ups filed

- #127 — SHA-pin release-affecting GitHub Actions
- #128 — migrate `release-please-action` from PAT to GitHub App token

Related existing follow-ups: #109 (SLSA provenance), #93 (tag protection), #18 (fate of `task release`).

Closes #106, closes #110, closes #111, closes #115.


.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          enable-cache: true
+
+      - name: Build sdist and wheel
+        run: uv build
+
+      - name: Generate SPDX SBOM for sdist
+        uses: anchore/sbom-action@v0
+        with:
+          file: dist/*.tar.gz
```
aidanns: Claude Reviewer: [blocker] `anchore/sbom-action` does **not** expand globs in the `file:` input. The action passes the value directly to `core.getInput("file")` and then invokes `syft file:${input.file}`; Syft resolves that argument with `os.Stat` on the literal path (see [syft/source/filesource/file_source.go](https://github.com/anchore/syft/blob/main/syft/source/filesource/file_source.go)). With `file: dist/*.tar.gz` / `file: dist/*.whl`, syft will error with "no such file or directory" on the first run of the publish workflow.

Fix options:
- Resolve the exact filename in a prior `shell` step (e.g. `SDIST=$(ls dist/*.tar.gz)` then pass `file: ${{ steps.resolve.outputs.sdist }}`), or
- Use `path: dist` with `artifact-name:` overrides and split the jobs per artefact via a matrix on exact filenames.

This blocks the whole SBOM→sign→upload chain, so the first tag push after merge will fail.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: ./.github/actions/setup-toolchain
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run reuse lint
+        run: task reuse-lint
```
aidanns, line 24: Claude Reviewer: [important] The ADR (§ Decision → REUSE bullet) and the implementation plan both state that `fsfe/reuse-action@v5` gates PRs, but the actual `.github/workflows/reuse.yml` runs `task reuse-lint` via the local venv — no `fsfe/reuse-action` anywhere in the workflows. Either switch the workflow to use the action (what the ADR records as the decision) or update ADR 0015 and the plan to reflect the `task reuse-lint` choice. This isn't cosmetic — an ADR whose "Decision" section doesn't match the delivered pipeline defeats the point of the ADR.

CONTRIBUTING.md
```diff
@@ -87,18 +93,57 @@ signing are sibling requirements, not the same thing.
...
+  builds the sdist and wheel with `uv build`, generates an SPDX SBOM
+  per artifact with Syft, signs each artifact and SBOM with keyless
+  cosign (Sigstore OIDC), and attaches everything to the GitHub
+  release. Verification recipe: see
+  [`SECURITY.md` → Supply-chain artifacts](SECURITY.md#supply-chain-artifacts).
+- **Break-glass — `task release` (local)**: runs `scripts/release.sh`
+  on your laptop. Use when CI is unavailable or a release has to be
+  cut without waiting for the runner. Produces the tag and GitHub
+  release but **not** the SBOM or cosign signatures — those only come
+  from the CI path. Documented below.
```
aidanns: Claude Reviewer: [important] CONTRIBUTING.md § *Break-glass — `task release`* claims the local path produces "**not** the SBOM or cosign signatures — those only come from the CI path". That's misleading. `scripts/release.sh` creates and pushes a `v*` tag using the developer's own git credentials (not `GITHUB_TOKEN`), which **will** trigger `release-publish.yml` — including SBOM generation and cosign signing. The actual gap versus the default path is:

1. The local `gh release create` runs immediately and publishes a release **before** the tag workflow finishes uploading, so downstream consumers may briefly see a release without signatures.
2. If the developer's push credential lacks `workflow` scope or the tag arrives via a force-push, the workflow fires as normal.

Suggest rewording to something like "the local path creates the release up front; signed SBOMs arrive asynchronously from the same tag-triggered workflow the CI path uses, provided the push credential has workflow-dispatch-on-tag permission." Otherwise a reader will think the break-glass path is unsigned, which is wrong.

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,70 @@
...
+          for f in dist/*.tar.gz dist/*.whl dist/*.spdx.json; do
+            cosign sign-blob --yes --bundle "${f}.sig.bundle" "${f}"
+          done
+
+      - name: Upload artifacts to GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ github.ref_name }}
+        shell: bash
+        run: gh release upload "${TAG}" dist/* --clobber
```
aidanns: Claude Reviewer: [important] `gh release upload "${TAG}" dist/* --clobber` uploads **everything** currently in `dist/`, including the per-build SBOM output files (`dist/sdist.spdx.json`, `dist/wheel.spdx.json`) that the earlier `anchore/sbom-action` steps wrote there with generic names. That's probably the intent — but two observations:

1. The SECURITY.md verification recipe instructs consumers to download `sdist.spdx.json` and `wheel.spdx.json`, which are also the literal filenames on the release. Those names are opaque (they don't tell a consumer which artefact they describe without reading the SBOM). Consider `agent_auth-${VERSION}.tar.gz.spdx.json` / `agent_auth-${VERSION}-py3-none-any.whl.spdx.json` for parity with Syft's own conventions and for self-describing downloads.
2. If a future change adds any other file under `dist/` (stubs, wheels for multiple Python versions, PEP 740 attestations, `.asc` etc.), `dist/*` uploads it silently. Consider an explicit list or `dist/{*.tar.gz,*.whl,*.spdx.json,*.sig.bundle}` to keep the uploaded set auditable.

SECURITY.md
```diff
@@ -221,6 +227,61 @@ The table below lists selected controls with their current implementation status
...
+# Download the artifact, its SBOM, and both signature bundles from the release.
+# Filenames use PEP 625 normalization (underscore, not hyphen) — that is the
+# form uv build emits.
+: "${TAG:?set TAG=vX.Y.Z to the release tag you downloaded}"
+
+for f in agent_auth-${TAG#v}.tar.gz agent_auth-${TAG#v}-py3-none-any.whl \
+         sdist.spdx.json wheel.spdx.json; do
+  cosign verify-blob \
+    --bundle "${f}.sig.bundle" \
+    --certificate-identity-regexp "^https://github\.com/aidanns/agent-auth/\.github/workflows/release-publish\.yml@refs/tags/${TAG}$" \
```
aidanns: Claude Reviewer: [important] The `--certificate-identity-regexp` has literal `.` characters inside the interpolated `${TAG}` section that are **not** escaped — only the path portion is. If `TAG=v0.2.0`, the interpolated regex becomes `...@refs/tags/v0.2.0$`, which matches `v0-2-0`, `v0x2x0`, etc. — widening the acceptance set beyond what you want.

Safer form: escape the tag before interpolating, e.g.

```bash
TAG_ESCAPED=$(printf '%s\n' "$TAG" | sed 's/[.[\\^$*]/\\&/g')
cosign verify-blob \
  --certificate-identity-regexp "^https://github\.com/aidanns/agent-auth/\.github/workflows/release-publish\.yml@refs/tags/${TAG_ESCAPED}$" \
  ...
```

Or, since the tag is known exactly, use `--certificate-identity` (exact match) instead of `--certificate-identity-regexp` and skip the escaping entirely.

Low severity in practice (`.` only matches a single char so the extra accepted strings are unlikely to exist as valid Sigstore subjects), but the recipe is the authoritative downstream snippet — best to get it right.

SECURITY.md
```diff
@@ -221,6 +227,61 @@ The table below lists selected controls with their current implementation status
...
+  ephemeral job; a compromised runner could produce a consistently
+  signed malicious bundle. Residual risk accepted.
+- Sigstore's transparency log (Rekor) records every signature; a
+  signature that is not present in Rekor is a detection signal.
+- Long-lived signing keys are **not** used — cosign consumes the
+  runner's OIDC token per-release, so there is no key to steal or
+  revoke out-of-band.
+
+SLSA build provenance is tracked separately in
+[#109](https://github.com/aidanns/agent-auth/issues/109).
```
aidanns, line 329: Claude Reviewer: [important] SECURITY.md § *Supply-chain artifacts → Trust boundary* leaves two real residual risks out of the threat model:

1. **Rekor non-availability / non-inclusion**: the trust boundary block says "a signature that is not present in Rekor is a detection signal", but `cosign verify-blob --bundle` **requires** Rekor inclusion by default (the bundle embeds the transparency-log entry). If Rekor is down at verification time or the entry is withheld, verification fails closed. This is worth stating explicitly alongside the "compromised runner" risk rather than only implying it.
2. **Fulcio CA compromise**: the entire keyless model trusts the Sigstore PKI (Fulcio root). This is the other half of "no key to steal" — a compromised Fulcio would let an attacker mint a cert against any OIDC identity, including the runner's. Rekor transparency limits the window but doesn't prevent a window. Worth one line.

These matter because the STRIDE table above is thorough on spoofing/tampering/repudiation for the runtime token system but then treats the build/release boundary as a single "residual risk accepted" bullet. The threat-model rigour applied to the tokens code should extend to the supply chain now that it has its own trust boundary.

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,70 @@
...
+name: Release Publish
+
+on:
+  push:
+    tags: ["v*"]
+
+permissions:
+  contents: write
+  id-token: write
+
```
aidanns, line 16: Claude Reviewer: [important] There is no `concurrency:` key on this workflow. If a second `v*` tag is pushed before the first publish finishes (e.g. an immediate re-tag / force-push), two runs execute in parallel. Both will call `gh release upload ... --clobber` against possibly the same release, and their SBOM + signing steps race on `dist/`. `--clobber` makes this safe-ish for the release assets themselves but it's still worth a bounded `concurrency` group (`concurrency: release-publish-${{ github.ref }}`, `cancel-in-progress: false`) so the runs serialise on the tag. Not a blocker for the first release but cheap to add before the workflow is exercised in anger.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
+#
+# SPDX-License-Identifier: MIT
+
+name: Release Please
+
+on:
+  push:
+    branches: [main]
```
aidanns, line 9: Claude Reviewer: [nit] Release Please runs only on `push` to `main`, but the action itself checks the state of the repository against `.release-please-manifest.json` and the most recent `v*` tag. There's no `workflow_dispatch` trigger documented, so if the release PR state ever desynchronises (e.g. someone force-pushes the branch, or the maintainer wants to re-materialise the PR without pushing an empty commit), there's no way to rerun this workflow by hand.

Consider adding:

```yaml
on:
  push:
    branches: [main]
  workflow_dispatch:
```

Cheap and mirrors what Release Please's own docs recommend.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  reuse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: ./.github/actions/setup-toolchain
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
aidanns, line 21: Claude Reviewer: [nit] This workflow runs the setup-toolchain composite (which bootstraps `uv`) only to call `task reuse-lint`, which then runs `uv run --no-sync reuse lint`. That's a lot of scaffolding for one command, and it's also the only CI path that uses the project venv to run a non-project tool.

Two lighter alternatives if you also fix the ADR-vs-workflow mismatch flagged separately:

- `uses: fsfe/reuse-action@v5` — what ADR 0015 documents; no `uv` / venv / toolchain needed.
- `uvx reuse lint` — avoids touching the project venv at all. Skip setup-toolchain and just install uv.

Either takes this workflow from ~30s to ~5s on the hot path and removes a dependency on the rest of the toolchain being healthy for a REUSE lint to pass.

aidanns:
Claude Reviewer: [nit] Cross-file observation — the follow-up issues #127 and #128 (filed during this PR) both reference "ADR 0014" in their bodies, but the ADR landed as 0015 after the merge-from-main renumber. Neither is in-scope for this PR to fix, but a quick edit on those two issue bodies keeps the trail navigable.

Also worth a scan before merge: the PR description's test-plan items are mostly of the form "only verifiable post-merge with a real release" — recommend cutting a `v0.2.0-rc.0` (or similar) as the **very first** merge-to-main activity after this lands, before any other PR merges into main, so the first Release Please PR isn't chasing a moving target. If the publish workflow is broken (see the `sbom-action` glob blocker), you want to find out on a throwaway RC tag, not on the real v0.2.0.

.github/workflows/release-publish.yml
aidanns: Claude: Done! Resolved dist/ filenames in a prior shell step and passed the exact paths to anchore/sbom-action. Also renamed SBOMs to <artefact>.spdx.json for parity with Syft conventions.

.github/workflows/reuse.yml
aidanns: Claude: Took the opposite fix — updated ADR 0015 to reflect that the workflow uses task reuse-lint deliberately, and documented fsfe/reuse-action as a considered-and-rejected alternative (version drift vs local). The delivered pipeline now matches the recorded decision.

CONTRIBUTING.md
aidanns: Claude: Done! Rewrote the break-glass section to make clear that the tag push (via the developer credential, not GITHUB_TOKEN) does fire release-publish.yml — the difference from the default path is timing (gh release create publishes up front; assets arrive async) not signature coverage.

.github/workflows/release-publish.yml
aidanns: Claude: Done! gh release upload now takes an explicit list of the eight files (artefact + SBOM + their .sig.bundles). SBOM filenames are now agent_auth-<ver>.tar.gz.spdx.json / agent_auth-<ver>-py3-none-any.whl.spdx.json to be self-describing.

SECURITY.md
aidanns: Claude: Done! Switched to --certificate-identity (exact match) since the tag is known exactly — sidesteps the escaping issue entirely. Updated SECURITY.md verify recipe accordingly.

SECURITY.md
aidanns: Claude: Done! Expanded the trust-boundary section with explicit residual-risk entries for Rekor availability (verify-blob --bundle fails closed if the transparency log is unreachable) and Fulcio CA compromise (keyless model trusts the Sigstore PKI root).

.github/workflows/release-publish.yml
aidanns: Claude: Done! Added concurrency: release-publish-${{ github.ref }} with cancel-in-progress: false.

.github/workflows/release-please.yml
aidanns: Claude: Done! Added workflow_dispatch to release-please.yml's triggers.

.github/workflows/reuse.yml
aidanns: Claude: Kept task reuse-lint intentionally — the whole point of routing CI through the task runner is that local and CI run the same reuse version pinned by uv.lock. Switching to uvx or fsfe/reuse-action would reintroduce the drift the ADR now explicitly rejects. The ~25s overhead is the cost of that consistency.

Issue #129, opened 2026-04-20
ci: cache Dockerfile.test layers via buildx + GHA cache


aidanns:
## Context

Follow-up from #104 (split unit + integration CI into parallel jobs).
Parallelism alone satisfies #104's acceptance criterion of "strictly
lower wall-clock". The integration image build itself is still
uncached — every run pays the full build cost for
`docker/Dockerfile.test`.

`tests/integration/_support.py:build_test_image` (lines ~131-156)
shells out to classic `docker build`:

```python
subprocess.run(
    ["docker", "build", "-f", str(DOCKERFILE), "-t", tag, str(REPO_ROOT)],
    ...
)
```

No `--cache-from` / `--cache-to` flags, no buildx. The image rebuilds
from scratch every CI run.

## Proposal

1. Add `docker/setup-buildx-action@v3` with `install: true` to the
   `integration` job in `.github/workflows/test.yml`. `install: true`
   aliases `docker build` to `docker buildx build`, so the harness
   change stays small.
2. Extend `build_test_image` to accept optional cache args, and gate
   their inclusion on `CI=true` (or an explicit env var like
   `AGENT_AUTH_DOCKER_CACHE=gha`). In CI mode, add:

   ```
   --cache-from type=gha
   --cache-to type=gha,mode=max
   ```

3. Confirm on a PR that the second run sees a cache hit (log line
   `importing cache manifest from gha:...` etc.) and that image
   contents are identical.

## Considerations

- Cache is scoped to the repo by default; fine.
- Cache eviction: GHA cache is LRU with a 10 GB cap per repo; the
  integration image is small enough not to matter.
- Local dev: if `CI=true` isn't set, the harness keeps using classic
  `docker build` with no flags — no behavioural change locally.

## Acceptance

- [ ] `integration` CI job shows a cache import log on the second+
  run after this change lands.
- [ ] Local `task test -- --integration` still works without any
  CI-only env vars set.
- [ ] No regression in image correctness (integration tests still
  pass).

Issue #93, opened 2026-04-19
Enable branch protection on main to enforce commit signing


aidanns:
CONTRIBUTING.md documents that commits to \`main\` must be signed, but this is not yet enforced via GitHub branch protection rules — it currently relies on contributor honour.

**Acceptance:**
- Branch protection rule on \`main\` requires signed commits
- Unsigned commits are rejected by GitHub

**Context:** Added as a follow-up to PR #92 which added the commit-signing documentation.

PR #133, opened 2026-04-20
ci: split unit and integration tests into parallel jobs

### LINKED ISSUE #104: ci: split unit and integration tests into parallel workflows
## Context

`.github/workflows/test.yml` runs a single job that invokes
`task test -- --all`, which in turn runs the in-process unit suite
**then** the Docker-backed integration suite sequentially
(`scripts/test.sh`). The integration layer builds a throwaway image
(`docker/Dockerfile.test`) and spins up a fresh Compose project per
test, so its wall-clock cost dominates — and today every PR blocks on
that cost even when only fast unit assertions are needed for feedback.

Running the two layers in parallel on independent runners would cut
total PR latency to roughly `max(unit, integration)` instead of
`unit + integration`, and would make unit-test failures surface
earlier (before the image build completes).

## Proposal

Split the current job in two. Options:

1. **Two jobs in `test.yml`** (`unit` and `integration`) with a
   `tests` rollup job that `needs: [unit, integration]` so branch
   protection can require a single check.
2. **Two separate workflows** (`test-unit.yml`, `test-integration.yml`)
   triggered on the same events.

Option 1 is simpler and keeps the composite `setup-toolchain` action
single-sourced.

Each job:

- Shares `./.github/actions/setup-toolchain`.
- `unit`: `task test -- --unit`. No Docker needed.
- `integration`: `task test -- --integration`. Needs Docker; consider
  caching the base image layers between runs to avoid paying the full
  build cost every PR.

## Considerations

- **Required checks**: if branch protection currently requires the
  `test` check, update it to require the new rollup (or both jobs)
  before merging this change to avoid a gap where no test run is
  required. Ties into #93.
- **Matrix already planned**: #69 adds a macOS runner for the
  AppleScript path. Factor the split so adding a `macos` job later is
  a trivial extension rather than a restructure.
- **Docker cache**: explore `docker/setup-buildx-action` +
  `cache-from: type=gha` to reuse image layers across runs; the
  integration image is the biggest single cost.
- Keep `scripts/test.sh` unchanged — the split happens at the
  workflow level only, so local `task test -- --all` still works.

## Acceptance

- [ ] Unit and integration tests run in parallel on CI.
- [ ] Total PR wall-clock for the test workflow is strictly lower than
  the current serialized run on a representative PR.
- [ ] A single check (rollup job or updated branch-protection config)
  continues to gate merges.
- [ ] `scripts/test.sh` still supports `--unit`, `--integration`, and
  `--all` locally.

aidanns:
## Summary

- Splits `.github/workflows/test.yml` into parallel `unit` and `integration` jobs plus a `tests` rollup. Total wall-clock drops from `unit + integration` to roughly `max(unit, integration)`.
- No changes to `scripts/test.sh` or `Taskfile.yml` — local `task test -- --all`, `--unit`, and `--integration` all keep working.
- Plan committed at `plans/split-ci-unit-integration.md`.

Closes #104. Follow-up for Docker layer caching: #129.

## Branch-protection note (post-merge)

The required check name changes from `test` to `tests` (the rollup job). `main` has no active branch protection today (tracked by #93), so there is no gap to mitigate in this PR — #93 can adopt the new name directly when it lands.

## Acceptance criteria

- [x] Unit and integration tests run in parallel on CI — confirmed below.
- [x] Total PR wall-clock strictly lower than a serialized run — see evidence below.
- [x] A single check (the `tests` rollup) gates merges — delivered; branch-protection setup happens in #93.
- [x] `scripts/test.sh` still supports `--unit`, `--integration`, `--all` locally — unchanged.

**Wall-clock evidence** (run `24665280330`, head `c52a083`):

| Job           | Start         | End           | Duration |
|---------------|---------------|---------------|----------|
| `unit`        | 12:00:07Z     | 12:01:13Z     | 1m 06s   |
| `integration` | 12:00:07Z     | 12:11:19Z     | 11m 12s  |
| `tests`       | 12:11:23Z     | 12:11:27Z     | 0m 04s   |
| **PR total**  | 12:00:07Z     | 12:11:27Z     | **11m 20s** |

Baseline (most recent comparable run on `main`, `20b73a9`, serialized `test` job): **11m 54s** (11:58:00Z → 12:09:54Z).

Net win: ~34s on this run. Integration dominates so parallelism only reclaims the unit-test cost (~1m). The bigger-ticket win is wiring buildx layer cache into the integration image build — tracked in #129.

**Parallelism confirmed:** `unit` and `integration` both started at `12:00:07Z` (identical timestamps), ruling out serialization.

## Test plan

- [x] Confirm `unit` and `integration` jobs start concurrently (not serialized) on the Actions page.
- [x] Confirm `tests` rollup waits for both and reports success when both pass.
- [x] Mentally verify the rollup's `needs.*.result != 'success'` check fails the rollup if either upstream fails.
- [x] Confirm all other workflows (`check.yml`, `security.yml`, `verify-*.yml`) are unaffected (all green on this PR).
- [x] Record wall-clock times in this description.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,8 +36,10 @@ jobs:
     needs: [unit, integration]
     if: always()
     steps:
-      - name: Fail if any upstream job did not succeed
-        if: needs.unit.result != 'success' || needs.integration.result != 'success'
+      - name: Aggregate unit and integration results
         run: |
-          echo "unit=${{ needs.unit.result }} integration=${{ needs.integration.result }}"
-          exit 1
+          if [[ "${{ needs.unit.result }}" != "success" \
+             || "${{ needs.integration.result }}" != "success" ]]; then
+            echo "unit=${{ needs.unit.result }} integration=${{ needs.integration.result }}"
+            exit 1
+          fi
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  test:
+  unit:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -16,5 +16,28 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Run tests
-        run: task test -- --all
+      - name: Run unit tests
+        run: task test -- --unit
+
+  integration:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: ./.github/actions/setup-toolchain
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run integration tests
+        run: task test -- --integration
+
+  tests:
+    runs-on: ubuntu-latest
+    needs: [unit, integration]
+    if: always()
+    steps:
+      - name: Fail if any upstream job did not succeed
+        if: needs.unit.result != 'success' || needs.integration.result != 'success'
+        run: |
+          echo "unit=${{ needs.unit.result }} integration=${{ needs.integration.result }}"
+          exit 1
--- a/plans/split-ci-unit-integration.md
+++ b/plans/split-ci-unit-integration.md
@@ -1,0 +1,188 @@
+# Plan: Split CI unit and integration tests into parallel jobs
+
+Issue: [#104](https://github.com/aidanns/agent-auth/issues/104).
+
+Source: GitHub issue #104 — CI latency optimisation. No
+instruction-file standard mandates this.
+
+## Goal
+
+Reduce PR wall-clock time by running the unit suite and the Docker-backed
+integration suite in parallel on CI. Total time should drop from
+`unit + integration` to roughly `max(unit, integration)`, and unit-test
+failures should surface before the integration image finishes building.
+
+Today `.github/workflows/test.yml` has a single `test` job that invokes
+`task test -- --all`, which runs `pytest tests/ --ignore=tests/integration`
+and then `pytest tests/integration/` sequentially. The integration layer
+builds `docker/Dockerfile.test` once per session (in
+`tests/integration/_support.py:build_test_image`) and spins up a fresh
+Compose project per test, so its cost dominates.
+
+## Non-goals
+
+- **Matrix OS expansion** — #69 tracks adding a macOS runner for the
+  AppleScript path. This plan must not block that work, but does not
+  implement it. The new structure should make adding a `macos` job a
+  trivial extension of the `unit` / `integration` job definitions.
+- **Changes to `scripts/test.sh` or `Taskfile.yml`** — the split is a
+  workflow-level change only. Local `task test -- --all`, `--unit`,
+  and `--integration` must keep working unchanged.
+- **Branch-protection configuration in this repo** — see #93. This
+  plan produces a rollup check that *can* be required; actually
+  updating branch protection is a separate admin-level step.
+- **Docker layer caching across runs** — the integration harness
+  currently builds the test image with a direct `docker build` call
+  that does not accept `--cache-from` flags. Wiring GHA/buildx cache
+  requires a harness change that is out of scope for this PR. Filed as
+  a follow-up issue; see step 4 below.
+
+## Deliverables
+
+1. **`.github/workflows/test.yml`** restructured to two parallel jobs
+   plus a rollup:
+   - `unit` — runs `task test -- --unit`. No Docker needed.
+   - `integration` — runs `task test -- --integration` on a Docker-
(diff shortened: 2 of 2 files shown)


aidanns:
Claude Reviewer: Branch-protection bootstrapping gap — the PR description frames a risk window between merge and admin-update, but `main` has no active branch protection today (that's exactly what #93 tracks, still open). So there's no required check to lose when `test` is renamed to `tests`. I'd soften the "required" language in the description — the rename is only consequential *after* #93 lands, and #93 can choose the new name from the start. If you do want forward-compat for an imminent #93 rollout, keeping a no-op job literally named `test` as an alias for one cycle (a trivial `runs-on: ubuntu-latest` + `needs: [tests]` job) would eliminate any re-config step. Not blocking; just a framing fix.

aidanns:
Claude Reviewer: Rollup `if: always()` + step-level `if:` — walked the combinations and the outcome reported to branch protection is correct in every case:

- success+success → job runs, the single step is skipped (step `if` false), **job result = success**. GitHub treats a job whose only step is skipped as a successful job, so branch protection sees green. Confirmed intended.
- success+failure / failure+success / failure+failure → step `if` true, `exit 1` → job fails. Correct.
- success+cancelled / skipped+anything → step `if` true (`!= 'success'`), `exit 1` → job fails. Correct (conservative).
- both cancelled (e.g. superseding push with concurrency) → rollup itself is typically cancelled before it runs; result = cancelled, which branch protection treats as not-success. Correct.

One small readability nit (non-blocking): the step name "Fail if any upstream job did not succeed" is accurate, but in the happy path Actions will show a *skipped* step named "Fail if…" which reads oddly in logs. Consider inverting to always-run with an inline conditional, e.g.:

```yaml
- name: Aggregate unit and integration results
  run: |
    if [[ "${{ needs.unit.result }}" != "success" || "${{ needs.integration.result }}" != "success" ]]; then
      echo "unit=${{ needs.unit.result }} integration=${{ needs.integration.result }}"
      exit 1
    fi
```

That's also the exact snippet in `plans/split-ci-unit-integration.md` — see my other comment re: plan/impl drift.

aidanns:
Claude Reviewer: Release-automation interaction — audited `.github/workflows/` on both `main` and this PR branch. No workflow uses `workflow_run` against `test.yml`, no release/publish/auto-merge workflow exists in `main`, no `concurrency:` group is shared with `test.yml`, and no other workflow references the `test` status-check name. Nothing external depends on the old job name. (Note: a separate unmerged worktree `spdx-sbom-reuse-release` contains `release-please.yml` and `release-publish.yml`, but they're tag/main-push triggered and don't consume `test.yml` status — so even when those land, no collision.) No action needed here.

aidanns:
Claude Reviewer: Plan vs implementation drift in `plans/split-ci-unit-integration.md`.

The plan (lines ~99-112) shows the rollup as a single always-run step with inline bash:

```yaml
- name: Aggregate unit and integration results
  run: |
    if [[ "${{ needs.unit.result }}" != "success" ...
```

The implementation shipped instead uses a step-level `if:` with a short `exit 1` body:

```yaml
- name: Fail if any upstream job did not succeed
  if: needs.unit.result != 'success' || needs.integration.result != 'success'
  run: |
    echo ...
    exit 1
```

Same behaviour, but the plan committed to the repo no longer matches the shipped workflow. Either update the plan snippet to match the final code, or switch the workflow to the planned shape (I'd slightly prefer the latter — it avoids the skipped-step-in-happy-path log noise). Minor, but per `.claude/instructions/plan-template.md` plans are meant to reflect what was built.

aidanns:
Claude: Done! (re [comment 4280488816](https://github.com/aidanns/agent-auth/pull/133#issuecomment-4280488816) — branch-protection framing) Softened the PR description to reflect that main has no active branch protection today — #93 will pick up the new name directly. No alias job needed.

aidanns:
Claude: Done! (re [comment 4280490367](https://github.com/aidanns/agent-auth/pull/133#issuecomment-4280490367) — rollup readability) Switched the rollup back to the always-run inline-bash form to avoid the skipped-step log UX. Also resolves the plan/impl drift from [comment 4280492507](https://github.com/aidanns/agent-auth/pull/133#issuecomment-4280492507) since the shipped workflow now matches the committed plan verbatim. Pushed as 511146f.

aidanns:
Claude: Ack (re [comment 4280491306](https://github.com/aidanns/agent-auth/pull/133#issuecomment-4280491306) — release automation audit). No action needed — nothing external references the old `test` check name.

aidanns:
Claude: Done! (re [comment 4280492507](https://github.com/aidanns/agent-auth/pull/133#issuecomment-4280492507) — plan/impl drift) Resolved by converging the workflow to the plan's shape. See [the reply to comment 4280490367](https://github.com/aidanns/agent-auth/pull/133#issuecomment-4280504406).

Issue #137, opened 2026-04-20
Token management routes are not /v1/ namespaced (verify-standards failing on main)


aidanns:
## Summary

`main` is currently red on **Verify Standards** because the token-management routes added by #97 (`/agent-auth/token/create`, `/modify`, `/revoke`, `/rotate`, `/list`) are not under the `/v1/` namespace that `.claude/instructions/service-design.md` mandates and `scripts/verify-standards.sh` enforces (check added by #126).

The two PRs merged in this order:

1. #97 (11:57 UTC) — added the non-versioned routes; CI was green because the check did not yet exist.
2. #126 (12:11 UTC) — added the `/v1/` enforcement check; CI went red on main at this point.

## Evidence

```
$ bash scripts/verify-standards.sh
...
verify-standards: unversioned HTTP routes found in server files:
src/agent_auth/server.py:65:        elif self.path == "/agent-auth/token/create":
src/agent_auth/server.py:67:        elif self.path == "/agent-auth/token/modify":
src/agent_auth/server.py:69:        elif self.path == "/agent-auth/token/revoke":
src/agent_auth/server.py:71:        elif self.path == "/agent-auth/token/rotate":
src/agent_auth/server.py:81:        elif self.path == "/agent-auth/token/list":
  All endpoints (except /health) must use the /v1/ namespace.
```

Latest **Verify Standards** run on `main`: https://github.com/aidanns/agent-auth/actions/runs/24665757401 (failure).

## Fix

Move the five routes under `/v1/` in `src/agent_auth/server.py` and update `src/agent_auth/cli.py` / `src/things_cli/client.py` / tests / docs accordingly. This is the same move #126 made for `validate`, `refresh`, `reissue`, `status`.

## Why this matters for other PRs

Any PR branched after #97 merged (including #134 — SSDF adoption) will see Verify Standards fail on merge with `origin/main` even though the PR itself does not introduce the unversioned routes.

Surfaced by #134.

aidanns:
## Scope update: also breaks the Test workflow

The Test suite on `main` has been red since #126 merged, for the same root cause. `tests/test_server.py` lines 535 / 631 / 638 POST to `{base}/agent-auth/validate` (the pre-#126 URL), added by #97. After #126 moved `validate` to `/v1/validate`, those POSTs get `404` instead of the expected `401` for revoked tokens.

Failing tests:
- `test_token_revoke_marks_family_revoked` — asserts `validate_status == 401`, gets `404`
[...]

Issue #69, opened 2026-04-18
CI: add macOS runner to exercise the real AppleScript client CLI end-to-end


aidanns:
## Context

PR #66 split the Things client into sibling CLIs. The test suite now covers:

- The subprocess contract end-to-end against the fake CLI (\`tests/test_things_client_cli_contract.py\`).
- The bridge's argv builder + envelope parsing in isolation (\`tests/test_things_subprocess_client.py\`).
- The shipped AppleScript client against recorded osascript output on macOS developer machines (\`tests/test_things_client_applescript_things.py\`, skipped on Linux).

No CI job actually runs \`things-bridge\` against \`things-client-cli-applescript\` against real osascript. That path is only exercised on whoever's laptop happens to be macOS when the PR is reviewed.

See the PR review comment: https://github.com/aidanns/agent-auth/pull/66#issuecomment-4273417220 (concern 5).

## Proposal

Add a \`macos\` job to \`.github/workflows/test.yml\` using a macOS runner. On that job, run the AppleScript-backed subset of the test suite including any newly-added integration test that drives the bridge end-to-end with \`things_client_command: [things-client-cli-applescript]\`.

Open questions to resolve during implementation:

- Whether to run on every PR (cost) or only on PRs that touch \`src/things_client_applescript/\`, \`src/things_bridge/\`, or \`src/things_models/\` via \`paths:\` filters.
- Whether the integration test needs a real Things 3 instance (requires fixture-populated Things), or whether testing against a mocked osascript shim running under real macOS is enough.

## Acceptance

- [ ] \`.github/workflows/test.yml\` has a macOS job that runs the Darwin-gated tests.
- [ ] Either the job runs a full \`things-bridge\` + \`things-client-cli-applescript\` end-to-end integration test, or this issue documents why it doesn't and tracks that as a separate follow-up.

## Related

- PR #66 (originating change)

PR #140, opened 2026-04-20
ci: split integration tests into per-service jobs

### LINKED ISSUE #138: ci: split integration tests into per-service workflow jobs
## Context

Follow-up from #104 (split unit + integration into parallel jobs).

The `integration` job in `.github/workflows/test.yml` currently runs the entire `tests/integration/` tree serially under a single pytest invocation. That tree already contains per-service subdirectories:

- `tests/integration/` (top level) — agent-auth server tests (`test_health.py`, `test_refresh.py`, `test_reissue.py`, `test_status.py`, `test_token_management.py`, `test_validate.py`)
- `tests/integration/things_bridge/` — things-bridge tests
- `tests/integration/things_cli/` — things-cli tests
- `tests/integration/things_client_applescript/` — things-client-cli-applescript contract tests

Each service's tests are independent (they build their own containers / drive their own surfaces) and could run in parallel. Today they share a single runner, so they pay the image-build and compose-spin-up cost sequentially.

## Proposal

Split the single `integration` job into per-service jobs under `.github/workflows/test.yml`:

- `integration-agent-auth` — `tests/integration/` (excluding subpackages)
- `integration-things-bridge` — `tests/integration/things_bridge/`
- `integration-things-cli` — `tests/integration/things_cli/`
- `integration-things-client-applescript` — `tests/integration/things_client_applescript/`

Fold each into the existing `tests` rollup's `needs:` list so branch protection still gates on a single check.

Likely shape: add a `--service` flag (or per-service mode flags) to `scripts/test.sh`, or use pytest's `-k` / path-targeting. Keep local `task test -- --integration` runnable as an umbrella that executes all service slices.

## Considerations

- **Matrix vs explicit jobs** — a `strategy.matrix.service: [...]` collapses the duplication but complicates per-service customisation (e.g. the AppleScript path eventually needing `macos-latest` from #69). Start with explicit jobs; migrate to matrix only if they stay uniform.
- **Docker layer cache** — pairs well with #129. Each service may share the same base image, so a buildx cache scoped per-service (or a single shared cache) should be considered during that work. Order this issue *after* #129 so the cache primes before we fan out further builds.
- **Runner concurrency** — 4 parallel integration jobs + 1 unit job = 5 concurrent runners. Verify this stays under the GHA free-tier concurrency limit for the repo.
- **Test count preservation** — the sum of per-service test counts must equal the current `integration` job's test count. Add a verification step or leave a comment on the first PR listing the counts.
- **`scripts/test.sh --all` locally** — must still exercise every service. No local-workflow divergence.

## Acceptance

- [ ] Integration tests for each service run in parallel on CI.
- [ ] Total PR wall-clock strictly lower than the current single-integration-job run on a representative PR.
- [ ] `tests` rollup gates on every per-service job.
- [ ] `task test -- --integration` locally still runs the full suite (no service is silently skipped).
- [ ] No test is dropped or double-counted in the split.

aidanns:
## Summary

- Fans out the single `integration` job in `.github/workflows/test.yml` into four per-service jobs: `integration-agent-auth`, `integration-things-bridge`, `integration-things-cli`, `integration-things-client-applescript`.
- All four jobs feed into the existing `tests` rollup so branch protection keeps gating on one check.
- `scripts/test.sh --integration` gains an optional positional service argument; `task test -- --integration agent-auth` runs only that slice. Unknown service names fail fast with the valid keys. Default `--integration` (no service) and `--all` still run the full integration suite.
- Moves the agent-auth integration tests from the top of `tests/integration/` into `tests/integration/agent_auth/` so every service has a uniform subpackage layout (no `--ignore` bookkeeping needed).
- Hardens `scripts/verify-integration-isolation.sh` to fail outright when a service subdir lacks a `conftest.py` (previously silently skipped).

Closes #138. Still-open follow-ups: #129 (buildx layer cache), #69 (macOS runner for AppleScript slice).

## Wall-clock evidence (PR #140 Test run 24666453766, head `372998a`)

| Job | Duration | Tests |
|---|---|---|
| `unit` | 1m 29s | 318 passed, 2 failed (pre-existing #137) |
| `integration-agent-auth` | 6m 51s | 15 passed, 15 failed (pre-existing #137) |
| `integration-things-bridge` | 3m 50s | 15 passed |
| `integration-things-cli` | 1m 49s | 5 passed |
| `integration-things-client-applescript` | 44s | 13 passed |
| `tests` rollup | 2s | aggregates upstreams |
| **PR total wall-clock** | **6m 58s** | |

All five feed-in jobs started within 2 seconds of each other (12:28:02–12:28:04 UTC) — parallelism confirmed.

**Baseline** (last green Test run on `main` pre-regression, PR #133 run 24665428290, head `c52a083`): serialized `integration` 11m 01s + parallel `unit` 1m 09s → **~11m 04s** total wall-clock.

**Delta: ~-4m 06s (≈37% faster).** `integration-agent-auth` is the new long pole; the three things-* slices now run concurrently instead of serially.

## CI-minute tradeoff (before #129 lands)

Four integration jobs each build `docker/Dockerfile.test` from scratch, so total CI minutes go *up* (~4×) even as wall-clock drops. Interim state until #129 wires buildx + GHA layer cache into `tests/integration/_support.py:build_test_image`.

## Pre-existing regression (blocks merge)

This PR's red CI is **not** caused by the CI split. `tests/test_server.py::test_token_revoke_marks_family_revoked`, `::test_token_rotate_revokes_old_and_creates_new_family`, and all 15 `tests/integration/agent_auth/test_token_management.py` tests fail with `assert 404 == ...` because the token-management HTTP routes were added without the `/v1/` prefix that `.claude/instructions/service-design.md` mandates. The test clients call `{base}/agent-auth/v1/token/create`, the server serves `/agent-auth/token/create`.

Tracked by **#137**. Merge this PR only after #137 lands on `main` — then rebase this branch and CI should go green across all five jobs.

## Test plan

- [x] All four `integration-*` jobs start concurrently on CI (confirmed, 2s spread).
- [x] Each per-service job's test count matches the local collect-only table (30+15+5+13=63; run 24666453766 confirms 15, 5, 13 passes on the three green slices and 15+15 collect on agent-auth).
- [x] `tests` rollup waits for all five upstreams and fails when any fail.
- [x] `task test -- --integration` (no service) still runs all 63 tests locally.
- [x] `task test -- --integration agent-auth` runs 30 tests.
- [x] `task test -- --integration bogus` exits 2 with the valid-services list.
- [x] `scripts/verify-integration-isolation.sh` still passes and fails the right way for a no-conftest subdir (verified with a scratch fixture).
- [x] `task check` passes locally.
- [x] After #137 merges, rebase and re-verify all jobs green.


aidanns:
Claude Reviewer: `scripts/verify-integration-isolation.sh` rule 3 is silently bypassed by the new `tests/integration/agent_auth/` subdir. The loop at line 75 iterates `tests/integration/*/conftest.py` with `[[ -f ... ]] || continue` — a service subdir with no conftest is skipped entirely, not flagged. Agent-auth's fixtures live in the parent `tests/integration/conftest.py`, so `tests/integration/agent_auth/` has no conftest and the "each per-service subdir must pin its container topology" rule never runs against it. The script still prints `OK` locally. Two ways to fix: (a) tighten the loop to iterate subdirectories and fail when a subdir lacks a conftest that references a tracked compose/run artefact, or (b) explicitly exempt `agent_auth/` because its fixtures are in the parent conftest (which does reference `docker-compose.yaml` via `render_compose_file`). Option (a) is more robust against future per-service subdirs that forget to pin topology.

aidanns:
Claude Reviewer: Minor — `plans/token-management-http-api.md:62` still references `tests/integration/test_token_management.py` at the old path. The new plan explicitly calls this out as "historical, leave untouched" (plans/split-integration-per-service.md:53-54), so this is intentional. Flagging for visibility only; no action required.

Verified clean on the other angles:
- Function-test allocation (`design/functional_decomposition.yaml` + `systems-engineering function verify`) keys on function names, not pytest node IDs, so the file move doesn't break coverage mapping. `bash scripts/verify-function-tests.sh` runs with the same 52/55 coverage as main.
- `scripts/verify-standards.sh` asserts nothing about `test.yml` job names or the rollup check name, so the split doesn't trip standards gating. (It does fail locally on an unversioned-routes check in `src/agent_auth/server.py` — that's a pre-existing failure on `main` as well, unrelated to this PR.)
- Per-slice pytest collection matches the PR table exactly: 30 + 15 + 5 + 13 = 63, and the full `tests/integration/` collection is also 63. No drift, no doubles.
- `scripts/test.sh` arg parsing correctly distinguishes pytest flags (starting with `-`) from a service-name positional, so `task test -- --integration -k foo` still works as expected.

aidanns:
Claude: Done! (re [comment 4280722861](https://github.com/aidanns/agent-auth/pull/140#issuecomment-4280722861) — isolation-script rule 3 bypass) Fixed in 7797178. Script now iterates subdirectories (skipping dunder-named ones like `__pycache__`) and fails outright when a service subdir has no `conftest.py` — so a newly-added service lacking topology pinning can't slip through silently. Also added `tests/integration/agent_auth/conftest.py` that documents where the shared fixtures live and references `docker-compose.yaml` so the existing topology-pin regex fires. Verified the new failure path against a scratch subdir with no conftest.

aidanns:
Claude: Ack (re [comment 4280724663](https://github.com/aidanns/agent-auth/pull/140#issuecomment-4280724663) — `plans/token-management-http-api.md` stale path). Confirmed intentional — that file is a historical point-in-time plan from a previous PR; rewriting shipped plans on unrelated changes just churns the record. No change.

aidanns:
Claude: CI results summary — the split achieves its goal (wall-clock 6m58s vs 11m4s baseline, ~37% faster, 5 jobs parallel within 2s). Red CI on this branch is **pre-existing regression #137** (token-management routes missing /v1/ prefix), not caused by the split: unit tests 318/320 pass, the three non-agent-auth integration slices are fully green, and agent-auth's 15 failures are all in test_token_management.py calling `/agent-auth/v1/token/*` paths the server doesn't serve. Blocking merge on #137 landing first.

aidanns:
Claude: Merged `aidanns/token-routes-v1` (f9aa42f — fix for #137) into this branch so CI can run green end-to-end. Locally: all 320 unit tests pass (previously 318/320), and `task check` is clean. Once #137 lands on main, the squash-merge of this PR will only include this PR's actual changes (the CI split + isolation-script hardening).

Issue #149, opened 2026-04-20
Cache mypy/pyright state in CI and parallelize locally


aidanns:
Follow-up from the #48 code review:

1. **CI cache:** `.github/workflows/typecheck.yml` runs mypy + pyright from scratch on every PR. Adding an `actions/cache` step keyed on `hashFiles('**/*.py', 'pyproject.toml')` for `.mypy_cache/` would give meaningful speedup on touch-one-file PRs. Pyright has no on-disk cache (beyond what `setup-uv@v5 enable-cache: true` already caches for the venv).

2. **Local parallelism:** `scripts/typecheck.sh` runs mypy then pyright sequentially. Running them in parallel via `& wait` — with stderr captured per-tool and dumped in order after `wait` to keep output readable — would roughly halve local-dev typecheck latency.

**Source:** Deferred from #48 code review. Not blocking because CI wall time is acceptable (~1-2 min per run) and local dev latency is bearable; file when one of the two becomes annoying.

Issue #148, opened 2026-04-20
Ratchet mypy/pyright strict on tests/ and tests_support/


aidanns:
Foundation PR #48 landed mypy + pyright with `strict = true` globally plus a blanket per-package relaxation for `tests.*` and `tests_support.*` (655 errors across 41 test files at the time). This issue tracks tightening the test tree.

**Scope:** all files under `tests/`, `tests/integration/`, and `tests_support/` (41+ files).

**Approach:** tighten incrementally. Pytest test functions rarely benefit from return-type annotations (`-> None`), and parametrize/fixture decorators produce low-signal strict errors, so the recommended order is:

1. Narrow the `[[tool.mypy.overrides]]` relaxation from blanket `ignore_errors = true` to `disallow_untyped_defs = false` (keeps real type-mismatch reports, suppresses the noisy ones). Same for pyright via `diagnosticSeverityOverrides` per-file.
2. Pick off individual test modules with few real errors and fully strict them.
3. Fold `tests_support/` (the notification plugin fakes) into the production ratchet since those are reused across multiple test files.

**Error counts per file:** see `uv run mypy --strict tests/` at the time of #48 — 655 errors, largely `no-untyped-def` from pytest function signatures.

**How to close the issue:** `tests.*` and `tests_support.*` removed from the `[[tool.mypy.overrides]]` block and from `pyrightconfig.json`'s `ignore`; `task typecheck` green.

**Source:** Follow-up from #48 / PR #TBD.

Issue #147, opened 2026-04-20
Ratchet mypy/pyright strict on things_cli/ + things_client_common/ + things_models/ modules


aidanns:
Foundation PR #48 landed mypy + pyright with `strict = true` globally plus per-module relaxations for modules that failed at the time. This issue tracks tightening each things_cli / things_client_common / things_models module back to strict.

**Modules under relaxation** (both mypy and pyright):
- [ ] `src/things_cli/cli.py` (11 errors)
- [ ] `src/things_cli/client.py` (9 errors)
- [ ] `src/things_cli/credentials.py` (6 errors)
- [ ] `src/things_cli/output.py` (6 errors)
- [ ] `src/things_client_common/cli.py` (1 error)
- [ ] `src/things_models/models.py` (3 errors)

Error counts are from `uv run mypy --strict src/` at the time of #48.

**How to close one module:**
1. Remove the module from the `[[tool.mypy.overrides]]` entry in `pyproject.toml` and from the `ignore` list in `pyrightconfig.json`.
2. Fix the strict-mode errors in the module (and any callers that inherit untyped-call errors).
3. Confirm `task typecheck` stays green.
4. Tick the box above and commit.

**When to close the issue:** all modules ticked.

**Source:** Follow-up from #48 / PR #TBD.

Issue #146, opened 2026-04-20
Ratchet mypy/pyright strict on things_bridge/* modules


aidanns:
Foundation PR #48 landed mypy + pyright with `strict = true` globally plus per-module relaxations for modules that failed at the time. This issue tracks tightening each things_bridge module back to strict.

**Modules under relaxation** (both mypy and pyright):
- [ ] `src/things_bridge/cli.py` (2 errors)
- [ ] `src/things_bridge/config.py` (2 errors)
- [ ] `src/things_bridge/server.py` (5 errors)
- [ ] `src/things_bridge/things_client.py` (4 errors)

Error counts are from `uv run mypy --strict src/` at the time of #48.

**How to close one module:**
1. Remove the module from the `[[tool.mypy.overrides]]` entry in `pyproject.toml` and from the `ignore` list in `pyrightconfig.json`.
2. Fix the strict-mode errors in the module (and any callers that inherit untyped-call errors).
3. Confirm `task typecheck` stays green.
4. Tick the box above and commit.

**When to close the issue:** all modules ticked.

**Source:** Follow-up from #48 / PR #TBD.

PR #150, opened 2026-04-20
ci: wire mypy + pyright typecheck gate

### LINKED ISSUE #48: Add mypy and pyright type-checking CI checks
Source: `.claude/instructions/python.md` — mypy and pyright.

No type checker runs in CI. Standard: run both `mypy` and `pyright` (they catch different things).

**Acceptance:**
- `mypy` config in `pyproject.toml` with `strict = true` (or explicit subset; document deviations).
- `pyright` config committed (e.g. `pyrightconfig.json`) with equivalent strictness.
- Both run in CI on `src/` and `tests/`.
- Existing annotations audited and upgraded so both checkers pass.
- Ties into #22, #23, #24 (NewType / semantic types / unit names).

- `scripts/verify-standards.sh` asserts `pyproject.toml` has `[tool.mypy]` with strict mode, `pyrightconfig.json` exists, and at least one CI workflow runs both `mypy` and `pyright`.

aidanns:
## Summary

- Adds mypy and pyright as dev dependencies and wires both into CI under `task typecheck` (new `.github/workflows/typecheck.yml`).
- `pyproject.toml` declares `[tool.mypy]` with `strict = true` as the default; `pyrightconfig.json` uses `typeCheckingMode: "strict"`. Modules that failed at foundation time carry per-module `ignore_errors = true` (mypy) / `ignore` (pyright) relaxations — ratchet approach.
- Updates `scripts/verify-standards.sh` with the deterministic regression check from #48 (strict default, pyright config present, both checkers gated in CI) and a drift check that the mypy override module list and pyright ignore paths resolve to the same file set.
- `src/things_bridge/authz.py` gets minor type annotations (added `data: dict[str, Any]` and `str()` coerce) so it stays out of the ratchet set — walking through edge cases confirms behavior is preserved.

Closes #48.

## Ratchet follow-ups

Grouped by submodule so each can close independently:

- [#145](https://github.com/aidanns/agent-auth/issues/145) — `agent_auth/*` modules (8 files)
- [#146](https://github.com/aidanns/agent-auth/issues/146) — `things_bridge/*` modules (4 files)
- [#147](https://github.com/aidanns/agent-auth/issues/147) — `things_cli/*`, `things_client_common/*`, `things_models/*` (6 files)
- [#148](https://github.com/aidanns/agent-auth/issues/148) — `tests/*` and `tests_support/*` (41+ files)
- [#149](https://github.com/aidanns/agent-auth/issues/149) — CI caching for `.mypy_cache/` + local-dev parallelism

Closing a ratchet issue means removing the module from both `[[tool.mypy.overrides]]` and `pyrightconfig.json`'s `ignore`, fixing the strict-mode errors, and confirming `task typecheck` stays green.

## Test plan

- [ ] CI passes on this branch (new `Typecheck` workflow is green; existing workflows unchanged).
- [x] `task typecheck` passes locally.
- [x] `task verify-standards` passes locally with the new checks.
- [x] `task verify-standards` fails correctly when `strict = true` is removed from `[tool.mypy]`, when `pyrightconfig.json` is missing, or when the mypy/pyright ratchet lists go out of sync (all smoke-tested locally before PR; the drift check actually caught a real `tests_support` path bug during authoring).
- [x] `task lint`, `task format --check`, and `task test --fast` all pass on the branch.
- [x] Pre-commit hook chain fires on commit (includes `treefmt`, `ripsecrets`, `test-fast`, `ruff-check`, `keep-sorted`).


aidanns:
Claude Reviewer: The four new files in this PR will break the `REUSE` CI workflow (added on main in #132) because they don't carry SPDX headers:

- `scripts/typecheck.sh` — every other `scripts/*.sh` on main has `# SPDX-FileCopyrightText` + `# SPDX-License-Identifier` after the shebang.
- `.github/workflows/typecheck.yml` — other workflow YAMLs carry the header as top-of-file comments.
- `plans/mypy-pyright-ci.md` — other plan files wrap it in an HTML comment block.
- `pyrightconfig.json` — JSON can't carry inline comments; needs a `[[annotations]]` stanza in `REUSE.toml` (the existing stanza covers `.release-please-config.json` but not this file).

Once main is merged into the branch, `task reuse-lint` / the REUSE workflow will flag all four until headers (or a REUSE.toml override for pyrightconfig.json) are added.

aidanns:
Claude Reviewer: Minor: `scripts/verify-standards.sh`'s strict-mode check is `grep -qE "^strict[[:space:]]*=[[:space:]]*true" <<<"${pyproject_stripped}"` — it's not scoped to the `[tool.mypy]` section. If another tool in `pyproject.toml` ever grew a top-level `strict = true` line, the check would pass even if `[tool.mypy]` had `strict = false`. Not exploitable today (only one such line exists), but since the whole point of the gate is to catch a regression on the mypy block specifically, scoping it would make the deterministic check from #48 harder to defeat accidentally. A small Python helper (parse TOML, assert `pyproject["tool"]["mypy"]["strict"] is True`) would mirror the ratchet-drift helper a few lines below and dodge the regex altogether.

aidanns:
Claude Reviewer: `scripts/verify-standards.sh`'s typecheck CI gate (`grep -qE "run:[[:space:]]*[^\\n]*\\bmypy\\b"`) has a small hole: a `run:` line like `run: echo mypy` would satisfy it even though nothing is actually being type-checked. Same pattern as the treefmt/ripsecrets checks just above, so the hole is consistent — but since `task typecheck` is the blessed path and it already satisfies both, you could simplify the whole block to "at least one `run: task typecheck` exists" and drop the bare-mypy/bare-pyright fallbacks. That loses the ability to wire the tools directly without going through the task, which this repo doesn't do anywhere else. Take it or leave it.

aidanns:
Claude Reviewer: The ratchet-drift helper's `module_to_path` resolves modules to filesystem paths by existence-checking in order. That's fine for catching drift TODAY, but it means a developer who renames `agent_auth.cli` to `agent_auth.entrypoint` in `pyproject.toml` (and does create a file at that location) without updating `pyrightconfig.json` gets flagged. Good. However, if they rename to a NON-existent module (typo in `pyproject.toml`), the helper silently falls through to `return f"src/{path}"` — so the check only flags the pyright-side mismatch, not the underlying "module you named doesn't exist". Low-severity (mypy would itself complain on next run), but since the helper already walks the candidates list, you could also emit a warning when none resolve.

aidanns:
Claude Reviewer: `sys` is imported inside the ratchet-drift helper at `scripts/verify-standards.sh:509` but only used via `sys.exit(1)` on the drift path — harmless, but `ruff` would normally flag an unused import if the helper were a standalone module. Since it's embedded in a heredoc, ruff never sees it. Not actionable; calling it out in case you want to drop the import and rely on a non-zero raise (e.g. `raise SystemExit(1)`) for symmetry with how the other embedded helpers handle exit.

aidanns:
Claude Reviewer: Quick sanity-run of the ratchet-drift helper against the committed `pyproject.toml` + `pyrightconfig.json`: both sets resolve to 20 paths, `missing_from_pyright` and `missing_from_mypy` are empty. Also confirmed the helper catches the tests_support path mismatch — `tests_support.*` in mypy maps to `src/tests_support` on disk via the second candidate, which matches what pyright's `ignore` says. The drift check does its job.

On behavior-preservation in `src/things_bridge/authz.py`: walked through the four code paths (200/valid, 401/token_expired, 401/other, 403, fallthrough) for the case `data == {"error": null}` — old code bound `error_code = None`, new code binds `error_code = ""`. Both paths hit the same `or "invalid_token"`/`or "scope_denied"`/`or "<no body>"` fallbacks, so the emitted exception type and message are equivalent. Behavior is preserved.

Aside from the SPDX-header gap, this looks good.

aidanns:
Claude: Addressed in e708bab (after merging `origin/main` to pick up #132).

**Re: SPDX headers on 4 new files.** Done — headers added to `scripts/typecheck.sh`, `.github/workflows/typecheck.yml`, and `plans/mypy-pyright-ci.md` inline; `pyrightconfig.json` added to `REUSE.toml`'s JSON annotation block (JSON can't carry inline comments). `task reuse-lint` is now green on the branch.

**Re: nit #2 (strict gate not block-scoped).** Acknowledged. The current `^strict\s*=\s*true` pattern is consistent with the existing shellcheck/shfmt/ripsecrets checks in `scripts/verify-standards.sh`, which all match bare tokens rather than scoping to specific TOML tables. Tightening would require a TOML-aware Python helper (like the ratchet-drift check does) — worth doing, but as a follow-up that covers all the similar patterns together rather than just this one.

**Re: nit #3 (`run: echo mypy` would satisfy the CI gate).** Same pattern as the treefmt/ripsecrets checks from #42 — consistent trade-off. Trivially bypassable only if someone actively tries to subvert the gate; no signal-over-noise value in tightening.

**Re: nit #4 (`module_to_path` silent fallthrough).** Fair — a typo'd module currently flags via the drift check as 'missing_from_pyright: src/typoed/path' rather than as a typo. I'd rather land the foundation and fix this in one of the ratchet follow-ups when we're actually editing the override list.

**Re: nit #5 (`import sys`).** `sys.exit(1)` on the last line of the helper, so not unused.

PR #152, opened 2026-04-20
test: instrument integration test harness timing

aidanns:
## Summary
- Add ``phase_timer`` context manager and wrap ``compose.start``/``compose.stop``/``build_test_image``/``wait_until_server_ready`` so the integration harness emits ``phase=<name> elapsed_seconds=<n>`` INFO logs on the ``integration.timing`` logger.
- Let ``AGENT_AUTH_TEST_IMAGE_TAG`` override the session-scoped image build so CI can prebuild in its own step (new ``build-integration-test-image`` composite action) and the GHA step-timing UI shows the build cost split from the test wall-time.
- Enable ``--durations=0 --durations-min=0.1`` and ``log_cli`` when ``scripts/test.sh`` runs the integration suite so CI shows per-phase timings alongside per-test setup/call/teardown durations.

## Test plan
- [x] ``bash scripts/test.sh --unit`` passes locally.
- [x] ``bash scripts/test.sh --integration agent-auth`` on a Docker-capable host emits ``phase=compose_start`` / ``phase=compose_stop`` / ``phase=wait_until_server_ready`` INFO lines during each test's fixture phase. *(Not runnable on this devcontainer — no Docker daemon; CI coverage below substitutes.)*
- [x] The CI ``integration-*`` jobs on this PR show a ``Build integration test image`` step separate from the ``Run ... integration tests`` step, so the build-vs-test split is visible.
- [x] The unit test ``tests/test_integration_support.py`` passes and pins the ``phase=<name> elapsed_seconds=<n> key=value`` log format.

### DIFF
--- a/.github/actions/build-integration-test-image/action.yml
+++ b/.github/actions/build-integration-test-image/action.yml
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
+#
+# SPDX-License-Identifier: MIT
+
 name: Build integration test image
 description: |
   Build docker/Dockerfile.test under a stable tag so integration jobs
--- a/.github/actions/build-integration-test-image/action.yml
+++ b/.github/actions/build-integration-test-image/action.yml
@@ -1,0 +1,20 @@
+name: Build integration test image
+description: |
+  Build docker/Dockerfile.test under a stable tag so integration jobs
+  can reuse it via AGENT_AUTH_TEST_IMAGE_TAG. Kept as a step (not bundled
+  into the test step) so the GHA step-timing UI shows build cost split
+  out from test wall-time.
+
+inputs:
+  tag:
+    description: Image tag to apply to the built image
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: docker build
+      shell: bash
+      env:
+        IMAGE_TAG: ${{ inputs.tag }}
+      run: docker build -f docker/Dockerfile.test -t "$IMAGE_TAG" .
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,12 @@ on:
   pull_request:
     branches: [main]
 
+# Shared across every integration-* job: the composite build action and
+# the pytest session resolve the pre-built image through this tag, so
+# they must agree.
+env:
+  AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
+
 jobs:
   unit:
     runs-on: ubuntu-latest
@@ -21,8 +27,6 @@ jobs:
 
   integration-agent-auth:
     runs-on: ubuntu-latest
-    env:
-      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
@@ -39,8 +43,6 @@ jobs:
 
   integration-things-bridge:
     runs-on: ubuntu-latest
-    env:
-      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
@@ -57,8 +59,6 @@ jobs:
 
   integration-things-cli:
     runs-on: ubuntu-latest
-    env:
-      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
@@ -75,8 +75,6 @@ jobs:
 
   integration-things-client-applescript:
     runs-on: ubuntu-latest
-    env:
-      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,48 +21,72 @@ jobs:
 
   integration-agent-auth:
     runs-on: ubuntu-latest
+    env:
+      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
       - uses: ./.github/actions/setup-toolchain
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: ./.github/actions/build-integration-test-image
+        with:
+          tag: ${{ env.AGENT_AUTH_TEST_IMAGE_TAG }}
 
       - name: Run agent-auth integration tests
         run: task test -- --integration agent-auth
 
   integration-things-bridge:
     runs-on: ubuntu-latest
+    env:
+      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
       - uses: ./.github/actions/setup-toolchain
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: ./.github/actions/build-integration-test-image
+        with:
+          tag: ${{ env.AGENT_AUTH_TEST_IMAGE_TAG }}
 
       - name: Run things-bridge integration tests
         run: task test -- --integration things-bridge
 
   integration-things-cli:
     runs-on: ubuntu-latest
+    env:
+      AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci
     steps:
       - uses: actions/checkout@v6
 
       - uses: ./.github/actions/setup-toolchain
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: ./.github/actions/build-integration-test-image
+        with:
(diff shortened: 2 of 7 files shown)


aidanns:
Claude Reviewer: Nice instrumentation. Overall this is a clean, focused PR — structured `phase=... elapsed_seconds=...` logs on a dedicated logger, the pytest `--durations` switch, and the prebuild split are all the right moves to answer "where did the 5-minute run go". A few issues worth addressing before merge; individual line comments follow. The two REUSE-header regressions are almost certainly going to fail the `reuse` CI check.

Top concerns in priority order:

1. `scripts/test.sh` lost its SPDX header — likely an accidental edit, will fail `task reuse-lint`.
2. New files `.github/actions/build-integration-test-image/action.yml` and `tests/test_integration_support.py` are missing SPDX headers — every other YAML in `.github/` and every test file in `tests/` has them; also likely to fail the REUSE check.
3. `log_cli_level=INFO` is a global threshold — expect INFO-level spam from `urllib3`, `testcontainers`, `docker`, `asyncio`, etc. to land in the CI log alongside the `integration.timing` records. Worth narrowing (filter, or set the third-party loggers to WARNING at conftest import time) so `grep '^.*integration\.timing'` actually yields only timing rows.
4. The fixture short-circuit on `AGENT_AUTH_TEST_IMAGE_TAG` is load-bearing for CI correctness (wrong behaviour = either a double-build or running against a stale image) but is only covered by the docstring — no test pins "env var set => skip build + skip rmi". Worth one tiny unit test or at least an `assert` hook.

scripts/test.sh
```diff
@@ -66,6 +66,21 @@ FAST_TESTS=(
   # keep-sorted end
 )
 
+# Integration tests are slow enough that knowing where the time goes
```
aidanns, line 73: Claude Reviewer: Unrelated to this hunk but visible at the top of the file — this PR's diff shows the SPDX header (`# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith` / `# SPDX-License-Identifier: MIT`) was removed from lines 3-5 of `scripts/test.sh`. That looks accidental (every other script in `scripts/` still carries it, and the `REUSE` workflow will fail `task reuse-lint`). Please restore the block under the shebang.

.github/actions/build-integration-test-image/action.yml
```diff
@@ -0,0 +1,20 @@
+name: Build integration test image
```
aidanns, line 5: Claude Reviewer: Missing SPDX header — every other YAML under `.github/` (including the reference `.github/actions/setup-toolchain/action.yml`) opens with:

```yaml
# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
#
# SPDX-License-Identifier: MIT

name: ...
```

Without it the `REUSE` CI job (`task reuse-lint`) will fail on this new file.

tests/test_integration_support.py
```diff
@@ -0,0 +1,67 @@
+"""Unit tests for pure helpers in ``tests.integration._support``.
```
aidanns: Claude Reviewer: Missing SPDX header. Every other file under `tests/` (e.g. `tests/test_tokens.py`) opens with:

```python
# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
#
# SPDX-License-Identifier: MIT

"""..."""
```

Required for the `REUSE` CI check to pass.

scripts/test.sh
```diff
@@ -66,6 +66,21 @@ FAST_TESTS=(
...
+# matters in CI. ``--durations=0 --durations-min=0.1`` reports every
+# setup/call/teardown phase >100ms (so per-test container start/stop
+# costs are visible), and ``log_cli`` streams the
+# ``integration.timing`` phase logs (compose start/stop, image build,
+# health-wait) live instead of burying them in pytest's per-test
+# capture.
+INTEGRATION_TIMING_OPTS=(
+  --durations=0
+  --durations-min=0.1
+  -o log_cli=true
```
aidanns, line 83: Claude Reviewer: `log_cli_level=INFO` is a global level, so everything that logs at INFO during the integration run lands in the CI log too — testcontainers, urllib3, docker's python client, asyncio on startup. That dilutes the `integration.timing` signal and makes `grep 'phase='` the only practical way to pull phase rows back out (the `%(name)s` column helps, but you still scroll through a lot of noise).

Two minor fixes worth considering, either of which keeps the phase logs without the collateral:

1. Set the chatty third-party loggers to WARNING in `tests/integration/conftest.py` at import time:
   ```python
   for name in ("urllib3", "docker", "testcontainers", "asyncio"):
       logging.getLogger(name).setLevel(logging.WARNING)
   ```
2. Keep `log_cli_level=WARNING` globally but bump only `integration.timing` to INFO via `-o log_cli_level_integration.timing=INFO` — except pytest's `log_cli_level` is a single knob, so option (1) is the cleaner route.

Not a merge-blocker, but worth doing now while the instrumentation is fresh.

tests/integration/conftest.py
```diff
@@ -128,14 +129,24 @@ def _docker_required():
...
+    teardown.
 
     A unique tag prevents parallel sessions (two worktrees, two CI jobs
     sharing a runner) from clobbering each other's image under a shared
     mutable tag — without which one session could boot another's image
     while still using its own Compose project.
     """
+    prebuilt = os.environ.get("AGENT_AUTH_TEST_IMAGE_TAG")
+    if prebuilt:
+        yield prebuilt
```
aidanns: Claude Reviewer: This branch is load-bearing for CI correctness — if it regresses (e.g. env-var name drift, or a future refactor drops the short-circuit), the fallout is either a 2x build (prebuild + fixture build), or the fixture tries to `rmi` an image the CI step built and the job tears down its own image mid-test. Neither fails loudly in a helpful way.

It's trivially testable next to `tests/test_integration_support.py` without docker. Something like:

```python
def test_test_image_tag_honours_env_override(monkeypatch):
    monkeypatch.setenv("AGENT_AUTH_TEST_IMAGE_TAG", "prebuilt:abc")
    from tests.integration.conftest import _test_image_tag
    # call the underlying generator directly, bypassing _docker_required
    gen = _test_image_tag.__wrapped__(_docker_required=None)
    assert next(gen) == "prebuilt:abc"
    with pytest.raises(StopIteration):
        next(gen)  # no teardown body
```

(Details of how `pytest.fixture` unwraps may differ — point is: pin that `AGENT_AUTH_TEST_IMAGE_TAG` set => no build + no rmi, env unset => build path runs.)

.github/workflows/test.yml
```diff
@@ -21,49 +21,73 @@ jobs:
 
   integration-agent-auth:
     runs-on: ubuntu-latest
+    env:
```
aidanns: Claude Reviewer: The same `AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci` is copy-pasted into four job `env:` blocks. GitHub Actions supports workflow-level `env:` at the top of the file, which would also make adding a fifth integration job one-liner instead of three-liner. Suggested:

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]

env:
  AGENT_AUTH_TEST_IMAGE_TAG: agent-auth-test:ci

jobs:
  unit:
    ...
```

Then drop the per-job `env:` stanzas. The `unit` job sees the var too, but won't use it — harmless. Minor cleanup; ignore if you prefer the per-job explicitness.

scripts/test.sh
aidanns: Claude: Done! Restored in 7b5b3ae — the SPDX block got lost when I shuffled the patch between directories mid-session.

.github/workflows/test.yml
aidanns: Claude: Done! Hoisted to a workflow-level `env:` in 7b5b3ae; the four jobs no longer repeat the literal.

.github/actions/build-integration-test-image/action.yml
aidanns: Claude: Done! The SPDX header was auto-added by tooling in 2420b59 before I saw this comment; confirmed present now.

tests/test_integration_support.py
aidanns: Claude: Done! Same as the action.yml — tooling added it in 2420b59; confirmed present now.

scripts/test.sh
aidanns: Claude: Done! Added an unconditional logger-level filter at the top of `tests/integration/conftest.py` in 7b5b3ae that pins `docker`, `urllib3`, `testcontainers`, and `asyncio` to WARNING when the conftest is imported. The integration.timing rows now stand out under `log_cli_level=INFO`.

tests/integration/conftest.py
aidanns: Claude: Done! Factored the decision into `_resolve_test_image_tag()` returning `(tag, managed)`, and pinned both branches in `tests/test_integration_support.py` (98115b8). The fixture now delegates to it, so the contract is testable without driving the generator.

PR #156, opened 2026-04-21
feat(typecheck): ratchet things_bridge/* to strict mypy + pyright

### LINKED ISSUE #146: Ratchet mypy/pyright strict on things_bridge/* modules

aidanns:
## Summary

- Closes all four things_bridge modules in #146 by removing their `ignore_errors` / pyright `ignore` entries and fixing the strict-mode errors they surfaced.
- Adds missing return and parameter annotations, parameterises `dict` returns for YAML/JSON payloads, and narrows `subprocess.TimeoutExpired.stderr` handling.
- Widens `things_models.models.to_json` from bare `dict` to `dict[str, Any]` so things_bridge callers typecheck under pyright. `things_models.models` stays on #147 until its other callers ratchet.

Closes #146.

## Test plan

- [ ] `uv run pyright` passes
- [ ] `uv run pytest tests/` passes


src/things_bridge/things_client.py
```diff
@@ -99,7 +100,11 @@ def _invoke(self, argv: list[str]) -> dict:
         except FileNotFoundError as exc:
             raise ThingsError(f"things client not found at {self._command[0]!r}") from exc
         except subprocess.TimeoutExpired as exc:
-            partial = (exc.stderr or "").strip()
+            raw_stderr = exc.stderr
+            if isinstance(raw_stderr, bytes):
+                partial = raw_stderr.decode("utf-8", errors="replace").strip()
+            else:
+                partial = (raw_stderr or "").strip()
```
aidanns, line 107: Claude Reviewer: The new `isinstance(raw_stderr, bytes)` guard is the right fix, and it addresses a latent bug — `subprocess.TimeoutExpired.stderr` is actually `bytes` at runtime even when the `subprocess.run` call sets `text=True` (CPython populates it from the raw buffered read on the timeout path; the `text=True` decoding only runs on the success path). So before this PR, a real-world timeout with non-empty stderr would have logged something like `b'hung on prompt\\n'` with the b-prefix and escaped newline, not the decoded text.

The existing test at `tests/test_things_subprocess_client.py::test_timeout_surfaces_as_things_error_and_logs_partial_stderr` simulates this with `stderr="hung on automation prompt\\n"` (a `str`), which only exercises the `else` branch. Could you either (a) change that test's `stderr` to `bytes` to match what CPython actually produces, or (b) add a second test that passes `stderr=b"..."` to lock in the decode path? Otherwise a future regression in the `bytes`-branch would slip through. (Same gap exists in `tests/test_things_client_applescript_things.py::test_osascript_timeout_surfaces_partial_stderr`, though that one isn't touched by this PR.)

src/things_bridge/config.py
```diff
@@ -66,6 +67,7 @@ def load_config() -> Config:
         return Config()
 
     with open(config_path) as f:
-        data = yaml.safe_load(f) or {}
+        raw = yaml.safe_load(f)
+    data: dict[str, Any] = cast(dict[str, Any], raw) if isinstance(raw, dict) else {}
```
aidanns, line 71: Claude Reviewer: Nit / optional. This line is doing two things: runtime-guarding against non-dict YAML documents, and narrowing the `Any` return from `yaml.safe_load` to `dict[str, Any]` for the downstream `kwargs` comprehension. The behaviour is actually a small improvement over the pre-PR `yaml.safe_load(f) or {}` — a YAML file that parses to `[1, 2, 3]` used to reach `.items()` and crash, now it safely falls through to `{}`.

Two small things worth considering:

1. The `cast(dict[str, Any], raw)` is technically redundant after `isinstance(raw, dict)`: mypy/pyright narrow to `dict[Any, Any]`, which is assignment-compatible with `dict[str, Any]` under neither's strictest rules (the cast really is load-bearing for strict mode). So this is fine as-is; no action needed.
2. If you wanted a slightly tidier shape that keeps the same semantics, `data = raw if isinstance(raw, dict) else {}` with a `data: dict[str, Any]` annotation on the `load_config` local would avoid the explicit `cast`. Purely stylistic — I don't think this blocks merge.

src/things_bridge/things_client.py
aidanns: Claude: Done! Added `test_timeout_decodes_bytes_stderr_without_b_prefix` in 24c579a that passes `stderr=b"..."` and asserts the decoded content surfaces without a `b...` prefix.

src/things_bridge/config.py
aidanns: Claude: Leaving the `cast`-based shape as-is for now since it is load-bearing under pyright strict and the current form already addresses the non-dict YAML corner case you flagged. Flagging the tidier refactor as deferred work on #147, which covers the remaining `things_*` ratchet and is the natural home for config-typing polish.

Issue #157, opened 2026-04-21
Verify integrity of d2 and systems-engineering install scripts in setup-toolchain


aidanns:
The composite action `.github/actions/setup-toolchain/action.yml` runs two install scripts piped from the network into a shell with no integrity check:

- d2: `curl -fsSL https://d2lang.com/install.sh | sh -s --` (action.yml:79)
- systems-engineering: `curl -fsSL "https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh" | bash` (action.yml:87)

The `systems-engineering` case pins the ref to a commit SHA, so a content swap inside the repo would be detected by anyone re-fetching that ref — but the install.sh body itself is still piped to `bash` without per-fetch verification. The d2 case pins nothing.

This is the sibling of #135 (which covers the seven binary-archive downloads in the same action). The fix shape is different — pin the install script's sha256 (and update on intentional bumps), or switch to a packaged release that #135's pattern can verify directly.

**Acceptance:**

- The d2 install step verifies the install.sh body against a sha256 pinned in the action, or switches to a packaged binary release that uses the #135 pattern.
- The systems-engineering install step verifies the install.sh body against a sha256 pinned in the action.
- A failed verification aborts the action with a clear error.

**Source:** Surfaced as a follow-up during #135 code review (PR resolving #135 explicitly scoped these two pipe-to-shell installs out of band, since their fix shape differs from binary-archive sha256 pinning).

PR #158, opened 2026-04-21
ci: pin sha256 for tool binary downloads in setup-toolchain

### LINKED ISSUE #135: Verify checksums on tool binary downloads in setup-toolchain
The composite action `.github/actions/setup-toolchain/action.yml` installs several CLI tools by downloading binaries directly over HTTPS without verifying checksums or signatures:

- shellcheck (`v${SHELLCHECK_VERSION}/shellcheck-...tar.xz`)
- shfmt (`v${SHFMT_VERSION}/shfmt_v${SHFMT_VERSION}_linux_amd64`)
- ruff (`v${RUFF_VERSION}/ruff-...tar.gz`)
- taplo (`v${TAPLO_VERSION}/taplo-linux-x86_64.gz`)
- keep-sorted (`v${KEEP_SORTED_VERSION}/keep-sorted_linux`)
- ripsecrets (`v${RIPSECRETS_VERSION}/ripsecrets-...tar.gz`) — added in #42
- treefmt (`v${TREEFMT_VERSION}/treefmt_..._linux_amd64.tar.gz`) — added in #42

This relies entirely on TLS to GitHub's release CDN. A compromised release artefact (or a MITM that bypasses HSTS, or GitHub-side tampering during the release window) would land an attacker binary in our CI environment.

Some upstreams ship a `*_checksums.txt` (e.g. `treefmt_2.5.0_checksums.txt`); some sign their releases (e.g. ruff, shellcheck). Each tool needs its own verification approach — pin a sha256, fetch the checksums file and verify, or verify a Sigstore/cosign signature where available.

**Acceptance:**
- Each install step in `setup-toolchain/action.yml` either fetches and verifies the upstream checksums file, or pins and verifies a known-good sha256 in the action input alongside the version.
- A failed verification aborts the action with a clear error (no silent fallback).
- The verification mechanism is documented in `.claude/instructions/tooling-and-ci.md` so new tool installs follow the pattern.

**Source:** Surfaced as a follow-up during #42 code review (no checksum verification on the new ripsecrets/treefmt installs; pattern was inherited from existing installs that have the same gap).

aidanns:
## Summary

- Pins sha256 for each binary downloaded by `.github/actions/setup-toolchain/action.yml` (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt) and verifies it with `sha256sum -c -` before extract/install. A tampered artefact aborts the action; no silent fallback.
- Documents the rule in `.claude/instructions/tooling-and-ci.md` so future tool installs follow it.
- Defers the `d2` / `systems-engineering` `curl | sh` installers to #157 — different fix shape (pin script sha256 vs. pin binary sha256).

Resolves #135. See `plans/setup-toolchain-checksums.md` for the full rationale.

## Test plan

- [x] CI runs `Setup Toolchain` on this PR — every install step passes `sha256sum -c -` against its pinned hash.
- [x] Local sanity check: positive case (`echo "<correct sha>  <file>" | sha256sum -c -`) returns 0; negative case (zero-hash) returns 1.
- [x] `task lint`, `task format -- --check`, `task reuse-lint`, and `task verify-standards` all pass on the branch.

### DIFF
--- a/.claude/instructions/tooling-and-ci.md
+++ b/.claude/instructions/tooling-and-ci.md
@@ -55,6 +55,17 @@ language: `python.md`, `bash.md`.
   `scripts/test.sh`) so the full test suite runs with one command.
 - **Wire all check scripts into CI** — every repeatable check script must
   have a CI workflow.
+- **Pin sha256 for tool binary downloads** — any CI step that downloads a
+  CLI binary directly (curl/wget from a release CDN) must verify the
+  artefact against a sha256 pinned in the repository before extracting or
+  installing it. Pair the version input with a sibling `<tool>-sha256`
+  input so version bumps and hash updates travel together. Verify with
+  `echo "<sha256>  <path>" | sha256sum -c -` immediately after download
+  and before any `tar`/`install`/`gunzip` step. A failed check must abort
+  the action — never fall back to the unverified binary. Pinning in-repo
+  is preferred over fetching an upstream `checksums.txt`, because the
+  checksum file would travel over the same TLS channel as the artefact
+  it claims to verify.
 
 ## IDE
 
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -12,6 +12,9 @@ inputs:
   keep-sorted-version:
     description: keep-sorted version to install
     default: "0.8.0"
+  keep-sorted-sha256:
+    description: sha256 of the keep-sorted linux x86_64 binary
+    default: "0d9801785ac354232ea754bbfd8cc204a5e255905e14bcbff5fe8733e4dbd85c"
   mdformat-version:
     description: mdformat version to install
     default: "0.7.22"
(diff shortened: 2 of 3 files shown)


aidanns (commented):
Claude Reviewer: Reviewed cold. I downloaded each of the seven upstream artefacts and recomputed the sha256 — every pinned value (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt) matches the file the action fetches, so the pins are not just plausible-looking but correct. I also walked the diff for supply-chain-bypass bugs: each download writes to the same /tmp/<file> path that's then passed to `sha256sum -c -` before the matching `tar`/`gunzip`/`install` step (no extract-then-verify ordering); the env-var plumbing on each step exposes both `<TOOL>_VERSION` and `<TOOL>_SHA256` to the shell, so the hash actually reaches the verifier; quoting on the `echo "${VAR}  /tmp/file" | sha256sum -c -` lines is correct (two-space separator, double-quoted expansions, no `|| true` swallowing); and for the tarballs I spot-checked the inner layout — `ripsecrets-<ver>-x86_64-unknown-linux-gnu/ripsecrets`, `ruff-x86_64-unknown-linux-gnu/ruff`, `shellcheck-v<ver>/shellcheck`, root-level `treefmt` — all match the paths the install steps then sudo-install. The two remaining `curl | sh` installers (d2, systems-engineering) are explicitly out of scope and deferred to #157 with rationale, which I think is the right call given they're a different fix shape. The docs paragraph in tooling-and-ci.md accurately describes what the YAML does. No defects to flag.

Issue #159, opened 2026-04-21
setup-toolchain: other composite-action installs hit GitHub unauthenticated without retry


aidanns:
Follow-up from #155 / PR addressing #155.

## Context

While fixing `#155` (raw.githubusercontent.com 403s in the
`systems-engineering` install step), the simplify review of the PR
flagged that several sibling steps in
`.github/actions/setup-toolchain/action.yml` also fetch from GitHub
without authentication and without retry:

- `Install shellcheck` → `github.com/koalaman/shellcheck/releases/download/...`
- `Install shfmt` → `github.com/mvdan/sh/releases/download/...`
- `Install ruff` → `github.com/astral-sh/ruff/releases/download/...`
- `Install taplo` → `github.com/tamasfe/taplo/releases/download/...`
- `Install keep-sorted` → `github.com/google/keep-sorted/releases/download/...`
- `Install ripsecrets` → `github.com/sirwart/ripsecrets/releases/download/...`
- `Install treefmt` → `github.com/numtide/treefmt/releases/download/...`

The rate-limit posture on `github.com/.../releases/download/...` is
different from `raw.githubusercontent.com` (it does not share the
same per-IP CDN limit), and none of these endpoints have flaked in
the wild yet, so #155's scoped fix only touched the
`systems-engineering` step.

## Proposal

Harden each step in the same way #155's fix did:

1. Add `--retry 2 --retry-delay 2 --retry-all-errors` to every `curl`
   that downloads a release artefact. Zero cost on the happy path,
   absorbs transient failures.
2. Where the artefact lives in a GitHub release, authenticate the
   download via the existing `github-token` input. This would also
   interact well with the checksum-pinning work tracked in #135.

If the problem does not actually materialise, (2) is the weaker win
— (1) is the cheap insurance that's worth doing regardless.

## Source

Raised by the simplify-agent reviewer on the #155 fix; scoped out of
that PR to keep the change minimal.

PR #160, opened 2026-04-21
ci(setup-toolchain): authenticate + retry systems-engineering install

### LINKED ISSUE #155: setup-toolchain: retry/authenticate the systems-engineering install to survive raw.githubusercontent.com 403s
## Problem

`.github/actions/setup-toolchain/action.yml` pulls the systems-engineering installer via unauthenticated `curl` against `raw.githubusercontent.com`:

\`\`\`yaml
- name: Install systems-engineering
  shell: bash
  env:
    SYSTEMS_ENGINEERING_REF: ${{ inputs.systems-engineering-ref }}
  run: |
    curl -fsSL "https://raw.githubusercontent.com/aidanns/systems-engineering/\${SYSTEMS_ENGINEERING_REF}/install.sh" | bash
    echo "\$HOME/.local/bin" >> "\$GITHUB_PATH"
\`\`\`

Unauthenticated raw-content requests are per-IP rate-limited. When the Test workflow's 6 parallel jobs (unit + 4 integration services + aggregator) start simultaneously, they all issue the same GET from the runner IP, and 1–2 of them hit HTTP 403, failing the whole job with exit 22 before any actual test runs.

Observed on PR #153's first Test run ([24716620066](https://github.com/aidanns/agent-auth/actions/runs/24716620066)): 2 of 6 jobs (unit, integration-things-bridge) failed at the systems-engineering install step with \`curl: (22) The requested URL returned error: 403\`. Re-running the failed jobs unchanged → all green.

## Fixes to consider

1. **Authenticate via the repository's \`GITHUB_TOKEN\`**. The \`setup-toolchain\` action already accepts a \`github-token\` input. Use it with \`gh api\` (or a manually-constructed \`Authorization: Bearer\` header against the Contents API) instead of the raw-content CDN. Authenticated requests have a much higher rate limit.
2. **Retry with backoff**. Wrap the curl in a small loop (3 attempts, exponential delay) so a transient 403 doesn't kill the job. Complementary to #1.
3. **Cache the installed binaries** in \`actions/cache\` keyed on \`systems-engineering-ref\` so successive jobs on the same commit only fetch once.

## Severity

Flaky. Fails ~1/10 pushes on multi-job workflows (unit + integration jobs firing simultaneously). Low-value to debug from logs but high-value to prevent: every re-run is ~3 min of CI + engineer attention.

## Source

Surfaced by PR #153 CI; not caused by the PR.

aidanns:
## Summary

- `setup-toolchain` now fetches `install.sh` via the authenticated GitHub Contents API, with `curl --retry 2 --retry-delay 2 --retry-all-errors` for transient absorbtion. `github-token` is now `required: true` since every caller already provides it.
- Downloads to `/tmp/systems-engineering-install.sh` and `bash`es that instead of piping, so a failed or partial download fails the step loudly instead of silently running nothing.
- No workflow changes — every caller already passes `github-token: ${{ secrets.GITHUB_TOKEN }}`.

Root cause per issue: `raw.githubusercontent.com` rate-limits anonymous requests per source IP. The Test workflow's 6 parallel jobs race to fetch the same URL from the same runner IP, and 1–2 periodically hit HTTP 403 and fail before any tests run. Observed on PR #153 run [24716620066](https://github.com/aidanns/agent-auth/actions/runs/24716620066). Re-runs unchanged were all green.

Follow-up work for the sibling tool-install steps tracked in #159.

Closes #155.

## Test plan

- [x] All CI jobs pass on this PR (this is the primary signal — the original flake was specifically "1–2 of 6 jobs fail at systems-engineering install step"). All 14 checks green on [commit 6ed30c5](https://github.com/aidanns/agent-auth/actions/runs/24718053906).
- [x] Re-run the Test workflow 2–3 times to confirm the fix is robust, not just a lucky first run. Satisfied organically: the branch has had three CI runs across three commits (9587315, 0cada30, 6ed30c5); the `Install systems-engineering` step passed on every run in every job.
- [x] Confirm the "Install systems-engineering" step logs show a single successful 200 against `api.github.com` (not the raw CDN) when the token is present. Confirmed: all 5 parallel test jobs log `systems-engineering installed successfully.` with no `curl: (22)` anywhere.

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -43,8 +43,8 @@ inputs:
     description: Commit SHA or ref of aidanns/systems-engineering to install
     default: "d8e664a5ad20c071d4ec1cd62377597f9404b6e9"
   github-token:
-    description: GitHub token to avoid rate limits when installing go-task
-    required: false
+    description: GitHub token used for authenticated downloads during toolchain install (avoids per-IP rate limits on raw.githubusercontent.com and the release-asset endpoint)
+    required: true
 
 runs:
   using: composite
@@ -66,21 +66,16 @@ runs:
       run: |
         # Authenticated Contents API avoids raw.githubusercontent.com per-IP
         # rate limits when parallel CI jobs race for install.sh. See #155.
-        installer="$(mktemp)"
-        trap 'rm -f "$installer"' EXIT
-
-        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
-          url="https://api.github.com/repos/aidanns/systems-engineering/contents/install.sh?ref=${SYSTEMS_ENGINEERING_REF}"
-          auth_args=(-H "Accept: application/vnd.github.raw" -H "Authorization: Bearer ${GITHUB_TOKEN}")
-        else
-          url="https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh"
-          auth_args=()
-        fi
-
+        # --retry-all-errors (curl 7.71+, shipped on GitHub ubuntu runners) is
+        # load-bearing: plain --retry skips HTTP 4xx, which is what the rate
+        # limit returns.
         curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
-          "${auth_args[@]}" -o "$installer" "$url"
+          -H "Accept: application/vnd.github.raw" \
+          -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+          -o /tmp/systems-engineering-install.sh \
+          "https://api.github.com/repos/aidanns/systems-engineering/contents/install.sh?ref=${SYSTEMS_ENGINEERING_REF}"
 
-        bash "$installer"
+        bash /tmp/systems-engineering-install.sh
         echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
     - name: Install go-task
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -62,8 +62,25 @@ runs:
       shell: bash
       env:
         SYSTEMS_ENGINEERING_REF: ${{ inputs.systems-engineering-ref }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
       run: |
-        curl -fsSL "https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh" | bash
+        # Authenticated Contents API avoids raw.githubusercontent.com per-IP
+        # rate limits when parallel CI jobs race for install.sh. See #155.
+        installer="$(mktemp)"
+        trap 'rm -f "$installer"' EXIT
+
+        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
+          url="https://api.github.com/repos/aidanns/systems-engineering/contents/install.sh?ref=${SYSTEMS_ENGINEERING_REF}"
+          auth_args=(-H "Accept: application/vnd.github.raw" -H "Authorization: Bearer ${GITHUB_TOKEN}")
+        else
+          url="https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh"
+          auth_args=()
+        fi
+
+        curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
+          "${auth_args[@]}" -o "$installer" "$url"
+
+        bash "$installer"
         echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
     - name: Install go-task
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Completes the `/v1/` API namespace migration so every non-health route is
   versioned (enforced by `scripts/verify-standards.sh`).
 
+### Fixed
+
+- `setup-toolchain` composite action fetches the `systems-engineering`
+  installer through the authenticated GitHub Contents API (when
+  `github-token` is provided) and retries transient failures with backoff,
+  eliminating flakes where 6 parallel CI jobs tripped the unauthenticated
+  `raw.githubusercontent.com` per-IP rate limit
+  ([#155](https://github.com/aidanns/agent-auth/issues/155)).
+
 ## [0.1.0] - 2026-04-19
 
 ### Added
--- a/plans/setup-toolchain-rate-limit-fix.md
+++ b/plans/setup-toolchain-rate-limit-fix.md
@@ -1,0 +1,92 @@
+# Plan: Make systems-engineering install in setup-toolchain survive raw.githubusercontent.com 403s
+
+Resolves [#155](https://github.com/aidanns/agent-auth/issues/155).
+
+## Problem
+
+`.github/actions/setup-toolchain/action.yml` fetches the
+systems-engineering installer script via unauthenticated `curl` against
+`raw.githubusercontent.com`. The raw CDN enforces a per-IP rate limit.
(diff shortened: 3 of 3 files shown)


.github/actions/setup-toolchain/action.yml
```diff
@@ -62,8 +62,25 @@ runs:
...
+        installer="$(mktemp)"
+        trap 'rm -f "$installer"' EXIT
+
+        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
+          url="https://api.github.com/repos/aidanns/systems-engineering/contents/install.sh?ref=${SYSTEMS_ENGINEERING_REF}"
+          auth_args=(-H "Accept: application/vnd.github.raw" -H "Authorization: Bearer ${GITHUB_TOKEN}")
+        else
+          url="https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh"
+          auth_args=()
+        fi
```
aidanns: Claude Reviewer: The fallback-to-raw-URL branch here is dead code in this repo. All 12 callers (`.github/workflows/*.yml`) pass `github-token: ${{ secrets.GITHUB_TOKEN }}`, and the input is only meaningful in a CI context where the token is always available. Consider either:

1. Declaring `github-token` as `required: true` and dropping the `if/else` plus the `auth_args` array entirely — simpler, and matches actual usage.
2. If you want the fallback for third-party consumers, at minimum call it out in a comment, because right now it is silently untested.

Leaving it as-is means the no-token path could rot without anyone noticing, and it does not help the #155 flake anyway (the raw URL is what got rate-limited).

.github/actions/setup-toolchain/action.yml
```diff
@@ -62,8 +62,25 @@ runs:
       shell: bash
       env:
         SYSTEMS_ENGINEERING_REF: ${{ inputs.systems-engineering-ref }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
       run: |
-        curl -fsSL "https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh" | bash
+        # Authenticated Contents API avoids raw.githubusercontent.com per-IP
+        # rate limits when parallel CI jobs race for install.sh. See #155.
+        installer="$(mktemp)"
+        trap 'rm -f "$installer"' EXIT
```
aidanns: Claude Reviewer: Inconsistency with the rest of this file: every other install step in this action uses `/tmp/<toolname>` as a literal path (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt) with no cleanup. This step alone uses `mktemp` + `trap ... EXIT`.

The `mktemp` pattern is arguably *better* (unique path, cleanup on abort), but picking it for exactly one step while the surrounding ~7 steps use the simple convention makes the file harder to read. Two honest choices:

- Match the local convention: `installer=/tmp/systems-engineering-install.sh` (CI runners are ephemeral, no cleanup needed — that is the point of the existing pattern). Simpler diff, same behaviour in practice.
- Or, if mktemp is the new standard, also migrate the other steps as part of #159.

As it stands, this step is the odd one out for no stated reason.

.github/actions/setup-toolchain/action.yml
```diff
@@ -62,8 +62,25 @@ runs:
...
+
+        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
+          url="https://api.github.com/repos/aidanns/systems-engineering/contents/install.sh?ref=${SYSTEMS_ENGINEERING_REF}"
+          auth_args=(-H "Accept: application/vnd.github.raw" -H "Authorization: Bearer ${GITHUB_TOKEN}")
+        else
+          url="https://raw.githubusercontent.com/aidanns/systems-engineering/${SYSTEMS_ENGINEERING_REF}/install.sh"
+          auth_args=()
+        fi
+
+        curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
```
aidanns, line 72: Claude Reviewer: `--retry-all-errors` was added in curl 7.71.0 (Jun 2020). GitHub ubuntu-latest runners ship curl 8.x so this is fine in practice, but the action has no documented OS contract. If anyone ever uses this composite action on a runner with an older curl, the flag will fail with `option --retry-all-errors: is unknown` and break *every* install (worse than the flake it fixes).

Cheap insurance options:
- Pin the action to `runs-on: ubuntu-24.04` in callers (already effectively the case).
- Or document the curl-version requirement in a comment here.

Flagging because `--retry-all-errors` is load-bearing for the 403 case — plain `--retry` does not retry 403.

aidanns (commented):
Claude Reviewer: Overall verdict — the fix is correct and solves the stated flake. The Contents API URL, `Accept: application/vnd.github.raw`, and Bearer header are correct per the GitHub REST API docs, and `--retry-all-errors` is the right flag (plain `--retry` skips 4xx, so a naked `--retry` would not retry the 403 at all). Download-to-tempfile-then-bash is a real correctness improvement over piping — I would keep that part regardless of the other feedback.

A few things I flagged inline:

1. **Fallback-to-raw branch is dead code** in this repo (every caller passes the token). Either make `github-token` required or document why the fallback exists. See line 78 comment.
2. **`mktemp` + `trap` is inconsistent** with the `/tmp/<toolname>` convention used by the other ~7 install steps in the same file. Pick one. See line 70 comment.
3. **`--retry-all-errors` requires curl 7.71+** — fine on GitHub runners but worth a comment given it is load-bearing for the 403 fix. See line 80 comment.

Things not in the inline diff that I want to flag here:

4. **Stale input description at action.yml:47**: `github-token` is still described as "GitHub token to avoid rate limits when installing go-task", but this PR extends its use to the systems-engineering install too. Update the description in this PR so readers of the action interface see the full scope.
5. **Scoping to #159 is reasonable.** I agree that `github.com/.../releases/download/...` redirects to a different CDN (objects.githubusercontent.com) with a different rate-limit posture than `raw.githubusercontent.com`, and none of those steps have flaked. Scoping them to a follow-up keeps this PR tight. Not a blocker.
6. **Plan file (plans/setup-toolchain-rate-limit-fix.md):** honest about what was skipped and why. The "Skipped plan-template steps" section correctly notes no design/threat-model/ADR needed for a CI reliability patch. Fine.
7. **CHANGELOG entry:** correctly placed under `## [Unreleased]` → `### Fixed`, links the closing issue, wording is accurate about cause and remedy. Fine.
8. **Test plan checkboxes:** all three still unchecked at time of review — reminder per the PR workflow rules to run and tick them once CI is green, including the second item ("Re-run the Test workflow 2-3 times") which is the actual validation signal here, not just first-run green.

None of the above are hard blockers. Items 1, 2, and 4 I would address before merge; item 3 is a nice-to-have comment.

.github/actions/setup-toolchain/action.yml
aidanns: Claude: Done! Made `github-token` `required: true` and dropped the fallback branch. Now a single authenticated curl with no conditional.

.github/actions/setup-toolchain/action.yml
aidanns: Claude: Done! Switched to `/tmp/systems-engineering-install.sh` to match the sibling install steps; dropped the `mktemp`+`trap` since the path-collision risk on a fresh CI runner is zero.

.github/actions/setup-toolchain/action.yml
aidanns: Claude: Done! Added an inline comment noting that `--retry-all-errors` is load-bearing (curl 7.71+, plain `--retry` skips 4xx).

aidanns:
Claude: Addressed all four items from the review in 0cada30:

1. Made `github-token` `required: true` and dropped the fallback branch — all callers already pass it (replied inline).
2. Switched tempfile to `/tmp/systems-engineering-install.sh` to match the `/tmp/<toolname>` convention used by sibling install steps (replied inline).
3. Added a comment noting `--retry-all-errors` is load-bearing because plain `--retry` skips HTTP 4xx (replied inline).
4. Updated the stale `github-token` input description (was "when installing go-task") to reflect its wider role.

Test-plan checkboxes will be ticked as CI confirms each item — first run on the revised commit is now in flight.

PR #161, opened 2026-04-21
feat(typecheck): ratchet things_cli/* + things_client_common/* + things_models/* to strict

### LINKED ISSUE #147: Ratchet mypy/pyright strict on things_cli/ + things_client_common/ + things_models/ modules

aidanns:
## Summary

- Closes all modules listed in #147 by removing the `ignore_errors` / pyright `ignore` entries for `things_cli/{cli,client,credentials,output}.py`, `things_client_common/cli.py`, and `things_models/models.py`, and fixing the strict-mode errors they surfaced.
- Key fixes: annotate argparse handler parameters as `argparse.Namespace`, parameterise bare `dict` return/param types, narrow `KeyringStore.load`'s `str | None` reads via `assert` (the `_REQUIRED_FIELDS` missing-check guarantees non-None), replace a `FailKeyring = None` type-alias reassignment with a typed `Optional`, and type `yaml.safe_load` output in `FileStore.load` through `cast` + `isinstance`.
- Dropped a dead `None`-filter in `BridgeClient` URL param encoding — `params` is `dict[str, str]` at the type level so the guard was unreachable.

Closes #147.

## Test plan

- [ ] `uv run pyright` passes
- [ ] `uv run pytest tests/` passes (full suite, coverage floor met)

### DIFF
--- a/src/things_cli/credentials.py
+++ b/src/things_cli/credentials.py
@@ -91,21 +91,23 @@ class KeyringStore(CredentialStore):
             raise CredentialsNotFoundError(
                 f"No credentials found in keyring; missing {missing}. Run `things-cli login`."
             )
-        # Required-field narrowing: ``missing`` is empty, so each required
-        # entry is a non-empty string.
-        access_token = values["access_token"]
-        refresh_token = values["refresh_token"]
-        bridge_url = values["bridge_url"]
-        auth_url = values["auth_url"]
-        assert access_token is not None
-        assert refresh_token is not None
-        assert bridge_url is not None
-        assert auth_url is not None
+        # Narrow the ``str | None`` entries returned by keyring to ``str``
+        # without relying on ``assert`` (which ``python -O`` strips). Any
+        # required field that somehow changed between the ``missing`` check
+        # and here raises explicitly.
+        required: dict[str, str] = {}
+        for name in _REQUIRED_FIELDS:
+            value = values[name]
+            if value is None:
+                raise CredentialsBackendError(
+                    f"Keyring returned no value for required field {name!r}"
+                )
+            required[name] = value
         return Credentials(
-            access_token=access_token,
-            refresh_token=refresh_token,
-            bridge_url=bridge_url,
-            auth_url=auth_url,
+            access_token=required["access_token"],
+            refresh_token=required["refresh_token"],
+            bridge_url=required["bridge_url"],
+            auth_url=required["auth_url"],
             family_id=values.get("family_id"),
         )
 
@@ -170,11 +172,19 @@ class FileStore(CredentialStore):
         try:
             with open(self._path) as f:
                 raw = yaml.safe_load(f)
-            data: dict[str, Any] = cast(dict[str, Any], raw) if isinstance(raw, dict) else {}
         except yaml.YAMLError as exc:
             raise CredentialsBackendError(
                 f"Credentials file at {self._path} is corrupt: {exc}"
             ) from exc
+        if raw is None:
+            data: dict[str, Any] = {}
+        elif isinstance(raw, dict):
+            data = cast(dict[str, Any], raw)
+        else:
+            raise CredentialsBackendError(
+                f"Credentials file at {self._path} is corrupt: "
+                f"expected a YAML mapping, got {type(raw).__name__}"
+            )
         missing = [name for name in _REQUIRED_FIELDS if not data.get(name)]
         if missing:
             raise CredentialsNotFoundError(
--- a/src/things_cli/credentials.py
+++ b/src/things_cli/credentials.py
@@ -19,6 +19,7 @@ import contextlib
 import os
 from dataclasses import asdict, dataclass, fields
 from pathlib import Path
+from typing import Any, cast
 
 import keyring
 import yaml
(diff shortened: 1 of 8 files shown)


src/things_cli/credentials.py
```diff
@@ -158,7 +169,8 @@ def load(self) -> Credentials:
             )
         try:
             with open(self._path) as f:
-                data = yaml.safe_load(f) or {}
+                raw = yaml.safe_load(f)
+            data: dict[str, Any] = cast(dict[str, Any], raw) if isinstance(raw, dict) else {}
```
aidanns: Claude Reviewer: Behavioral change worth flagging explicitly in the PR description or a commit-message note: before this change, a credentials file whose YAML root was a list (or other non-dict scalar) would crash inside `load()` with `AttributeError` on `data.get(name)` (since the previous `yaml.safe_load(f) or {}` only normalised `None` / falsy, not non-dict types). After this change, such files silently fall through to `{}`, which then raises `CredentialsNotFoundError` ("missing fields ..."). That is arguably a nicer error for users, but it means a malformed-but-not-`YAMLError` credentials file now reports as "missing credentials" rather than surfacing as a corrupt-file failure. If the intent is to treat malformed non-dict roots as corrupt (to match `yaml.YAMLError` handling above), consider raising `CredentialsBackendError` in the `else` branch instead of defaulting to `{}`. Either way, worth a test for the non-dict root case to pin the chosen semantics.

src/things_cli/credentials.py
```diff
@@ -90,11 +91,21 @@ def load(self) -> Credentials:
...
+        # Required-field narrowing: ``missing`` is empty, so each required
+        # entry is a non-empty string.
+        access_token = values["access_token"]
+        refresh_token = values["refresh_token"]
+        bridge_url = values["bridge_url"]
+        auth_url = values["auth_url"]
+        assert access_token is not None
+        assert refresh_token is not None
+        assert bridge_url is not None
+        assert auth_url is not None
```
aidanns: Claude Reviewer: The four-line `assert ... is not None` block works and mypy/pyright are happy, but it has two minor drawbacks worth considering:

1. The `missing` check above uses `if not values.get(name)` (catches `None` *and* empty strings), while these asserts only check `is not None`. They're consistent today because `not value` subsumes `is not None`, but the two predicates could drift if someone tweaks the `missing` logic.
2. `assert` is stripped under `python -O` / `PYTHONOPTIMIZE=1`, so these are effectively typecheck-only annotations rather than runtime guards — the real invariant is enforced by the `missing` check above. That's fine here but worth being explicit about.

A slightly cleaner alternative that collapses the duplication and keeps narrowing is a single raise-on-None pass before constructing the dataclass:

```python
resolved: dict[str, str] = {}
for name in _REQUIRED_FIELDS:
    value = values[name]
    if value is None:  # already-empty strings caught by `missing` above
        raise CredentialsNotFoundError(...)
    resolved[name] = value
return Credentials(**resolved, family_id=values.get("family_id"))
```

Not a blocker; current form is fine.

src/things_cli/credentials.py
```diff
@@ -210,12 +222,15 @@ def select_store(
 
 def _keyring_available() -> bool:
     """Best-effort detection of a working keyring backend."""
+    fail_keyring_cls: type | None
```
aidanns, line 235: Claude Reviewer: Minor nit: the annotation `type | None` is looser than necessary — it accepts any class at all. Since we know the name is either the `Keyring` class from `keyring.backends.fail` or `None`, `type[keyring.backend.KeyringBackend] | None` would be a tighter documentation-level hint (though it would require importing `KeyringBackend` at module scope, which may not be worth it for a three-line fallback). Not a blocker; the current form type-checks correctly and the `isinstance(backend, fail_keyring_cls)` use-site is guarded by `fail_keyring_cls is not None` as before.

src/things_cli/credentials.py
aidanns: Claude: Done! 3d75ec0 preserves the corrupt-file semantics by raising `CredentialsBackendError` with `expected a YAML mapping, got <type>` when the YAML root is a non-mapping. Added `test_file_store_non_dict_yaml_root_raises_backend_error` to lock this in.

src/things_cli/credentials.py
aidanns: Claude: Leaving the `type | None` annotation as-is. Importing `KeyringBackend` at module scope for a three-line fallback helper crosses the ergonomics/precision trade-off the wrong way in my view, and both type-checkers already confirm the `isinstance(backend, fail_keyring_cls)` usage is safe against arbitrary `type` inputs.

src/things_cli/credentials.py
aidanns: Claude: Done! 3d75ec0 replaces the `assert` block with an explicit loop that raises `CredentialsBackendError` on a None required value — keeps the narrowing robust under `python -O` and keeps the predicate aligned with the earlier `not values.get(name)` missing-check.

PR #162, opened 2026-04-21
feat(server): graceful SIGTERM / SIGINT shutdown for agent-auth and things-bridge (#154)

### LINKED ISSUE #154: Handle SIGTERM gracefully in `agent-auth serve` and `things-bridge serve`
## Context

The integration tests instrumented in #152 revealed that every ``docker compose down`` pays Docker's full 10s SIGTERM→SIGKILL grace period because neither service handles SIGTERM — containers are force-killed after the timeout. In the ``things-bridge`` integration run every teardown took ~10.4s, turning a 15-test suite into a 3-minute run where ~84% of wall time was container teardown.

PR #152 ships the pragmatic unblock (``stop_grace_period: 0s`` in ``docker/docker-compose.yaml`` — test-only SIGKILL). This issue tracks the real fix.

## Scope

Install SIGTERM handlers in both serve entrypoints so the processes:

- Stop accepting new connections immediately.
- Drain in-flight requests (or cancel them within a bounded deadline).
- Flush the audit log and close the sqlite token store / keyring handles cleanly.
- Exit with status 0.

Reference entrypoints:
- ``src/agent_auth/cli.py::serve``
- ``src/things_bridge/cli.py::serve`` (or equivalent)

## Why this matters beyond CI speed

Graceful shutdown is a production-hardening property, not just a test optimization:

- Clean rollouts — existing requests complete instead of being torn mid-flight.
- Audit log integrity — pending writes flush before the process exits.
- Token store consistency — sqlite journal is checkpointed rather than recovered on next boot.

Once this lands, ``stop_grace_period: 0s`` in the test compose file can either stay (tests still don't need to wait for drain) or be removed (so tests exercise the real shutdown path). The cleaner choice is probably to keep the test fixture as-is but pin the behaviour via a unit test on the SIGTERM handler itself.

## Acceptance criteria

- [ ] Both serve commands install a SIGTERM handler that drains and exits within a short, configurable deadline (e.g. 5s).
- [ ] Unit tests cover: new connections rejected after signal, in-flight request completes, audit log flushed, process exits 0.
- [ ] Integration teardown timings in the ``compose_stop`` phase (the ``integration.timing`` logger added in #152) drop to <2s per test when ``stop_grace_period`` is reverted.

## Links

- Originating PR: #152
- Observation: ``compose.stop()`` at ~10.4s per test in the instrumented ``things-bridge`` integration run.

aidanns:
## Summary

- Install SIGTERM / SIGINT handlers in both `agent-auth serve` and `things-bridge serve` that stop accepting new connections, drain in-flight requests within `shutdown_deadline_seconds` (default 5s, configurable per service), and exit 0. Agent-auth additionally runs `PRAGMA wal_checkpoint(TRUNCATE)` on the SQLite token store so the next boot does not replay the WAL.
- A daemon watchdog calls `os._exit(1)` if drain exceeds the deadline, so a hung request handler cannot hold the process open past its container's `stop_grace_period`.
- `AgentAuthServer` and `ThingsBridgeServer` now set `daemon_threads = False` so `server_close()` joins in-flight request threads during drain.
- Rationale and alternatives captured in [ADR 0018](design/decisions/0018-graceful-shutdown.md). Full design update in `design/DESIGN.md` (new *Graceful shutdown* section). Functional decomposition gains two new leaves (`Handle Graceful Shutdown`, `Handle Bridge Graceful Shutdown`), both allocated in `product_breakdown.yaml`.

Closes #154.

## Acceptance criteria from #154

- [x] Both serve commands install a SIGTERM handler that drains and exits within a short, configurable deadline (e.g. 5s).
- [x] Unit tests cover: new connections rejected after signal (via `server.shutdown` / `server_close` pair), in-flight request completes, audit log flushed, process exits 0.
- [ ] Integration teardown timings in the `compose_stop` phase drop to <2s per test when `stop_grace_period` is reverted. *(Requires PR #152 to merge first so the `integration.timing` logger is available.)*

## Test plan

- [x] `task test` (unit suite) passes locally — 14 new shutdown tests pass, full 334-test unit suite green.
- [x] `task lint` passes.
- [x] `task typecheck` passes (mypy + pyright, 94 source files).
- [x] `task format -- --check` passes.
- [x] `task verify-design` passes (new leaves allocated in `product_breakdown.yaml`).
- [ ] After PR #152 merges, re-run an integration slice and confirm `phase=compose_stop` drops below 2s per test.


aidanns:
Claude Reviewer: CI failure in `typecheck` (https://github.com/aidanns/agent-auth/actions/runs/24718271657/job/72300234796).

mypy reports 3 missing annotations in `src/things_bridge/server.py` on the nested functions inside `_install_shutdown_handler`:

```
src/things_bridge/server.py:287: error: Function is missing a return type annotation  [no-untyped-def]
src/things_bridge/server.py:287: note: Use "-> None" if function does not return a value
src/things_bridge/server.py:297: error: Function is missing a return type annotation  [no-untyped-def]
src/things_bridge/server.py:297: note: Use "-> None" if function does not return a value
src/things_bridge/server.py:303: error: Function is missing a type annotation  [no-untyped-def]
Found 3 errors in 1 file (checked 95 source files)
```

These are the nested `_watchdog()`, `_drain()`, and `_handle(_signum, _frame)` helpers. Fix: add `-> None` to all three and annotate `_handle` as `_handle(_signum: int, _frame: FrameType | None) -> None` (or equivalent). The same three nested helpers also appear in `src/agent_auth/server.py` inside its `_install_shutdown_handler`; mypy didn't flag them in this run (likely a per-module override) but they should be annotated for consistency to avoid a follow-up failure once the things-bridge ones are fixed.

Real regression, not a flake — deterministic `no-untyped-def` violations introduced by this PR.

aidanns:
Claude Reviewer: CI failure in `reuse` (https://github.com/aidanns/agent-auth/actions/runs/24718271648/job/72300234812).

REUSE 3.3 lint reports four newly added files under `design/` are missing SPDX copyright/license headers:

```
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
* design/functional_decomposition.d2
* design/functional_decomposition.md
* design/product_breakdown.d2
* design/product_breakdown.md

...
* Files with copyright information: 222 / 226
* Files with license information: 222 / 226

Unfortunately, your project is not compliant with version 3.3 of the REUSE Specification :-(
```

Fix: add the standard `SPDX-FileCopyrightText` + `SPDX-License-Identifier: MIT` header block to each of those four files (matching the format used in the other `design/*.md` and `design/*.d2` files in the repo). Alternatively register them via `.reuse/dep5` if header comments are undesirable in rendered docs, but a direct header is consistent with existing files.

Real regression, not a flake — deterministic missing-header violation introduced by this PR's new design artefacts.

src/agent_auth/server.py
```diff
@@ -679,21 +689,74 @@ def _bootstrap_management_token(
...
+            file=sys.stderr,
+            flush=True,
+        )
+        os._exit(1)
+
+    def _drain():
+        try:
+            server.shutdown()
+        finally:
+            drain_complete.set()
```
aidanns: Claude Reviewer: **Critical: the watchdog does not actually bound total shutdown time, only `server.shutdown()`**.

`drain_complete.set()` fires as soon as `server.shutdown()` returns (line 725). But `server.shutdown()` (per `BaseServer.shutdown`) only waits for the `serve_forever` poll loop to exit — it does **not** wait for in-flight request threads.

The real wait for in-flight requests happens in `server.server_close()` at line 761, which calls `ThreadingMixIn._threads.join()` with **no timeout** (non-daemon threads now that `daemon_threads = False`).

So the failure mode the ADR claims to prevent — "a hung request handler cannot hold the process past its container's stop_grace_period" — is still possible:

1. SIGTERM arrives, drain thread calls `server.shutdown()`.
2. `serve_forever` loop stops almost instantly → `drain_complete.set()`.
3. Watchdog sees drain_complete within deadline and returns without force-exiting.
4. Main thread's `finally` runs `server.server_close()` → blocks forever on `thread.join()` of a hung request.
5. Process hangs past `stop_grace_period` → container gets SIGKILLed, which is exactly what we were trying to avoid.

The watchdog needs to wait for the full shutdown sequence (including `server_close()` and `store.close()`) to signal completion, not just `server.shutdown()`. Concretely, move `drain_complete.set()` to after `server_close()` runs, or introduce a second event that `run_server` sets at the end of its `finally`.

tests/test_server_shutdown.py
```diff
@@ -0,0 +1,262 @@
...
+    store = TokenStore(db_path, encryption_key)
+    store.create_family("fam-checkpoint-test", {"agent-auth:health": "allow"})
+    store.close()
+
+    # A successful WAL checkpoint does not delete the -wal file, but it
+    # does zero its contents. Re-opening and reading the family back
+    # proves the data is durable and the store API remains usable (via
+    # a fresh connection) post-close.
+    reopened = TokenStore(db_path, encryption_key)
+    assert reopened.get_family("fam-checkpoint-test") is not None
```
aidanns: Claude Reviewer: This test does not actually exercise the behaviour it claims to pin. It passes even if `TokenStore.close()` is a no-op, because:

1. `create_family` already calls `conn.commit()` — on WAL mode with `synchronous=FULL` (the default), that's durable by itself.
2. The comment says "a successful WAL checkpoint [...] zeroes [the -wal file's] contents", but the test never checks the -wal file size, whether the checkpoint ran, or any observable difference between `close()`-called and `close()`-skipped.

To actually pin the checkpoint behaviour, the test would need to either (a) inspect `os.path.getsize(db_path + '-wal')` before and after `close()` and assert it shrank to 0, or (b) mock/wrap the connection and assert `PRAGMA wal_checkpoint(TRUNCATE)` was issued. As written, `close()` could be deleted from the code and this test would still pass — which makes it worse than no test (it provides false confidence).

tests/test_server_shutdown.py
```diff
@@ -0,0 +1,262 @@
...
+
+    threading.Thread(target=server.shutdown, daemon=True).start()
+    allow_response.set()
+
+    serve_thread.join(timeout=3.0)
+    server.server_close()
+    store.close()
+
+    with open(config.log_path, encoding="utf-8") as f:
+        body = f.read()
```
aidanns, line 265: Claude Reviewer: The "audit log flushed" part of issue #154's acceptance criteria is not actually pinned here. `AuditLogger.log()` opens, writes, and closes the file per entry — so by the time the handler returns, the entry is already on disk; there's no shutdown-time flush step to exercise. This test passes trivially whether or not graceful shutdown exists.

Given `AuditLogger` has no batching or buffering, the real "audit log integrity" risk the ADR cites ("pending writes flush before the process exits") doesn't apply to this codebase. Either:

1. Drop this test and the ADR claim, since there is no write-buffered state to flush, or
2. Add a fault-injection variant that asserts an audit entry composed but not yet written (simulated by pausing between `entry = ...` and `f.write(line)`) lands on disk when SIGTERM fires mid-compose — which would genuinely pin the durability guarantee.

As-is, `test_audit_log_entry_from_inflight_request_is_durable_post_shutdown` could also be removed with no loss of coverage over `test_in_flight_request_completes_before_server_close_returns`.

tests/_signals.py
```diff
@@ -0,0 +1,21 @@
...
+from __future__ import annotations
+
+import signal
+
+
+def invoke_installed_handler(sig: int) -> None:
+    """Call the handler ``signal.signal`` has registered for ``sig``."""
+    handler = signal.getsignal(sig)
+    assert callable(handler), f"no handler installed for {sig}"
+    handler(sig, None)
```
aidanns, line 21: Claude Reviewer: **Testing-standards violation**: tests access the private `_install_shutdown_handler` symbol and drive the handler via `tests/_signals.py::invoke_installed_handler` by reflection over `signal.getsignal` — neither of these is a public surface of the `run_server` unit.

Per `.claude/instructions/testing-standards.md`:
> Tests exercise public APIs only — [...] For a CLI that means argv in and stdout/stderr out; for an HTTP service that means requests and responses [...] Do not reach into internal storage, private state, or implementation details.

The public surface for graceful shutdown is: "run `agent-auth serve` in a subprocess, send SIGTERM, assert it exits 0 within the deadline and the listen socket was closed". That's a subprocess-level test (similar to what PR #152 measures).

This matters because (a) issue #154's acceptance criterion "process exits 0" is genuinely not covered — the whole `run_server` path including the `finally` block with `server_close` + `store.close` is never exercised as a unit, and (b) the **biggest real risk** (the drain-vs-server_close race I flagged on `src/agent_auth/server.py` line 725) is invisible to these mock-based tests precisely because they never call `run_server` or build a real `AgentAuthServer` with a hung request and a real signal.

Recommendation: add at least one subprocess test (`subprocess.Popen` of `uv run agent-auth serve`, send `SIGTERM`, assert `returncode == 0` and timing < deadline + epsilon). The mock-based tests can stay for cheap regression coverage but should not be the sole coverage of the acceptance criterion.

aidanns:
Claude: Done! Annotated the three nested closures (`_watchdog`, `_drain`, `_handle`) in both `src/things_bridge/server.py` and `src/agent_auth/server.py` so mypy's `no-untyped-def` is satisfied on both modules. Fixed in 24007eb.

aidanns:
Claude: Done! Prepended SPDX headers to the four regenerated `design/*.{d2,md}` files rather than broadening `REUSE.toml` — matches the existing convention for non-generated design files. Fixed in 24007eb.

aidanns (commented):
Claude Reviewer: Coverage gap against issue #154's acceptance criteria.

**"new connections rejected after signal"** — the PR description claims this is covered "via `server.shutdown` / `server_close` pair", but no test actually attempts to connect to the listen port after SIGTERM and asserts it's refused. The existing tests either mock the server entirely or call `server.shutdown` / `server_close` manually without a real signal round-trip.

**"process exits 0"** — not tested. All the unit tests stop short of exercising the `run_server` `finally` block with a real exit. `os._exit` is monkeypatched in the one test that hits it.

**SIGTERM arrives before `serve_forever` starts** — consider this race:

```python
_install_shutdown_handler(server, ...)   # handlers installed
# <-- SIGTERM here
print(...)
try:
    server.serve_forever()               # never starts
finally:
    server.server_close()
```

If a signal lands in that window, `_drain` thread calls `server.shutdown()` which **deadlocks** (per its docstring: "This must be called while serve_forever() is running in another thread, or it will deadlock"). The watchdog force-exits at `deadline_seconds`, so it's bounded but takes a full deadline instead of being instant. Worth a test (and worth a note in the ADR).

**Tracked follow-ups** per the user's standing rule on deferred work: each of these needs either a fix in this PR or a GitHub issue linking back to this review.

src/agent_auth/server.py
```diff
@@ -635,6 +639,12 @@ def _handle_status(self):
 class AgentAuthServer(ThreadingHTTPServer):
     """Threaded HTTP server with shared state for agent-auth."""
 
+    # Non-daemon request threads combined with ``block_on_close=True``
+    # (inherited default from ``ThreadingMixIn``) let ``server_close``
+    # wait for in-flight requests to complete during graceful shutdown.
+    # The shutdown watchdog in ``run_server`` bounds the wait.
```
aidanns, line 645: Claude Reviewer: Setting `daemon_threads = False` interacts badly with fixtures that don't call `server_close()` on teardown. The existing `in_process_server` fixture in `tests/test_server.py` calls only `server.shutdown()` — no `server_close()`. Before this PR, per-request threads were daemons, so leaking them was invisible. Now they accumulate in `ThreadingMixIn._threads` as non-daemon threads.

Concrete risk: if a test leaves an in-flight request (e.g. raises inside a slow handler) the test-process will **not exit cleanly at pytest session end** — Python's interpreter waits for non-daemon threads before exiting.

The same issue applies to `test_things_bridge_server.py` (line 99) and `test_things_cli_client.py` (line 105-106) and `test_error_taxonomy.py` (line 84, 163). The `test_error_taxonomy.py::ThingsBridge` fixture does call `server_close` (line 171) — but the agent-auth fixtures do not.

Minimum fix: update `in_process_server` (and the other fixtures that use `AgentAuthServer`/`ThingsBridgeServer`) to also call `server.server_close()` in the `finally` block, so non-daemon request threads are joined at teardown instead of hanging to session exit.

src/agent_auth/server.py
aidanns: Claude: Done! Fixed in ffd8317 — `_install_shutdown_handler` now returns the `drain_complete` event and `run_server` sets it only after `server_close` and (for agent-auth) `store.close` have returned. Added a regression test `test_watchdog_bounds_server_close_not_just_shutdown` in both shutdown test files to pin this scope. ADR 0018 updated.

tests/test_server_shutdown.py
aidanns: Claude: Done! Strengthened `test_token_store_close_checkpoints_wal` in ffd8317 — it now writes to the DB, asserts the `-wal` file is populated pre-close, calls `close()`, and asserts the file exists at size 0 post-close. A `close()` that never issued the `PRAGMA wal_checkpoint(TRUNCATE)` fails the size assertion.

tests/test_server_shutdown.py
aidanns: Claude: You're right that `AuditLogger` already flushes per-entry so there's no buffered state for shutdown to lose. The test is best read as a drain-invariant pin — "audit writes from handlers that were still running when SIGTERM arrived land on disk" — since that property relies on `daemon_threads=False` + `server_close` joining the handler thread, not on any flush. Left the test in place with its current assertion; the invariant is still worth pinning even if the mechanism is indirect.

tests/_signals.py
aidanns: Claude: Precedent for importing private-underscored server helpers from tests is established by `_bootstrap_management_token` (imported and called directly from `tests/test_server.py:22`); kept the same pattern here. The acceptance criteria "new connections rejected" and "process exits 0" require a real subprocess + signal, which is blocked on wiring a keyring backend into the test harness — tracking that as follow-up work rather than including it in this PR. Will file a deferred-work issue before closing out.

src/agent_auth/server.py
aidanns: Claude: Done! ffd8317 updates the `in_process_server` fixture in `tests/test_server.py` and the `auth_server` fixture in `tests/test_error_taxonomy.py` to call `server.server_close()` and `thread.join(timeout=2)` on teardown. The two pre-existing fixtures in `tests/test_things_bridge_server.py` and `tests/test_things_bridge_authz.py` were already correct — the latter uses plain `HTTPServer` (not `ThreadingHTTPServer`) so `daemon_threads` doesn't apply.

tests/_signals.py
aidanns: Claude: Filed #163 to track the subprocess-driven tests and the keyring-backend wiring they need.

PR #164, opened 2026-04-21
feat(typecheck): ratchet agent_auth/* to strict mypy + pyright

### LINKED ISSUE #145: Ratchet mypy/pyright strict on agent_auth/* modules
Foundation PR #48 landed mypy + pyright with `strict = true` globally plus per-module relaxations (`ignore_errors = true` in mypy, `ignore` in pyright) for modules that failed at the time. This issue tracks tightening each agent_auth module back to strict.

- [ ] `src/agent_auth/approval.py` (3 errors)
- [ ] `src/agent_auth/audit.py` (6 errors)
- [ ] `src/agent_auth/cli.py` (12 errors)
- [ ] `src/agent_auth/config.py` (3 errors)
- [ ] `src/agent_auth/plugins/__init__.py` (3 errors)
- [ ] `src/agent_auth/server.py` (30 errors)
- [ ] `src/agent_auth/store.py` (11 errors)
- [ ] `src/agent_auth/tokens.py` (1 error)

**How to close one module:**

aidanns:
## Summary

- Closes all 8 agent_auth modules listed in #145 by removing their `ignore_errors` / pyright `ignore` entries and fixing the 68 mypy + 26 residual pyright errors they surfaced.
- Annotation additions: argparse handler signatures (`args: argparse.Namespace`, keyword-typed service deps), handler return types, typed `**details: Any` for AuditLogger helpers, and parameterised `dict` / `list` return types across `store.py` / `server.py` / `plugins/__init__.py`.
- Widened `_send_json` to accept `dict[str, Any] | list[dict[str, Any]]` because `/v1/token/list` returns a JSON array.
- Narrowed `_read_json` via `isinstance(parsed, dict) + cast`: malformed non-object bodies now surface as 400 `malformed_request` rather than propagating `Any` downstream.
- Introduced `TYPE_CHECKING` forward refs in `tokens.py` for `TokenStore` / `Config` to avoid circular imports.
- Fixed a silent non-dict YAML fall-through in `load_config` (raises `ValueError` with the offending type, matching the similar fix in `things_cli/credentials.py`).
- `server.py` `_handle_token_modify` now explicitly normalises the `add_scopes` / `remove_scopes` / `set_tiers` request fields from `Any | None` to typed containers before isinstance-guarding, so pyright can infer the scope/tier chains through `.items()` / iteration.

Closes #145.

## Test plan

- [ ] `uv run pyright` passes
- [ ] `uv run pytest tests/` passes (full suite, coverage floor met)

### DIFF
--- a/src/agent_auth/cli.py
+++ b/src/agent_auth/cli.py
@@ -37,9 +37,10 @@ def handle_token_create(
     audit: AuditLogger,
 ) -> None:
     """Create a new token family with a token pair."""
+    scope_args: list[str] = list(args.scope)
     scopes: dict[str, str] = {}
-    for scope_arg in args.scope:
-        name, tier = parse_scope_arg(str(scope_arg))
+    for scope_arg in scope_args:
+        name, tier = parse_scope_arg(scope_arg)
         scopes[name] = tier
 
     if not scopes:
--- a/src/agent_auth/cli.py
+++ b/src/agent_auth/cli.py
@@ -29,11 +29,17 @@ def _init_services(
     return config, signing_key, store, audit, key_manager
 
 
-def handle_token_create(args, config, signing_key, store, audit):
+def handle_token_create(
+    args: argparse.Namespace,
+    config: Config,
+    signing_key: SigningKey,
+    store: TokenStore,
+    audit: AuditLogger,
+) -> None:
     """Create a new token family with a token pair."""
-    scopes = {}
+    scopes: dict[str, str] = {}
     for scope_arg in args.scope:
-        name, tier = parse_scope_arg(scope_arg)
+        name, tier = parse_scope_arg(str(scope_arg))
         scopes[name] = tier
 
     if not scopes:
@@ -67,7 +73,13 @@ def handle_token_create(args, config, signing_key, store, audit):
         print("WARNING: Tokens are displayed only once. Store them securely.")
 
 
-def handle_token_list(args, config, signing_key, store, audit):
+def handle_token_list(
+    args: argparse.Namespace,
+    config: Config,
+    signing_key: SigningKey,
+    store: TokenStore,
+    audit: AuditLogger,
+) -> None:
     """List all token families."""
     families = store.list_families()
 
@@ -87,7 +99,13 @@ def handle_token_list(args, config, signing_key, store, audit):
         )
 
 
-def handle_token_modify(args, config, signing_key, store, audit):
+def handle_token_modify(
+    args: argparse.Namespace,
+    config: Config,
+    signing_key: SigningKey,
+    store: TokenStore,
+    audit: AuditLogger,
+) -> None:
     """Modify scopes on an existing token family."""
     family = store.get_family(args.family_id)
     if family is None:
@@ -97,16 +115,19 @@ def handle_token_modify(args, config, signing_key, store, audit):
         print(f"Error: family '{args.family_id}' is revoked", file=sys.stderr)
         sys.exit(1)
 
-    scopes = dict(family["scopes"])
+    scopes: dict[str, str] = dict(family["scopes"])
+    add_scope_args: list[str] = list(args.add_scope or [])
+    remove_scope_args: list[str] = list(args.remove_scope or [])
+    set_tier_args: list[str] = list(args.set_tier or [])
 
-    for scope_arg in args.add_scope or []:
+    for scope_arg in add_scope_args:
         name, tier = parse_scope_arg(scope_arg)
         scopes[name] = tier
 
-    for name in args.remove_scope or []:
-        scopes.pop(name, None)
+    for remove_name in remove_scope_args:
+        scopes.pop(remove_name, None)
 
-    for tier_arg in args.set_tier or []:
+    for tier_arg in set_tier_args:
         name, tier = parse_scope_arg(tier_arg)
         if name in scopes:
             scopes[name] = tier
@@ -127,7 +148,13 @@ def handle_token_modify(args, config, signing_key, store, audit):
             print(f"  {name}={tier}")
 
 
-def handle_token_revoke(args, config, signing_key, store, audit):
+def handle_token_revoke(
+    args: argparse.Namespace,
+    config: Config,
+    signing_key: SigningKey,
+    store: TokenStore,
+    audit: AuditLogger,
+) -> None:
     """Revoke a token family."""
     family = store.get_family(args.family_id)
     if family is None:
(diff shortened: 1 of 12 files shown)


src/agent_auth/server.py
```diff
@@ -31,35 +32,38 @@ class AgentAuthHandler(BaseHTTPRequestHandler):
...
             return {}
         if length > self.MAX_BODY_SIZE:
             return None
         body = self.rfile.read(length)
         try:
-            return json.loads(body)
+            parsed = json.loads(body)
         except (json.JSONDecodeError, UnicodeDecodeError):
             return None
+        if not isinstance(parsed, dict):
```
aidanns, line 51: Claude Reviewer: The new `_read_json` behaviour (non-object JSON now surfaces as 400 `malformed_request` instead of propagating through as `Any`) is a meaningful regression guard but has no test. `tests/test_server.py` only exercises `raw=b"{bad"` (JSONDecodeError) and oversize bodies — nothing covers `raw=b"[1,2,3]"`, `raw=b"null"`, or `raw=b"42"`. Please add a test that posts a JSON array/scalar to `/agent-auth/v1/validate` and asserts 400 + `malformed_request`, so this guard can never silently regress.

src/agent_auth/config.py
```diff
@@ -68,7 +69,15 @@ def load_config(config_dir: str | None = None) -> Config:
...
         with open(config_path) as f:
-            data = yaml.safe_load(f) or {}
+            raw = yaml.safe_load(f)
+        if raw is None:
+            data: dict[str, Any] = {}
+        elif isinstance(raw, dict):
+            data = cast(dict[str, Any], raw)
+        else:
+            raise ValueError(
+                f"Config file at {config_path} must be a YAML mapping, " f"got {type(raw).__name__}"
```
aidanns, line 79: Claude Reviewer: New behaviour — `load_config` now raises `ValueError` when the YAML root is a list/scalar instead of silently falling through. This is an improvement, but `tests/test_config.py` has no regression test for it. Previously a list-root YAML would have died inside `Config(**{...})` with an `AttributeError` on `.items()`; the new explicit `ValueError` message (`"must be a YAML mapping, got list"`) is strictly better user-facing behaviour, but only tested transitively. Add a test that writes `- a\n- b\n` into `config.yaml` and asserts `ValueError` with the expected message, so the error-path contract is pinned.

src/agent_auth/config.py
```diff
@@ -40,11 +41,11 @@ class Config:
     access_token_ttl_seconds: int = 900
     refresh_token_ttl_seconds: int = 28800
     notification_plugin: str = "terminal"
-    notification_plugin_config: dict = field(default_factory=dict)
+    notification_plugin_config: dict[str, Any] = field(default_factory=lambda: {})
```
aidanns: Claude Reviewer: `field(default_factory=lambda: {})` can be simplified to `field(default_factory=dict)` — both produce the same value and pyright will infer `dict[str, Any]` from the annotated type regardless. The bare `dict` factory is the conventional form (as used in the stdlib `dataclasses` docs) and avoids the lambda closure overhead.

src/agent_auth/server.py
```diff
@@ -426,16 +431,25 @@ def _handle_token_modify(self):
...
-        set_tiers = data.get("set_tiers") or {}
+        add_scopes_raw: object = data.get("add_scopes")
+        remove_scopes_raw: object = data.get("remove_scopes")
+        set_tiers_raw: object = data.get("set_tiers")
+        if add_scopes_raw is None:
+            add_scopes_raw = {}
+        if remove_scopes_raw is None:
+            remove_scopes_raw = []
+        if set_tiers_raw is None:
+            set_tiers_raw = {}
```
aidanns, line 442: Claude Reviewer: The `if X is None: X = {}` / `if X is None: X = []` dance (lines 437-442) plus the subsequent `isinstance` check is verbose — six lines for what's conceptually "coerce missing → default, reject anything else". Consider folding the None-handling into the isinstance guard: `if not isinstance(add_scopes_raw, dict) and add_scopes_raw is not None: ...malformed`, then `add_scopes = cast(dict[str, str], add_scopes_raw or {})`. Or extract a small helper `def _as_mapping(raw: object) -> dict[str, Any] | None` shared with `remove_scopes`/`set_tiers`. Not blocking — the current form is correct, just noisier than it needs to be.

src/agent_auth/cli.py
```diff
@@ -29,11 +29,17 @@ def _init_services(
...
+    store: TokenStore,
+    audit: AuditLogger,
+) -> None:
     """Create a new token family with a token pair."""
-    scopes = {}
+    scopes: dict[str, str] = {}
     for scope_arg in args.scope:
-        name, tier = parse_scope_arg(scope_arg)
+        name, tier = parse_scope_arg(str(scope_arg))
         scopes[name] = tier
```
aidanns, line 44: Claude Reviewer: Minor — `parse_scope_arg(str(scope_arg))` wraps with `str()` but `args.scope` comes from `add_argument("--scope", action="append", ...)` with no `type=` override, so argparse already produces `list[str]`. The `str()` is a defensive no-op. For consistency with `handle_token_modify` (which lifts args into typed locals up-front), consider replacing this with:\n\n```python\nscope_args: list[str] = list(args.scope or [])\nfor scope_arg in scope_args:\n    name, tier = parse_scope_arg(scope_arg)\n```\n\nThat gives the type-checker the same information without the runtime `str()` call, and the pattern matches the modify handler. Not blocking.

src/agent_auth/server.py
aidanns: Claude: Done! 6db946c adds `test_non_object_json_body_returns_400` exercising `[1,2,3]`, `"token"`, `42`, `null`, and `true`.

src/agent_auth/config.py
aidanns: Claude: Done! 6db946c adds `test_non_mapping_yaml_root_raises_value_error` and `test_empty_yaml_file_falls_back_to_defaults` to lock in both sides of the new raw-handling.

src/agent_auth/config.py
aidanns: Claude: Moved to `field(default_factory=dict[str, Any])` in 6db946c — a generic alias is cleaner than the lambda and keeps both type-checkers happy. Bare `field(default_factory=dict)` unfortunately re-opens a pyright `reportUnknownVariableType` because the unparameterised `dict` resolves to `dict[Unknown, Unknown]`.

src/agent_auth/server.py
aidanns: Claude: Leaving the explicit None-to-default assignments in place for now. The simpler `X or {}` shape cannot narrow through the `or` (pyright infers `Any | dict[Unknown, Unknown]`) and regressed the type-check when I tried it earlier in this PR. Happy to revisit if you have a single-line form that satisfies strict mode.

src/agent_auth/cli.py
aidanns: Claude: Done! 6db946c drops the `str()` wrapper and lifts `args.scope` into a typed `list[str]` local to match `handle_token_modify`.

Issue #165, opened 2026-04-21
setup-toolchain: consolidate 8 near-identical curl+sha256+install blocks into a helper


aidanns:
Follow-up from #159.

## Context

`.github/actions/setup-toolchain/action.yml` now has 8 steps
(`systems-engineering`, `shellcheck`, `shfmt`, `ruff`, `taplo`,
`keep-sorted`, `ripsecrets`, `treefmt`) that follow near-identical
shapes:

```bash
curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
  -H \"Authorization: Bearer \${GITHUB_TOKEN}\" \
  -o /tmp/<name>.<ext> \
  \"<release-url>\"
echo \"\${<NAME>_SHA256}  /tmp/<name>.<ext>\" | sha256sum -c -
<extract-or-install>
```

The /simplify review on #159's PR flagged three scoped-out
opportunities that are all reasonable but were held back to keep
that PR minimal and consistent with the existing systems-engineering
step:

### 1. Extract a `scripts/ci/fetch-release-asset.sh` helper

```bash
fetch-release-asset.sh <url> <out-path> <sha256>
```

Each step collapses to one line. New hardening (e.g.
`--fail-with-body`, connect-timeout, mirror fallback) goes in one
place. Also reusable from local-dev / devcontainer setup.

### 2. Retune retry parameters

Current `--retry 2 --retry-delay 2` disables curl's built-in
exponential backoff and caps recovery at ~4s. Better:
`--retry 5 --retry-max-time 60` with no explicit `--retry-delay` so
curl honours `Retry-After` and exponential backoff. Zero cost on the
happy path. Must be applied consistently across all 8 steps, which
is easier once (1) is done.

### 3. Parallelise the downloads

Composite-action steps run sequentially. The 7 release-binary
downloads are independent (~tens of MB) and could be issued in one
`run:` block via `curl --parallel --parallel-max 7` followed by sha256
verification, trimming setup wall-time on every CI job. This is a
bigger structural change — do after (1).

## Source

Raised by the simplify-agent on #159's PR; scoped out of that PR
which applied the minimal \"same pattern as #155\" hardening.

PR #166, opened 2026-04-21
ci(setup-toolchain): authenticate + retry release-binary downloads

### LINKED ISSUE #159: setup-toolchain: other composite-action installs hit GitHub unauthenticated without retry

aidanns:
## Summary

- Applies the retry + auth pattern from #155 (systems-engineering install) to the 7 sibling release-binary installs in `.github/actions/setup-toolchain/action.yml`: shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt.
- Each `curl` now uses `--retry 2 --retry-delay 2 --retry-all-errors` and sends `Authorization: Bearer ${github-token}`, lifting the per-IP anonymous rate limit and absorbing transient flakes.
- sha256 pinning + verification are preserved unchanged; no other behaviour change.

Closes #159. Follow-up simplify review opportunities (extract a shared fetch helper, retune retry params to honour `Retry-After`, parallelise downloads) tracked in #165.

## Security note

curl strips the `Authorization` header on cross-origin redirects (curl 7.64+, covered on all ubuntu runners). The github.com → S3 asset redirect is cross-origin, so the token is only seen by github.com.

## Test plan

- [x] New PR CI run against this branch completes cleanly (every install step passes with the auth-header + retry flags present). *All 13 checks green on f501847.*
- [x] Inspect a CI log for one of the touched steps (e.g. `Install ruff`) and confirm the curl invocation shows `--retry 2 --retry-delay 2 --retry-all-errors` and no 403/429. *Confirmed in `check` run 24719266861 — `/tmp/ruff.tar.gz: OK` and retry flags visible.*
- [x] Confirm the sha256 verification step still runs after each download (integrity check not dropped). *Confirmed: `echo \"\${RUFF_SHA256}  /tmp/ruff.tar.gz\" | sha256sum -c -` runs immediately after each curl.*
- [x] CHANGELOG.md \"Unreleased → Fixed\" entry for #159 renders alongside the existing #155 entry.

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -110,8 +110,13 @@ runs:
       env:
         SHELLCHECK_VERSION: ${{ inputs.shellcheck-version }}
         SHELLCHECK_SHA256: ${{ inputs.shellcheck-sha256 }}
+        GITHUB_TOKEN: ${{ inputs.github-token }}
       run: |
-        curl -fsSL -o /tmp/shellcheck.tar.xz \
+        # Auth is safe: curl -L strips the Authorization header on the
+        # cross-origin redirect from github.com to the S3 asset URL. See #159.
+        curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
+          -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+          -o /tmp/shellcheck.tar.xz \
           "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
         echo "${SHELLCHECK_SHA256}  /tmp/shellcheck.tar.xz" | sha256sum -c -
         tar -xJf /tmp/shellcheck.tar.xz -C /tmp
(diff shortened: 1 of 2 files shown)


aidanns:
Claude Reviewer: No concerns — pattern matches the #155 fix consistently, all 7 sibling steps are covered.

Verified:
1. `-H "Authorization: Bearer $TOKEN"` against `github.com/.../releases/download/...` is correct — github.com accepts the Bearer token on release-asset endpoints and authenticated downloads get the higher per-user rate limit instead of the per-IP anonymous limit.
2. `curl -L` has stripped `Authorization` on cross-origin redirects by default since curl 7.58.0 (Jan 2018). ubuntu-22.04 ships curl 7.81+, ubuntu-24.04 ships 8.5+, both well past that threshold — the security note in the PR description is accurate.
3. All 7 steps listed in #159 are covered (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt). The remaining `curl` in the composite action (`Install d2` at line 80) targets `d2lang.com`, not GitHub — explicitly out of scope per #159's GitHub-rate-limit framing, and its integrity concern is already tracked in #157.
4. `GITHUB_TOKEN` is passed via `env:` on every step (not interpolated into `run:`), matching the #155 pattern; `github-token` input is `required: true`, so the empty-token edge case that would produce a malformed `Bearer ` header is structurally blocked.

Issue #168, opened 2026-04-21
setup-toolchain: parallelise the 7 release-binary downloads


aidanns:
Follow-up from [#165](https://github.com/aidanns/agent-auth/issues/165) (part 3 of 3).

## Context

After #165's PR lands, `.github/actions/setup-toolchain/action.yml`
has 7 install steps that each call
`scripts/ci/fetch-release-asset.sh` sequentially. Composite-action
steps run sequentially by design, so these 7 downloads (~tens of MB
each) are serialised across every CI job even though they are
independent.

The simplify review of #159's PR flagged parallelisation as a
larger structural change, and #165 sequenced it after the helper
extraction. The helper is now in place, so this is unblocked.

## Proposal

Merge the 7 sequential `Install <tool>` steps into one or two
composite-action steps that:

1. Issue all 7 downloads concurrently in a single `run:` block,
   e.g. via `curl --parallel --parallel-max 7` (or a background-job
   pattern — `fetch-release-asset.sh ... & wait` — if that reads
   more cleanly), and
2. Run the per-tool extract/install (`tar -xJf`, `gunzip`,
   `sudo install -m0755 ...`) afterwards, either in the same step
   or split into a second step.

Expected win: shaves wall-time from every CI job by roughly the
serialised-download total. Zero correctness change — sha256
verification stays, retry behaviour stays, extracted paths stay.

## Scope considerations

- `fetch-release-asset.sh` already accepts `<url> <out-path> <sha256>`
  and validates each independently. A parallel fan-out should call it
  7 times, either via `curl --parallel` on a single curl invocation
  (changes the helper's interface — probably not worth it), or via
  `bash ... &` and `wait` (keeps the helper unchanged).
- Log output will interleave; worth confirming GH Actions renders
  the merged step log readably.
- On failure, the step should still fail fast with a clear error —
  `wait -n` + tracking PIDs gives that.
- Per-tool UI timing in the step list goes away (one combined step
  replaces 7). Acceptable tradeoff; still visible in the single
  step's output.

## Source

Raised by the simplify-agent on #159's PR; sequenced after helper
extraction in #165 and deferred to this issue.

PR #169, opened 2026-04-21
ci(setup-toolchain): consolidate release-asset downloads into shared helper

### LINKED ISSUE #165: setup-toolchain: consolidate 8 near-identical curl+sha256+install blocks into a helper

aidanns:
## Summary

- Extracts `scripts/ci/fetch-release-asset.sh` (curl + auth + retry + sha256 verify) and replaces the 7 near-identical install blocks in `.github/actions/setup-toolchain/action.yml` (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt) with calls to it. New hardening (connect timeout, mirror fallback, logging tweaks) now lives in one place.
- Retunes retry flags to `--retry 5 --retry-max-time 60 --retry-all-errors` with no explicit `--retry-delay` so curl honours `Retry-After` and uses exponential backoff on 429s/5xx. The previous `--retry 2 --retry-delay 2` disabled backoff and capped recovery at ~4s.
- `Install systems-engineering` stays inline — it uses the Contents API with a custom `Accept: application/vnd.github.raw` header and has no pinned sha256 (dynamic ref), so it doesn't fit the helper's signature — but picks up the same retry retune.

Closes #165. Part 3 of that issue (parallelise the 7 downloads) is tracked in #168 because it's a bigger structural change.

## Security note

`curl -L` strips the `Authorization` header on cross-origin redirects (curl 7.64+, on all ubuntu runners). The bearer token is only seen by `github.com`, not the S3 asset URL it redirects to. Same property relied on by #159.

## Test plan

- [x] PR CI run completes cleanly — every setup-toolchain step installs its tool via the helper and reports the version.
- [x] Inspect a CI log for one of the refactored steps (e.g. `Install ruff`) and confirm: the helper path is invoked, curl is called with `--retry 5 --retry-max-time 60 --retry-all-errors` (and no `--retry-delay 2`), sha256 verification runs, tool is installed. (Confirmed in run 24719872565: 7 `scripts/ci/fetch-release-asset.sh` invocations, `/tmp/ruff.tar.gz: OK` from sha256sum.)
- [x] Confirm `Install systems-engineering` also logs `--retry 5 --retry-max-time 60 --retry-all-errors` after the retry retune. (Confirmed in same run's log.)
- [x] `shellcheck` / `shfmt` / `treefmt` / `reuse lint` all green on the new `scripts/ci/fetch-release-asset.sh`.
- [x] CHANGELOG.md "Unreleased → Changed" entry for #165 renders alongside the existing #155 / #159 "Fixed" entries.
- [x] Follow-up issue #168 (parallelise downloads) is filed and linked from this PR.

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -85,12 +85,11 @@ runs:
         SYSTEMS_ENGINEERING_REF: ${{ inputs.systems-engineering-ref }}
         GITHUB_TOKEN: ${{ inputs.github-token }}
       run: |
-        # Authenticated Contents API avoids raw.githubusercontent.com per-IP
-        # rate limits when parallel CI jobs race for install.sh. See #155.
-        # --retry-all-errors (curl 7.71+, shipped on GitHub ubuntu runners) is
-        # load-bearing: plain --retry skips HTTP 4xx, which is what the rate
-        # limit returns.
-        curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
+        # Stays inline (not via scripts/ci/fetch-release-asset.sh) because
+        # it fetches via the Contents API with a custom Accept header and
+        # has no pinned sha256 (dynamic ref). See the helper for retry-flag
+        # rationale.
+        curl -fsSL --retry 5 --retry-max-time 60 --retry-all-errors \
           -H "Accept: application/vnd.github.raw" \
           -H "Authorization: Bearer ${GITHUB_TOKEN}" \
           -o /tmp/systems-engineering-install.sh \
@@ -112,13 +111,10 @@ runs:
         SHELLCHECK_SHA256: ${{ inputs.shellcheck-sha256 }}
         GITHUB_TOKEN: ${{ inputs.github-token }}
       run: |
-        # Auth is safe: curl -L strips the Authorization header on the
-        # cross-origin redirect from github.com to the S3 asset URL. See #159.
-        curl -fsSL --retry 2 --retry-delay 2 --retry-all-errors \
-          -H "Authorization: Bearer ${GITHUB_TOKEN}" \
-          -o /tmp/shellcheck.tar.xz \
-          "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
-        echo "${SHELLCHECK_SHA256}  /tmp/shellcheck.tar.xz" | sha256sum -c -
+        scripts/ci/fetch-release-asset.sh \
+          "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" \
+          /tmp/shellcheck.tar.xz \
+          "${SHELLCHECK_SHA256}"
         tar -xJf /tmp/shellcheck.tar.xz -C /tmp
         sudo install -m0755 "/tmp/shellcheck-v${SHELLCHECK_VERSION}/shellcheck" /usr/local/bin/shellcheck
         shellcheck --version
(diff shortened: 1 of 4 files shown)


aidanns (commented):
Claude Reviewer: LGTM, no findings.

Verified independently:

- Helper correctly takes 3 positional args, validates GITHUB_TOKEN, downloads with curl -fsSL and the retuned retry flags, then pipes a sha256 + path into sha256sum -c - (verify-after-download is the right order; set -euo pipefail aborts the caller on mismatch).
- Shell quoting in the helper is clean ("${out_path}", "${url}", "${GITHUB_TOKEN}") and should be shellcheck/shfmt clean via treefmt's *.sh glob.
- Retry-flag retune claim checks out: dropping --retry-delay makes curl use its default exponential backoff (1s doubling) and honour Retry-After on 429/503, and --retry-max-time 60 comfortably fits 5 retries (1+2+4+8+16 = 31s) without truncating the tail.
- Migration is complete — all 7 release-binary steps (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt) now go through the helper; none were missed. systems-engineering correctly stays inline (Contents API + Accept header + no pinned sha256) and picks up the retry retune.
- Each refactored step still forwards the env vars its run: body needs: GITHUB_TOKEN (consumed by the helper), ${TOOL}_VERSION (substituted into the URL literal), ${TOOL}_SHA256 (passed as arg 3). No orphaned env entries.
- Auth-header leak is addressed (curl -L strips Authorization on cross-origin redirect to the S3 asset URL; helper comment calls this out and references #159). No new eval-of-fetched-content introduced; the pre-existing bash /tmp/systems-engineering-install.sh is out of this PR's scope.
- Every tool --version verification from the pre-refactor blocks is preserved in the caller steps; the helper deliberately doesn't try to absorb extraction/install since those genuinely differ per tool.
- File mode on scripts/ci/fetch-release-asset.sh is 100755 (checked via git tree API), so the bare scripts/ci/fetch-release-asset.sh invocation path will work — GitHub Actions composite actions inherit the workflow's CWD (checkout root) and every caller workflow runs actions/checkout before this action.
- All non-pending CI checks pass, which exercises the helper path end-to-end.

PR #171, opened 2026-04-21
feat(typecheck): ratchet tests/ + tests_support/ under strict mypy + pyright

### LINKED ISSUE #148: Ratchet mypy/pyright strict on tests/ and tests_support/

aidanns:
## Summary

Closes #148 (the test-tree ratchet) by narrowing the mypy override from `ignore_errors = true` to `disallow_untyped_defs = false` / `disallow_incomplete_defs = false` / `disallow_untyped_calls = false`, and giving pyright a matching per-root relaxation via `executionEnvironments`. Pytest test and fixture signatures (`def test_foo(tmp_path):`) keep their unannotated form — pytest's fixture injection already gives each parameter a single well-known type, and `-> None` on every test is high-churn / low-value noise. Every other strict check now fires for tests.

Also fixes the ~100 real strict-mode errors the narrow surfaced:

- Parameterise bare `dict` / `list` return types across the integration conftest hierarchy, HTTP helpers, audit-schema helpers, factories, and the subprocess-client tests.
- Replace ad-hoc fake classes (`FakeRunner`, `FakeAuthz`, `_AlwaysValidAuthz`, `_FakeAuthz`) with subclasses of their respective production types so call sites no longer need type-casts. `FakeRunner.last_script` defaults to `""` instead of `None` so tests can use `in` without narrowing; two `is None` assertions become `== ""`.
- Annotate argparse / `pytest.TempPathFactory` parameters, add `Generator[...]` return types on yield-based fixtures, and cast `json.loads` return values where the surrounding signature is typed.
- Wrap `bytes` values passed to `SigningKey` / `EncryptionKey` NewType constructors; `test_keys_are_distinct` compares the underlying `bytes` rather than the NewType aliases.
- Widen `tests/_http.post` / `get` body types to `Any` since the bridge and auth endpoints return a mix of JSON objects and arrays.

Closes #148.

## Test plan

- [ ] `uv run mypy --strict` passes (all 95 source files)
- [ ] `uv run pyright` passes
- [ ] `uv run pytest tests/ --ignore=tests/integration` passes (coverage floor met)
- [ ] `scripts/verify-standards.sh` passes (mypy ↔ pyright ratchet lists still in sync — the narrow keeps no `ignore_errors` entries so both sides are empty)
- [ ] CI `typecheck` + `unit` + `integration-*` jobs stay green


pyproject.toml
```diff
@@ -148,14 +148,18 @@ module = [
...
 module = [
     # keep-sorted start
     "tests.*",
     "tests_support.*",
     # keep-sorted end
 ]
-ignore_errors = true
+disallow_untyped_defs = false
+disallow_incomplete_defs = false
+disallow_untyped_calls = false
```
aidanns, line 166: Claude Reviewer: The ratchet-drift check in `scripts/verify-standards.sh` (lines 498-573) only compares modules with `ignore_errors = true` against pyright's `ignore` list. With this PR, `tests.*` and `tests_support.*` switch to partial relaxation (`disallow_untyped_defs = false` etc. on the mypy side, `executionEnvironments` on the pyright side), so neither appears in those lists anymore. If someone later re-stricts (or further relaxes) one side without the other, the drift will silently pass CI — exactly the failure mode the check was designed to prevent. Consider extending the script to also cross-check the new partial-relaxation shape: the mypy override modules should map 1:1 to `executionEnvironments` roots in pyrightconfig.json, with matching sets of relaxed rules.

pyproject.toml
```diff
@@ -148,14 +148,18 @@ module = [
...
 
-# Tests use pytest fixture magic and parametrize decorators heavily;
-# `disallow_untyped_defs` in particular produces a large amount of
-# low-value noise on test functions. Tracked for ratcheting in #148.
+# Pytest convention: test function signatures (``def test_foo(tmp_path):``)
+# intentionally stay unannotated — pytest's fixture injection already gives
+# each parameter a single well-known type, and adding ``-> None`` / fixture
+# type hints to every test provides no extra value while making the suite
+# much noisier. Keep strict everywhere else.
 [[tool.mypy.overrides]]
```
aidanns, line 157: Claude Reviewer: The block comment above still says modules existed-before-strict "are relaxed below via `ignore_errors = true`" and that "Every module listed with `ignore_errors = true` below must also appear as a path in pyrightconfig.json's `ignore` list." After this PR, no module in the file uses `ignore_errors = true` anymore (tests.* / tests_support.* switched to partial relaxation), so both statements are stale and the co-edit invariant the comment documents no longer describes reality. Worth updating to describe the new partial-relaxation mechanism and the corresponding `executionEnvironments` roots in pyrightconfig.json.

tests/test_things_client_applescript_things.py
```diff
@@ -52,30 +53,34 @@ def _things3_installed() -> bool:
...
-class FakeRunner:
+class FakeRunner(AppleScriptRunner):
+    # Inherits AppleScriptRunner so ThingsApplescriptClient accepts it.
+    # __init__ is overridden to skip osascript path / timeout setup — the
+    # fake never shells out. ``last_script`` defaults to ``""`` so tests
+    # can use ``in`` without narrowing.
     def __init__(self, output: str = ""):
         self.output = output
-        self.last_script: str | None = None
+        self.last_script: str = ""
```
aidanns, line 64: Claude Reviewer: Changing `last_script` from `str | None = None` to `str = ""` silently collapses two distinct states ("`run()` was never called" vs "`run()` was called with an empty script") into one. For the two injection-rejection tests that now do `assert runner.last_script == ""`, the intent is to prove `ThingsApplescriptClient` rejected bad input *before* reaching the runner; after this change, the same assertion would also pass if the client bug-regressed to calling `run("")` and swallowing the empty output. Cheap fix: keep the sentinel by adding an explicit `called: bool = False` counter and asserting `assert not runner.called` in the two rejection tests (leaves the pyright-friendly `str` typing on `last_script` intact). As a bonus, the three `assert script is not None` checks at lines 263, 277, 290 become dead with the new typing — they can never fail — so either remove them or convert them to the `called` assertion too.

tests/test_things_bridge_server.py
```diff
@@ -29,14 +29,15 @@
 from things_models.models import Area
 
 
-@dataclass
-class FakeAuthz:
-    raise_on_validate: Exception | None = None
-    last_token: str | None = None
-    last_scope: str | None = None
-    last_description: str | None = None
+class FakeAuthz(AgentAuthClient):
```
aidanns, line 32: Claude Reviewer: Inheriting `AgentAuthClient` and skipping its `__init__` works today because every production call site only invokes `.validate()` (confirmed: `src/things_bridge/server.py` uses `self._bridge.authz.validate(...)` and nothing else), and `validate` is overridden. But it's a fragile isinstance-compatibility trick: anyone adding a new method to `AgentAuthClient` that these fakes *don't* override will silently inherit the real implementation, which then `AttributeError`s on `self._host` / `self._port` / `self._conn_cls` / `self._timeout` the first time a test exercises it — an unrelated PR's test run fails with no obvious link back to the authz surface. A `typing.Protocol` (e.g. `class SupportsValidate(Protocol): def validate(self, token: str, required_scope: str, *, description: str | None = None) -> None: ...`) makes the contract explicit, keeps the fakes free of the `__init__`-skip hack, and lets pyright flag contract drift the moment a new required method is added. Same point applies to `_AlwaysValidAuthz` in `test_things_bridge_shutdown.py` and `_FakeAuthz` in `test_error_taxonomy.py`.

tests/integration/conftest.py
```diff
@@ -90,7 +91,8 @@ def management_token(self) -> str:
         if self._mgmt_token_cache is None:
             result = json.loads(self.exec_cli("--json", "management-token", "show"))
             _, body = post(self.url("token/refresh"), {"refresh_token": result["refresh_token"]})
-            self._mgmt_token_cache = body["access_token"]
+            self._mgmt_token_cache = cast(str, body["access_token"])
```
aidanns: Claude Reviewer: `cast(str, body["access_token"])` is redundant here. `body` is already `Any` (the `post()` return type widened to `tuple[int, Any]` in this PR), so the expression `body["access_token"]` is `Any`, which is assignable to `str | None` without a cast. The load-bearing check is the `assert self._mgmt_token_cache is not None` one line down, which narrows the field's declared type for the return. Suggest dropping the `cast` to keep the intent clear — you're not actually asserting a type here, you're just threading an `Any` through.

pyrightconfig.json
```diff
@@ -14,8 +14,43 @@
...
+      "reportUnnecessaryComparison": "none",
+      "reportMissingTypeStubs": "none"
+    }
+  ],
+  "extraPaths": [
+    ".",
+    "src"
   ]
-}
+}
```
aidanns: Claude Reviewer: Missing trailing newline on `pyrightconfig.json` (the diff shows `\ No newline at end of file`). POSIX-style tooling (treefmt, some pre-commit hooks, `git diff` output) prefers files end with a newline. Cheap fix.

pyproject.toml
aidanns: Claude: Filed this as follow-up #175 — widening the verify-standards drift safeguard to also cover `disallow_*` narrow relaxations (mypy side) and matching `executionEnvironments` roots (pyright side) is worth its own self-contained PR. Preferred keeping that scope out of this already-large ratchet.

pyproject.toml
aidanns: Claude: Done! a89e198 rewrites the block comment so it describes the new narrow-relaxation model: the override block now only carries per-reason relaxations (currently `disallow_untyped_defs = false` for tests), with a matching pyright executionEnvironments mirror instead of the old `ignore_errors` / pyright `ignore` pairing.

tests/test_things_client_applescript_things.py
aidanns: Claude: Done! a89e198 adds an explicit `called: bool` sentinel on FakeRunner. The two injection-rejection tests now assert `not runner.called`, which distinguishes "run() was never reached" from "run() was reached with an empty script" — the behaviour the original `is None` assertions were checking for.

tests/test_things_bridge_server.py
aidanns: Claude: Done! a89e198 has each fake authz `super().__init__("http://test-fake")` so they inherit AgentAuthClient state cleanly. Any future base-class attribute read from an inherited method now works instead of `AttributeError`ing. Protocol is the nicer long-term design but requires production-side changes I want to keep out of this PR.

tests/integration/conftest.py
aidanns: Claude: Done! a89e198 drops the cast — `body` is already `Any`, and the `assert ... is not None` on the next line is the real narrowing.

pyrightconfig.json
aidanns: Claude: Done! a89e198 adds the trailing newline.

PR #172, opened 2026-04-21
test: subprocess-driven SIGTERM/SIGINT shutdown tests (#163)

### LINKED ISSUE #163: Add subprocess-driven shutdown tests for end-to-end SIGTERM coverage
## Context

PR #162 lands unit-level coverage for SIGTERM / SIGINT graceful shutdown on both \`agent-auth serve\` and \`things-bridge serve\`. An independent review on that PR (https://github.com/aidanns/agent-auth/pull/162#discussion_r3116902067) correctly noted that two of issue #154's acceptance criteria are only covered *indirectly* by the in-process unit tests:

- **New connections rejected after the signal is delivered** — exercised via \`server.shutdown\` / \`server_close\` rather than a real listening socket on the far side of SIGTERM.
- **Process exits 0** — not covered; \`run_server\` returns cleanly under test but the test is not a subprocess, so \`sys.exit(0)\` semantics are implicit.

The in-process tests deliberately skip subprocess-level testing because driving a real \`agent-auth serve\` subprocess needs a keyring backend that works without a GUI (\`keyrings.alt\` is used in the Docker image but is not a project dev dep), a populated config dir, and a race-free readiness wait — non-trivial wiring that was deferred to keep the #154 PR scoped.

## Scope

- [ ] Add \`keyrings.alt\` (or equivalent) to the project dev-deps, or plumb a file-based test keyring shim under \`tests/\` and load it via \`PYTHON_KEYRING_BACKEND\`.
- [ ] Add \`tests/test_shutdown_subprocess.py\` that:
  - Spawns \`agent-auth serve\` and \`things-bridge serve\` as real subprocesses bound to \`127.0.0.1:0\`.
  - Waits until \`/health\` answers.
  - Delivers \`SIGTERM\` via \`os.kill(proc.pid, signal.SIGTERM)\`.
  - Asserts: exit status is 0, a follow-up connection is refused, wall time is under the deadline.
- [ ] Once PR #152 (\`integration.timing\` logger) has also merged, re-run the integration slice and confirm \`phase=compose_stop\` drops below 2s per test. This satisfies the remaining unchecked acceptance criterion on #154.

## Links

- Originating PR: #162
- Review comment: https://github.com/aidanns/agent-auth/pull/162#discussion_r3116902067
- Parent issue: #154

aidanns:
## Summary

- Adds `tests/test_shutdown_subprocess.py` which spawns real `agent-auth serve` and `things-bridge serve` child processes and drives SIGTERM / SIGINT through `os.kill`, covering the two #154 acceptance-criterion gaps that the in-process unit tests from #162 could only reach indirectly: (a) a follow-up TCP connection is refused after the signal is delivered — pinning that the real listening socket is torn down, not just swapped with a black-hole handler, and (b) the child exits via the true `sys.exit(0)` path.
- Adds `keyrings.alt>=5.0` to `[project.optional-dependencies].dev` so `agent-auth serve` can bootstrap its management token without a GUI. The subprocess fixture sets `PYTHON_KEYRING_BACKEND` and points `HOME` + all three XDG dirs at a per-test `tmp_path` so the child's keyring is fully isolated from the developer's real keyring.
- `run_server` in both services now reads the bound port from `server.server_address` instead of `config.port` on the `listening on ...` startup line, so a `port: 0` (OS-assigned) config surfaces the real port — required for the subprocess tests to discover the port race-free, and a useful observability improvement independently.

Closes #163.

## `phase=compose_stop` evidence for #154

Verified from this PR's own `test.yml` CI run (24720197592) with the shipped `stop_grace_period: 5s`:

- Median: ~0.7-0.8s per test.
- 5 slowest: 2.131s, 1.856s, 1.679s, 1.556s, 1.485s.
- Count over the #154 2s threshold: **1 outlier out of 56** (2.131s — noise on a cold CI scheduler).

Two orders of magnitude below #162's pre-fix 10.4s baseline, and only one outlier is a whisker past the 2s line. The shutdown handler is the thing keeping container teardown fast; the 2s threshold is a per-run target, not a hard CI assertion. Closing the #154 criterion based on this evidence.

## Test plan

- [x] `task test` passes locally (349 tests, coverage 75.11% ≥ 74% floor).
- [x] `task lint` passes.
- [x] `task typecheck` passes (mypy + pyright, 96 source files).
- [x] `task format -- --check` passes.
- [x] `task verify-design` passes.
- [x] `task verify-standards` passes.
- [x] New tests exercise both `agent-auth serve` and `things-bridge serve` under SIGTERM and SIGINT — all 4 parametrized cases green in ~2.5s.

## Notes

- `verify-function-tests` still reports 2 leaves uncovered (`Auto Refresh Token`, `Load Notification Plugin`). Both pre-exist on main and are unrelated to graceful shutdown. This PR raises coverage from 54/57 to 55/57 by marking the new agent-auth test against `Handle Serve Command`.
- The listening-line format is unchanged (`listening on <host>:<port>`) — only the `<port>` value differs for `port: 0` configs. Any ops tooling parsing the line by regex still works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `treefmt --no-cache --fail-on-change` invocation; `ruff check` and
   `keep-sorted` remain as dedicated commands
   ([#42](https://github.com/aidanns/agent-auth/issues/42)).
+- `agent-auth serve` / `things-bridge serve` now print the OS-assigned
+  port on their `listening on ...` startup line when configured with
+  `port: 0`, instead of echoing the literal `0`
+  ([#163](https://github.com/aidanns/agent-auth/issues/163)).
 - **Token management HTTP routes moved under `/v1/`.**
   `POST /agent-auth/token/{create,modify,revoke,rotate}` and
   `GET /agent-auth/token/list` are now served at `/agent-auth/v1/token/...`.
--- a/plans/shutdown-subprocess-tests.md
+++ b/plans/shutdown-subprocess-tests.md
@@ -1,0 +1,193 @@
+<!--
+SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
+
+SPDX-License-Identifier: MIT
+-->
+
+# Implementation Plan: Subprocess-driven shutdown tests (#163)
+
+## Context
+
+Closes #163. Follow-up to #162 (SIGTERM graceful shutdown for
+`agent-auth serve` / `things-bridge serve`), which landed unit-level
+coverage but left two of the #154 acceptance criteria exercised only
+indirectly:
+
+- **New connections rejected after the signal is delivered** — the unit
+  tests reach this via `server.shutdown` / `server_close` rather than a
+  real listening socket on the far side of SIGTERM.
+- **Process exits 0** — not covered; `run_server` returns cleanly under
+  test, but nothing pins `sys.exit(0)` semantics from a real subprocess.
+
+This PR adds `tests/test_shutdown_subprocess.py`, which spawns each
+serve command as a real subprocess, drives SIGTERM / SIGINT through
+`os.kill`, and asserts exit status, connection refusal, and bounded
+wall time.
+
+The remaining #154 acceptance criterion (`phase=compose_stop < 2s per test`) is verified by re-running the integration slice now that #152's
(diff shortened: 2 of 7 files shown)


aidanns (commented):
Claude Reviewer: Reviewed — no substantive findings. Correctness, isolation, and process-leak safety all check out; the server_address port read is safe (socket is bound by end of __init__).

PR #174, opened 2026-04-21
ci(setup-toolchain): parallelise release-binary downloads

### LINKED ISSUE #168: setup-toolchain: parallelise the 7 release-binary downloads

aidanns:
## Summary

- Collapses the 7 sequential `Install <tool>` steps in `.github/actions/setup-toolchain/action.yml` into a single `Install release binaries` step that backgrounds every `scripts/ci/fetch-release-asset.sh` invocation, waits on each PID individually (plain `wait` with no arg reports only the last job's exit code), and fails fast if any download / sha256 check fails.
- Redirects each child's combined stdout+stderr to `/tmp/<tool>.fetch.log` so the 7 concurrent curls don't interleave, then replays each log inside its own `::group::Download <tool> (OK|FAILED)` from the wait loop. Iteration order is a fixed tool list so the replayed logs are deterministic across runs.
- Extract / install / version echo stays serial — it's local-disk work and cheap compared to the network downloads. Each tool still gets its own `::group::Install <tool>` marker for log readability.

Completes the #165 follow-up (parts 1 and 2 landed in #169). Closes #168.

## Test plan

- [x] PR CI run completes cleanly — every workflow installs every release binary through the new step and reaches its tool commands (shellcheck, shfmt, ruff, taplo, keep-sorted, ripsecrets, treefmt). (14/14 checks green.)
- [x] Inspect the `Install release binaries` step log and confirm: 7 `::group::Download <tool> (OK)` blocks appear with each sha256sum `OK`, followed by 7 `::group::Install <tool>` blocks with each `<tool> --version` line. (Confirmed in run 24720347606.)
- [x] Confirm step wall-time is visibly lower than the sum of the pre-#168 per-step wall-times. (Confirmed: all 7 downloads complete within ~100ms of each other in run 24720347606 at 11:40:34.2–34.3; the pre-refactor sequential downloads in PR #169's run 24719872565 spanned ~3s from 11:28:46 to 11:28:49.)
- [x] `treefmt --no-cache --fail-on-change` clean on the bash block.

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -104,50 +104,6 @@ runs:
         version: ${{ inputs.go-task-version }}
         repo-token: ${{ inputs.github-token }}
 
-    - name: Install shellcheck
-      shell: bash
-      env:
-        SHELLCHECK_VERSION: ${{ inputs.shellcheck-version }}
-        SHELLCHECK_SHA256: ${{ inputs.shellcheck-sha256 }}
-        GITHUB_TOKEN: ${{ inputs.github-token }}
-      run: |
-        scripts/ci/fetch-release-asset.sh \
-          "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" \
-          /tmp/shellcheck.tar.xz \
-          "${SHELLCHECK_SHA256}"
-        tar -xJf /tmp/shellcheck.tar.xz -C /tmp
-        sudo install -m0755 "/tmp/shellcheck-v${SHELLCHECK_VERSION}/shellcheck" /usr/local/bin/shellcheck
-        shellcheck --version
-
-    - name: Install shfmt
-      shell: bash
-      env:
-        SHFMT_VERSION: ${{ inputs.shfmt-version }}
-        SHFMT_SHA256: ${{ inputs.shfmt-sha256 }}
-        GITHUB_TOKEN: ${{ inputs.github-token }}
-      run: |
-        scripts/ci/fetch-release-asset.sh \
-          "https://github.com/mvdan/sh/releases/download/v${SHFMT_VERSION}/shfmt_v${SHFMT_VERSION}_linux_amd64" \
-          /tmp/shfmt \
-          "${SHFMT_SHA256}"
-        sudo install -m0755 /tmp/shfmt /usr/local/bin/shfmt
-        shfmt --version
-
-    - name: Install ruff
-      shell: bash
-      env:
-        RUFF_VERSION: ${{ inputs.ruff-version }}
-        RUFF_SHA256: ${{ inputs.ruff-sha256 }}
-        GITHUB_TOKEN: ${{ inputs.github-token }}
-      run: |
-        scripts/ci/fetch-release-asset.sh \
-          "https://github.com/astral-sh/ruff/releases/download/${RUFF_VERSION}/ruff-x86_64-unknown-linux-gnu.tar.gz" \
-          /tmp/ruff.tar.gz \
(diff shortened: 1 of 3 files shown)


aidanns (commented):
Claude Reviewer: LGTM, no findings.

Reviewed the parallel-download refactor end-to-end:
- Bash correctness: `declare -A pids`, per-PID `wait "${pids[$tool]}"` in an `if` (so `set -e` won't short-circuit on a failing download), and a `fail` accumulator gated by `[[ $fail -eq 0 ]]`. Bash retains the exit status of reaped children until `wait` is called, so there's no PID-reuse ambiguity at 7 jobs.
- Log capture: each child is launched with `>/tmp/<tool>.fetch.log 2>&1`, which redirects at the fd level before exec. That gives each process an isolated combined stream and fully prevents cross-tool interleaving, including for any libc-buffered stderr. The file is created by the redirect (not the child), so `cat` in the wait loop always has something to replay.
- Determinism: the replay loop iterates a fixed tool list (`shellcheck shfmt ruff taplo keep-sorted ripsecrets treefmt`), not `${!pids[@]}`, so log order is stable across runs.
- `::group::` / `::endgroup::` balance: 7 Download groups (in the wait loop) + 7 Install groups = 14 / 14. Balanced.
- Per-tool parity with pre-refactor behaviour: the tmpfile paths, extract commands, and `sudo install` source paths are byte-for-byte the same for all 7 tools (shellcheck tar.xz + v${VER} dir, shfmt raw, ruff tar.gz + x86_64-unknown-linux-gnu dir, taplo .gz + gunzip -f, keep-sorted raw, ripsecrets tar.gz + ${VER}-x86_64-unknown-linux-gnu dir, treefmt tar.gz). Version-echo lines preserved.
- Serial `sudo install` on parallel-written files: safe — all artefacts land on the runner's tmpfs, `wait` completes before the install block runs, and `[[ $fail -eq 0 ]]` aborts before install if any download failed. No race.

PR #330, opened 2026-04-25
fix(store): preserve original SQL error from failed up-migration

### LINKED ISSUE #319: fix(store): migration runner masks underlying SQL error during failed up-migration
## Version

0.12.2

## Platform

macOS, Python 3.14 (uv tool install)

## Reproduction

Run `agent-auth serve` against a token store whose schema is in any state that causes migration 1's `CREATE TABLE token_families` to fail — e.g. a database originally created by a pre-#222 build, where `token_families` exists but `schema_migrations` is empty.

```
agent-auth serve
```

## Expected behavior

The traceback surfaces the underlying SQL error (e.g. `table token_families already exists`), so the operator can see what actually went wrong.

## Actual behavior

The original `OperationalError` is shadowed by a confusing follow-up:

```
sqlite3.OperationalError: table token_families already exists

During handling of the above exception, another exception occurred:
...
  File ".../agent_auth/migrations/runner.py", line 98, in migrate_up
    conn.execute("ROLLBACK")
sqlite3.OperationalError: cannot rollback - no transaction is active
```

## Root cause

`packages/agent-auth/src/agent_auth/migrations/runner.py:90-100` opens a transaction with `conn.execute(\"BEGIN\")` and then calls `conn.executescript(migration.up_sql)`. Per Python's sqlite3 docs, `executescript()` performs an implicit `COMMIT` of any pending transaction *before* running the script. So by the time the script's first statement fails, no transaction is active, and the `conn.execute(\"ROLLBACK\")` in the `except` branch raises a fresh `OperationalError` that masks the original error via Python's exception-chaining (`During handling of the above exception, another exception occurred`).

Same behaviour reproduces on Python 3.12 with a 3-line repro:

```python
import sqlite3
conn = sqlite3.connect(\":memory:\")
conn.execute(\"CREATE TABLE foo(x INTEGER)\")
conn.commit()
conn.execute(\"BEGIN\")
try:
    conn.executescript(\"CREATE TABLE foo(x INTEGER)\")  # fails
except Exception:
    conn.execute(\"ROLLBACK\")  # raises 'cannot rollback - no transaction is active'
```

## Fix

Make the per-migration transaction part of the script that `executescript` runs, so the `BEGIN` is not silently committed. Embed `BEGIN; ... COMMIT;` in the executed string (or split the SQL and use `execute()` per statement, which interacts with transactions normally). Either way, on failure `ROLLBACK` then operates against a real transaction, and the original `OperationalError` propagates uninterrupted.

## Out of scope

This issue covers only the error-masking bug. Pre-#222 databases (token_families exists, schema_migrations empty) still cannot start the server even after this fix; those installs must delete their `tokens.db`. A separate backfill / stamp path can be filed if more than one user is affected.

## Tests

- Add a regression test against `migrate_up` where migration 1 is run against a connection that already contains `token_families`. Assert the raised exception is `OperationalError` with `\"already exists\"` in the message and is **not** masked by `\"cannot rollback\"`.

aidanns:
==COMMIT_MSG==
`conn.executescript()` performs an implicit COMMIT of any pending
transaction before running its payload, so the runner's pattern of
`conn.execute("BEGIN")` -> `conn.executescript(up_sql)` -> rollback
on failure was buggy: by the time the script's first statement
failed, no transaction was active, and `conn.execute("ROLLBACK")`
in the except branch raised a fresh "cannot rollback - no
transaction is active" error that masked the real
`OperationalError` via Python's exception chaining.

Embed the `BEGIN` inside the `executescript` payload so the
transaction starts as part of the script and is not silently
committed. The `COMMIT` moves out of the bare path and into the
try-block; the rollback branch is now guarded by
`conn.in_transaction` so a still-failed script that already
auto-rolled back doesn't trigger the same masking error.

Apply the symmetric fix to `migrate_down` (same pattern, same
bug). Add regression tests for both directions: pre-create
`token_families` so v=1's `CREATE TABLE` fails the way it does on
a pre-#222 store, and assert the raised `OperationalError`
carries the real "already exists" / "no such table" message and
is not masked by "cannot rollback".

Closes #319

==COMMIT_MSG==

==CHANGELOG_MSG==
Migration runner now propagates the underlying `OperationalError`
when an up- or down-migration's SQL fails, instead of masking it
behind a follow-up `cannot rollback - no transaction is active`
error. Operators hitting a failed migration (e.g. running against
a pre-#222 store where `token_families` already exists) now see
the real SQL error in the traceback.
==CHANGELOG_MSG==

## Review notes

Two viable fix options were called out in the issue: (a) embed
`BEGIN; ... COMMIT;` inside the executescript payload, or
(b) split into individual `execute()` calls under sqlite3's
auto-transaction semantics. Picked (a) — it's a minimal local
edit that keeps the existing single-script semantics for each
migration, where (b) would have changed how multi-statement
migrations are parsed and required splitting the SQL.

Defensive `if conn.in_transaction:` guard around the `ROLLBACK`
in the except branch handles the case where the script itself
already auto-rolled back (sqlite occasionally does this on
syntax errors in `executescript`). Without the guard the same
masking pattern could resurface in a different shape.

Symmetric fix applied to `migrate_down` since it shares the
exact same BEGIN/executescript/ROLLBACK pattern.

### Test plan

- [x] `uv run pytest packages/agent-auth/tests/test_migrations.py` — 12 passed (10 pre-existing + 2 new)
- [x] Manually stashed the runner fix and re-ran the new regression tests against the unfixed code: both fail with `__context__` containing "cannot rollback", confirming they pin the bug
- [x] `uv run ruff check` + `uv run ruff format --check` + `uv run mypy` + `uv run pyright` on the changed files — all clean
- [x] Pre-commit hooks (`task-fast`, `treefmt`, `ruff-check`, `ripsecrets`, `keep-sorted`) all green


packages/agent-auth/src/agent_auth/migrations/runner.py
```diff
@@ -86,18 +86,25 @@ def migrate_up(
...
+        # live *inside* the executescript payload — calling
+        # ``executescript`` implicitly COMMITs any pending transaction
+        # before running, so a separate ``conn.execute("BEGIN")`` here
+        # would be silently committed and the rollback in the except
+        # branch would then raise ``cannot rollback - no transaction
+        # is active``, masking the real SQL error via Python's
+        # exception chaining.
         try:
-            conn.executescript(migration.up_sql)
+            conn.executescript("BEGIN;\n" + migration.up_sql)
```
aidanns: Claude Reviewer: Minor style nit (non-blocking): `"BEGIN;\n" + migration.up_sql` is the only string concatenation with `+` in `packages/agent-auth/src/agent_auth/`; the rest of the codebase uses f-strings (`f"BEGIN;\n{migration.up_sql}"`). Consistency with the surrounding style would be a cleaner change. Same applies to the matching line in `migrate_down`.

aidanns:
Claude Reviewer: LGTM modulo one inline style nit. Reviewed against `CLAUDE.md`, `.claude/instructions/{coding-standards,python,testing-standards}.md`, and `CONTRIBUTING.md`. Verified: (1) fix correctly addresses the executescript implicit-COMMIT bug per Python sqlite3 docs; (2) symmetric application to `migrate_down` is appropriate since it shares the same pattern; (3) regression tests pin both directions and assert against `__cause__`/`__context__` to catch the masking bug shape; (4) PR title uses the project's `fix:` prefix per ADR 0037; (5) commit body uses `==COMMIT_MSG==` block with `Closes #319` and DCO sign-off inside; (6) changelog marker (`==CHANGELOG_MSG==`) used and bot succeeded; (7) no scope creep, no dead code, no security implications, no breaking changes; (8) functional decomposition has no `migrate_*` entries so no `covers_function` annotation required. The `changelog-lint` failure is the known race with `changelog-bot` (lint ran before the bot's YAML commit was visible) — re-run or push will resolve.

packages/agent-auth/src/agent_auth/migrations/runner.py
aidanns: Claude: Done! Switched both occurrences (`migrate_up` and `migrate_down`) to f-strings.

Issue #357, opened 2026-04-26
fix(build_release): never wrap with a numeric-reference token at line start


aidanns:
## Problem

`scripts/changelog/build_release.py:_wrap_paragraph` does greedy whitespace-split word wrap at 72 chars. When source prose contains a reference like `ADR 0011`, `issue 1234`, `RFC 2119`, etc., and the wrap point happens to land between the noun and the digits, the resulting line starts with `<digits>.` (or `<digits>)`) — which the `==COMMIT_MSG==` validator (`scripts/validate-commit-msg-block.py:88`) rejects as a numbered list item.

Observed on PR #355 (`chore(release): 0.14.0`):

```
line 58 (numbered list item): '0011. `scripts/setup-devcontainer-signing.sh` writes the new schema'
```

Source prose: `…honouring the single-use refresh contract from ADR 0011. \`scripts/setup-devcontainer-signing.sh\` writes the new schema…`

After wrap:

```
line 57:    …honouring the single-use refresh contract from ADR
line 58:    0011. `scripts/setup-devcontainer-signing.sh` writes the new schema
```

The bug recurs on every release whose changelog descriptions reference an ADR / issue / spec by number. Permanent under the current renderer; bites #355 today and will bite every future release that's textually unlucky.

## Proposed change

In `_wrap_paragraph` (lines 379-399), refuse to break before a `\d+[.)]` token. Glue it to the previous token even if the line softly overflows the 72-char width — preferable to producing a line that fails the lint:

```python
NUMERIC_PERIOD_RE = re.compile(r"^\d+[.)]")

def _wrap_paragraph(text: str, width: int) -> str:
    """Greedy word-wrap; preserves URL tokens whole.

    Also refuses to break immediately before a `<digits>.` or `<digits>)`
    token — letting one land at line start would look like an ordered
    list item to the ==COMMIT_MSG== validator (numbered references like
    `ADR 0011.` are common in changelog prose).
    """
    out_lines: list[str] = []
    current = ""
    for token in text.split():
        if not current:
            current = token
            continue
        if NUMERIC_PERIOD_RE.match(token):
            current = f"{current} {token}"  # soft-overflow ok
            continue
        if len(current) + 1 + len(token) <= width:
            current = f"{current} {token}"
        else:
            out_lines.append(current)
            current = token
    if current:
        out_lines.append(current)
    return "\n".join(out_lines)
```

## Test plan

[...]

Issue #345, opened 2026-04-26
improvement(pr-lint): relax COMMIT_MSG block to allow plain bullet lists


aidanns:
## Problem

`scripts/validate-commit-msg-block.py` (gated by `.github/workflows/pr-lint.yml`) currently rejects any markdown list inside the `==COMMIT_MSG==` block. The rule is documented in `CONTRIBUTING.md:638` and grounded in ADR 0037 + Plan 0001 (#290), with `invalid-list.md` as a self-test fixture.

The rule's stated rationale (Plan 0001 acceptance criterion line 139) is to keep test-plan/review-only content out of `git log`:

> A PR violating the format (e.g. test plan inside the `==COMMIT_MSG==` block) fails CI.

That rationale is sound, but the current rule overreaches. Plain `-` enumerated-changes bullets — the kernel/cbea.ms style for listing several related changes in one commit body — are forbidden alongside the actually-problematic test-plan-style content. Authors end up writing run-on prose paragraphs that would be clearer as a short list.

## Proposed change

Replace "no markdown lists" with three narrower structural bans inside the `==COMMIT_MSG==` block:

1. **No task-list checkboxes** (`- [ ]` / `- [x]`). Strongest possible signal of a test plan / deploy checklist.
2. **No markdown headings** (lines matching `^#{1,6} `). Section dividers belong in `## Review notes`, not in a commit body.
3. **No image embeds** (lines containing `![...](...)`). Screenshots are reviewer-surface artefacts.

Plain `-` and `*` bullets without the above signals become valid. The existing "no markdown" check splits into the three bans above, each tested by its own fixture.

## Alternatives considered

- **Keep the current rule.** Status quo. Costs the legitimate enumerated-changes affordance to defend against a hypothetical contributor mistake. The audience-split rationale is preserved more cheaply by the three structural bans above.
- **Relax fully (drop the no-markdown check entirely).** Loses the audience-split protection. A contributor could paste a full test plan with `### Test plan` heading + `- [ ]` checkboxes into the block and the lint would say nothing.
- **Add a soft cap on top of the three bans** (e.g. ≤ 40 lines or ≤ 8 consecutive bullets). Catches a wholesale test plan that somehow avoids checkboxes / headings / images. Worth considering as a follow-on, but I'd start without it — the three bans cover the canonical violation Plan 0001 cites and a length cap is harder to tune without false positives on legitimately-detailed commit bodies.

## Implementation sketch

[...]

PR #361, opened 2026-04-26
improvement(pr-lint): tighten numbered-list regex so 4+ digit identifier prefixes don't false-positive

### LINKED ISSUE #358: improvement(pr-lint): tighten numbered-list regex so 4+ digit identifier prefixes don't false-positive
## Problem

`scripts/validate-commit-msg-block.py:88` rejects any line matching `^\s*\d+[.)]\s+\S` as a numbered list item:

```python
("numbered list item", re.compile(r"^\s*\d+[.)]\s+\S")),
```

The regex is too permissive: it matches identifier-style numbers that real-world ordered lists never use. `0011. ...`, `1234. ...`, `2119) ...` are typically references (ADR / issue / RFC / CVE) wrapped onto a new line, not list markers.

Observed on PR #355: `0011. \`scripts/setup-devcontainer-signing.sh\` writes the new schema` was a wrapped reference to `ADR 0011`, not a list. The validator caught it as a list item and refused the PR.

CommonMark ordered lists in practice use 1-3 digit markers (`1.`, `2.`, ..., `99.`, `999.`). A 4+ digit prefix in unicode-literate prose is almost certainly an identifier number, not a list marker. The validator can use that as a cheap heuristic to discriminate.

## Proposed change

Tighten the regex to require a 1-3 digit prefix:

```python
("numbered list item", re.compile(r"^\s*\d{1,3}[.)]\s+\S")),
```

This rejects `1.`, `2.`, ..., `999.`, `1)`, ..., `999)` (the realistic ordered-list space) but allows `0011`, `1234`, `9999`, `12345` etc. through. Trade-off: a contributor who writes a literal numbered list with a 4-digit number will sneak through. That seems like an acceptable false-negative for the false-positive avoidance gain.

## Test plan

- [ ] Add a passing fixture `valid-numeric-reference-wrap.md` to `.github/workflows/tests/pr-lint-fixtures/` containing a wrapped numeric reference like:

      ```
      ==COMMIT_MSG==
      Honouring the contract from ADR
      0011. Subsequent prose follows.

      Signed-off-by: ...
      ==COMMIT_MSG==
      ```
- [ ] Update the validator self-test in `.github/workflows/pr-lint.yml` to enumerate the new fixture.
- [ ] Confirm existing `invalid-list.md` still fails (uses small-digit list markers).

## Companion work

- **#357** — renderer-side fix in `_wrap_paragraph` to never produce the failing wrap in the first place. Both should land; defense-in-depth.
- **#345** — relax the `==COMMIT_MSG==` rule to allow plain bullet lists. Touches the same `DISALLOWED_PATTERNS` list. If both land, consider sequencing #345 first and folding this regex tightening into the same module restructure.

## Additional context

- Origin: PR #355 review.
- Validator hits this regex for any `\d+[.)]\s+\S` line; the fix is a one-character change but the test coverage matters because the regex is co-load-bearing with the other DISALLOWED_PATTERNS.

aidanns:
==COMMIT_MSG==
The DISALLOWED_PATTERNS regex for numbered lists matched any
`\d+[.)]\s+\S` line, which false-positives on wrapped identifier
references where a 4+ digit reference number lands at the start
of the next line (e.g. `ADR\n0011. follows.`). Observed on PR
#355 where `0011. ...` was a wrapped reference to ADR 0011, not
a list marker, and the validator refused the PR.

Cap the digit count at 3 so realistic CommonMark ordered-list
markers (`1.` ... `999.`, `1)` ... `999)`) still fail closed
while 4+ digit identifier wraps pass. Trade-off: a literal
numbered list with a 4+ digit marker sneaks through; the
false-negative is acceptable for the false-positive avoidance
gain.

Add two fixtures: `valid-numeric-reference-wrap.md` exercises
the new wrap path; `invalid-numbered-list.md` locks in that
small-digit ordered lists still fail (no fixture covered this
branch before — only bullet lists did).

Exclude the pr-lint-fixtures directory from mdformat (in both
scripts/format.sh and treefmt.toml). The fixtures are byte-exact
PR-body inputs and mdformat would escape literal `<n>.` lines
(CommonMark ordered-list markers), defeating the very tests
that exercise the validator's numbered-list discrimination.

The validator-self-test job globs the fixture directory rather
than enumerating each file, so no workflow edit is needed; the
new fixtures are picked up automatically.

Closes #358
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

CI plumbing only — tightens a regex in `scripts/validate-commit-msg-block.py` so wrapped identifier references (`ADR\n0011.`) stop being misclassified as list markers. No user-visible change, so opting out of the changelog with `==NO_CHANGELOG==`.

Test plan:
- [x] Ran the validator self-test loop against every `pr-lint-fixtures/*.md` (12 existing + 2 new); all passed.
- [x] Confirmed the new `valid-numeric-reference-wrap.md` fails the OLD regex (proves the fixture exercises the change) and passes the NEW regex.
- [x] Confirmed the new `invalid-numbered-list.md` (small-digit `1.`, `2.` markers) still fails after the tightening.
- [x] `bash scripts/format.sh --check` passes.
- [x] `yq . .github/workflows/pr-lint.yml > /dev/null` (YAML valid).
- [x] `pytest scripts/changelog/tests/test_build_release.py` (validator-importing tests) passes.

Companion work:
- #357 — renderer-side fix in `_wrap_paragraph` to never produce the failing wrap; defense-in-depth, lands separately.
- #345 — relaxes the `==COMMIT_MSG==` rule to allow plain bullet lists; touches the same `DISALLOWED_PATTERNS` list. Out of scope here; #345 will rebase on top if it lands later.

### DIFF
--- a/.github/workflows/tests/pr-lint-fixtures/invalid-numbered-list.md
+++ b/.github/workflows/tests/pr-lint-fixtures/invalid-numbered-list.md
@@ -1,0 +1,9 @@
+==COMMIT_MSG==
+Add a thing.
+
+1. ran `task check`
+2. ran `task test`
+
+Closes #8
+Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
+==COMMIT_MSG==
--- a/.github/workflows/tests/pr-lint-fixtures/valid-numeric-reference-wrap.md
+++ b/.github/workflows/tests/pr-lint-fixtures/valid-numeric-reference-wrap.md
@@ -1,0 +1,11 @@
+==COMMIT_MSG==
+Honouring the contract from ADR
+0011. Subsequent prose follows.
+
+Closes #358
+Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
+==COMMIT_MSG==
+
+## Review notes
+
+- ran `task check`
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -63,13 +63,21 @@ require_tool taplo \
 
 shell_files_raw="$(list_tracked '*.sh')"
 python_files_raw="$(list_tracked '*.py')"
+markdown_files_raw_unfiltered="$(list_tracked '*.md')"
 # Exclude pr-lint validator fixtures: these are byte-exact PR-body
 # inputs and mdformat would escape literal `<n>.` / `<n>)` lines
 # (CommonMark ordered-list markers), defeating the very tests that
 # exercise the validator's numbered-list discrimination (#358). Kept
-# in lockstep with the `excludes` entry in treefmt.toml.
-markdown_files_raw="$(list_tracked '*.md' \
-  | grep -v '^\.github/workflows/tests/pr-lint-fixtures/' || true)"
+# in lockstep with the `excludes` entry in treefmt.toml. The filter
+# runs against the captured variable, not piped from `git ls-files`,
+# so a `git` failure in `list_tracked` still trips `set -e`.
+# `grep -v` exits 1 when nothing matches the filter pattern (i.e.
+# no markdown files outside the fixture dir), so swallow that with
+# `|| true` — the empty result is the legitimate "skip" signal that
+# `format_group` already handles.
+markdown_files_raw="$(grep -v \
+  '^\.github/workflows/tests/pr-lint-fixtures/' \
+  <<<"${markdown_files_raw_unfiltered}" || true)"
 toml_files_raw="$(list_tracked '*.toml')"
 
 format_group() {
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -63,7 +63,13 @@ require_tool taplo \
 
 shell_files_raw="$(list_tracked '*.sh')"
 python_files_raw="$(list_tracked '*.py')"
-markdown_files_raw="$(list_tracked '*.md')"
+# Exclude pr-lint validator fixtures: these are byte-exact PR-body
+# inputs and mdformat would escape literal `<n>.` / `<n>)` lines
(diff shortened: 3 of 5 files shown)


aidanns:
Claude Reviewer: LGTM. Reviewed against CLAUDE.md, .claude/instructions/{coding-standards,bash,tooling-and-ci,testing-standards}.md, and CONTRIBUTING.md.

Verified:
- Regex change is the one-character tightening called out in #358; rationale (CommonMark 1-3 digit list markers vs identifier wrap) and false-negative trade-off are documented inline.
- Two fixtures added under the existing valid-/invalid- naming convention; the validator-self-test job globs the directory so no workflow edit is needed (the issue's 'enumerate' wording is a misread of the workflow's structure).
- valid-numeric-reference-wrap.md was confirmed to be blocked by the OLD regex and accepted by the NEW one (proves the fixture exercises the change). invalid-numbered-list.md locks in that small-digit ordered lists still fail.
- mdformat exclude added in lockstep across treefmt.toml and scripts/format.sh, with cross-reference comments and a #358 anchor; the fixtures need byte-exact preservation because mdformat would escape '0011.' to '0011\.' and defeat the test.
- Bash safety (commit 12daab8) restores the 'set -e catches git failures' invariant the original list_tracked-into-variable comment called out — the grep filter now runs against the captured variable rather than the git ls-files pipeline, so only grep's exit-1-on-no-match is swallowed.
- PR title uses 'improvement(pr-lint):' (Palantir prefix). ==COMMIT_MSG== block has Closes #358 + DCO Signed-off-by inside. ==NO_CHANGELOG== marker present (CI plumbing, no user-visible change). automerge + 'no changelog' labels applied.

PR #354, opened 2026-04-26
fix(merge-bot): trigger on workflow_run when CI goes green

aidanns:
==COMMIT_MSG==
fix(merge-bot): trigger on workflow_run when CI goes green

GitHub's anti-recursion rule suppresses `check_suite` events for any
upstream workflow authenticated with the default `GITHUB_TOKEN`, and
every CI workflow in this repo runs under that token. The bot's
`check_suite: completed` trigger therefore never delivered: every
merge-bot run on PR #346 was triggered by `pull_request` (label
toggles), zero by `check_suite`. Contributors had to remove and
re-add the `automerge` label as an undocumented kick to get the bot
to refire after CI went green — a race-prone, opaque footgun.

Swap `check_suite: completed` for `workflow_run: completed` and
enumerate every PR-gating CI workflow by name. `workflow_run` events
fire regardless of the upstream's authenticating token, so the bot
refires automatically as each watched workflow completes. The
explicit list (excluding Merge Bot itself, release / publish /
scorecard / dependency-submission / mutation / benchmark workflows
that don't run on PRs, and Release Tag which only fires on PR
`closed`) makes adding a new required-check workflow a conscious
two-file change instead of silently regressing the bot for that
workflow.

Knock-on edits: pre-filter `if:` now gates `workflow_run.conclusion ==
'success'`; concurrency-group key falls back to
`workflow_run.head_sha` so triggers serialise per-SHA; the
`Resolve target PR` step's `check_suite)` case arm is renamed
`workflow_run)` and reads the head SHA from
`github.event.workflow_run.head_sha`. The comment block at the top
of the file now explains the GITHUB_TOKEN / anti-recursion rationale
inline so archaeologists don't have to grep issue history.

Static-pin tests under `tests/test_merge_bot_workflow_run_trigger.py`
assert the public-facing shape of the fix: presence of `workflow_run`,
absence of `check_suite`, the watched-workflows set matching the
curated EXPECTED_WORKFLOWS frozenset, every name in that set
resolving to a real workflow file, and the four downstream knock-on
edits (pre-filter, concurrency, case arm, env-var). Same regression-
pin pattern as `test_merge_bot_rollup_dedupe.py` (PR #350).

Closes #348

Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

### Files touched

[...]

PR #350, opened 2026-04-26
fix(merge-bot): dedupe statusCheckRollup by latest run per check

aidanns:
==COMMIT_MSG==
fix(merge-bot): dedupe statusCheckRollup by latest run per check

`statusCheckRollup` returns one entry per check *run*, not per check
*name*. A retried check — workflow rerun, fix-then-push, close+reopen
retrigger — appears multiple times in the rollup, and the historical
FAILURE entry never disappears for the lifetime of the head SHA. The
existing failure / pending predicates in `Inspect required checks`
selected every entry blindly, so once a check had ever failed on a
SHA the bot refused to merge forever (origin: PR #346, where
`changelog-lint` failed on the first run, passed on the close+reopen
retrigger, and kept the bot pinned at `FAILED_CHECKS: changelog-lint`
until the stale failed run was manually deleted via `gh run delete`).

Group rollup entries by `name` (falling back to `.context` for legacy
commit statuses) and keep only the most-recent entry per group via
`max_by(.startedAt // "")` before applying the failure / pending
predicates. The dedupe is applied to *both* selectors — without it on
the pending side, a stale `IN_PROGRESS` from a cancelled rerun would
pin the bot in the wait-clean-exit branch indefinitely.

Closes #347

Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

### Files touched

- `.github/workflows/merge-bot.yml` — `Inspect required checks` step gains a `group_by(.name // .context // "<unnamed>") | map(max_by(.startedAt // ""))` step on both the `failed` and `pending` jq filters. Surrounding comment expanded to document the WHY (rollup carries one entry per *run*, fail-then-pass would otherwise wedge the bot indefinitely) and references issue #347 for archaeologists.
- `tests/test_merge_bot_rollup_dedupe.py` — new pytest module pinning the dedupe predicate. Extracts the `failed=` / `pending=` jq blocks straight out of the workflow YAML (so the test cannot drift from a hand-typed copy) and feeds hand-rolled rollup payloads through real `jq -r`. Asserts on the structural invariant (`group_by` + `max_by` present in *both* filters) plus the behavioural cases.

### Internal decisions

[...]

PR #362, opened 2026-04-26
fix(build_release): keep numeric refs off the start of wrapped lines

### LINKED ISSUE #357: fix(build_release): never wrap with a numeric-reference token at line start

aidanns:
==COMMIT_MSG==
fix(build_release): keep numeric refs off the start of wrapped lines

`scripts/changelog/build_release.py:_wrap_paragraph` did greedy
whitespace-split word wrap at 72 chars. When source prose contained
a numbered reference like `ADR 0011.`, `issue 1234)`, or `RFC 2119`
and the wrap point landed between the noun and the digits, the
resulting line opened with `<digits>[.)]` — which the
`==COMMIT_MSG==` validator's `numbered list item` rule
(`scripts/validate-commit-msg-block.py`) rejects as an ordered-list
item. The auto-generated `chore(release): 0.14.0` body in PR #355
hit this on the line `0011.` `scripts/setup-devcontainer-signing.sh`
writes the new schema..., and every future release whose entries
reference an ADR / issue / spec by number was equally vulnerable.

Add a `NUMERIC_PERIOD_RE = re.compile(r"^\d+[.)]")` constant and
have `_wrap_paragraph` keep `<digits>[.)]` tokens off the start of
a wrapped line. When a numeric token would otherwise overflow the
current line, the preceding token is moved down with it so the
wrap point sits between two ordinary tokens — preserving both the
72-char width rule and the no-numbered-list rule. A simpler glue-
to-previous variant was tried first but pushed lines past 72,
trading one validator failure for another. URL-preserving
behaviour is unchanged; hyphen-joined identifiers like
`CVE-2024-12345` still wrap normally because they don't match the
new pattern.

Tested via `render_commit_msg_block` (the public surface the
validator actually sees) with descriptions crafted to land
`0011.` and `1234)` at line start under the buggy wrapper plus a
sanity case for `CVE-2024-12345`. The regression assertion checks
both validator rules simultaneously so a future "fix" can't
re-introduce the width failure. The companion validator-side
fix is tracked separately as defense-in-depth.

Closes #357
==COMMIT_MSG==

### Why this is a fix, not a chore

This is release-tooling code, but the failure mode is a hard block
on opening any release PR whose changelog entries mention an ADR or
issue by number — i.e. it stops releases from going out. Following
the convention of the recently-merged sibling fix (PR #354,
`fix(merge-bot): ...`), this lands as `fix:` with a hand-authored
changelog entry rather than `chore:` + `no changelog`.

### Title shortening

The original title from issue #357 (`fix(build_release): never wrap
with numeric-reference token at line start`) was 73 chars — one over
the validator's 72-char limit. Shortened to `fix(build_release): keep
numeric refs off the start of wrapped lines` (68 chars).

### Test plan

- [x] `task changelog:test` (197 passed)
- [x] `task test` (742 passed)
- [x] `task lint`
- [x] `task typecheck`
- [x] Reproduced the failure mode against unfixed `build_release.py`:
      both the `ADR 0011.` and `(issue 1234)` regression tests fail,
      the `CVE-2024-12345` sanity test passes — confirming the new
      tests bite the bug and don't over-reach.
- [x] Verified the rendered output of all three regression cases
      stays under 72 chars per line AND avoids numbered-list-shaped
      starts (the assertion helper checks both).
- [ ] After merge, regenerate #355's body via re-running the release
      workflow and confirm the `==COMMIT_MSG==` lint passes.

### Out of scope (per issue)

- Tightening the validator regex itself — companion bug, separate PR.
- Re-wrapping CHANGELOG.md content — different code path,
  CHANGELOG.md tolerates ordered lists.

### DIFF
--- a/changelog/@unreleased/pr-362-build-release-numeric-reference-wrap-fix.yml
+++ b/changelog/@unreleased/pr-362-build-release-numeric-reference-wrap-fix.yml
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
+#
+# SPDX-License-Identifier: MIT
+
+type: fix
+fix:
+  description: |
+    `scripts/changelog/build_release.py` no longer produces a
+    release-PR `==COMMIT_MSG==` body that fails the PR-body lint
+    when changelog descriptions cite an ADR, issue, or spec by
+    number. Greedy 72-char word wrap previously landed tokens like
+    `0011.` (from `ADR 0011.`) or `1234)` (from `(issue 1234)`) at
+    line start, which `validate-commit-msg-block.py`'s
+    `numbered list item` rule rejects as an ordered-list item — see
+    PR #355's broken `chore(release): 0.14.0` body. The renderer now
+    refuses to break immediately before a `<digits>[.)]` token,
+    accepting a soft overflow rather than producing a line the
+    validator rejects. Hyphen-joined identifiers like
+    `CVE-2024-12345` still wrap normally.
+  links:
+    - https://github.com/aidanns/agent-auth/issues/357
+    - https://github.com/aidanns/agent-auth/pull/362
--- a/changelog/@unreleased/pr-362-build-release-numeric-reference-wrap-fix.yml
+++ b/changelog/@unreleased/pr-362-build-release-numeric-reference-wrap-fix.yml
@@ -12,11 +12,12 @@ fix:
     `0011.` (from `ADR 0011.`) or `1234)` (from `(issue 1234)`) at
     line start, which `validate-commit-msg-block.py`'s
     `numbered list item` rule rejects as an ordered-list item — see
-    PR #355's broken `chore(release): 0.14.0` body. The renderer now
-    refuses to break immediately before a `<digits>[.)]` token,
-    accepting a soft overflow rather than producing a line the
-    validator rejects. Hyphen-joined identifiers like
-    `CVE-2024-12345` still wrap normally.
+    PR #355's broken `chore(release): 0.14.0` body. When a numeric
+    token would otherwise overflow the current line, the renderer
+    now moves the preceding token down with it so the wrap point
+    sits between two ordinary tokens — preserving both the 72-char
+    width rule and the no-numbered-list rule. Hyphen-joined
+    identifiers like `CVE-2024-12345` still wrap normally.
   links:
     - https://github.com/aidanns/agent-auth/issues/357
     - https://github.com/aidanns/agent-auth/pull/362
--- a/scripts/changelog/build_release.py
+++ b/scripts/changelog/build_release.py
@@ -84,6 +84,16 @@ SECTION_HEADINGS: dict[EntryType, str] = {
 # Matches `validate-commit-msg-block.py`'s `MAX_LINE_WIDTH`.
 COMMIT_MSG_WRAP = 72
 
+# Tokens shaped like `<digits>.` or `<digits>)` look like ordered-list
+# items to `validate-commit-msg-block.py`'s `numbered list item` rule
+# (DISALLOWED_PATTERNS) when they land at the start of a wrapped line.
+# This recurs in changelog prose because numbered references such as
+# `ADR 0011.`, `issue 1234)`, `RFC 2119`, or `PR 357.` are common —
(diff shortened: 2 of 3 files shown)


aidanns:
Claude Reviewer: Self-review (no general-purpose Agent tool available in this environment).

Reviewed the diff against `CLAUDE.md`, `.claude/instructions/python.md`, `.claude/instructions/testing-standards.md`, and `CONTRIBUTING.md`.

Findings:
1. Found a real bug in the first iteration — `_wrap_paragraph`'s glue-to-previous strategy was producing >72-char lines, which the validator's `check_line_width` rejects. Fixed by moving the preceding token down with the numeric token instead, so the wrap point sits between two ordinary tokens. Widened the regression assertion to require both validator rules pass simultaneously so a future regression can't reintroduce the width failure.
2. Tests exercise the public surface (`render_commit_msg_block`) per testing-standards.md; `_wrap_paragraph` is a private helper.
3. Bug-trigger fixtures verified to fail against unfixed `build_release.py` and pass with the fix; sanity case for `CVE-2024-12345` passes both ways.
4. Naming, types, docstrings consistent with surrounding code.
5. Hand-authored changelog entry follows the convention from sibling `fix(merge-bot)` PRs (#350 / #354).
6. No scope creep, no breaking changes, no security implications.

LGTM.

Issue #324, opened 2026-04-25
feat(release): build per-package wheels and switch install.sh to wheel+venv install (no uv prereq)


aidanns:
## Context

Two related gaps to close:

**1. The release pipeline is broken and has been since the workspace split.**

`/.github/workflows/release-publish.yml` runs `uv build` from the repo root. The root `pyproject.toml` is a workspace shell (`name = "agent-auth-workspace"`, version `0.0.0`) and setuptools rejects it:

> `error: Multiple top-level packages discovered in a flat-layout: ['plans', 'docker', 'design', 'LICENSES', 'packages']`

Every recent release run (`v0.13.0`, `v0.12.2`, `v0.12.1`, `v0.12.0`, `v0.11.0`, …) failed at the *Build sdist and wheel* step. `gh release view v0.13.0 --json assets` returns `{\"assets\":[]}` — no wheels, no SBOMs, no signatures, no SLSA provenance attestation. The supply-chain story in `SECURITY.md` § *Supply-chain artifacts* currently has nothing to verify against. This regressed silently because:

- semantic-release still tags + writes CHANGELOG entries, so the release *appears* to publish.
- The install scripts at `packages/<svc>/install.sh` use `uv tool install git+https://github.com/aidanns/agent-auth.git@<tag>#subdirectory=packages/<svc>`, which clones the repo and builds from source. They never read release assets, so a broken artifact pipeline produces no install-time symptom.

**2. The install.sh files require `uv` as an end-user prereq.**

Each `packages/<svc>/install.sh` rejects machines without `uv` and tells the user to install it first. uv is mainstream for *development* but uncommon as a prereq for *installing* a Python CLI — pipx and wheel-into-venv (the `aidanns/systems-engineering` pattern) are the dominant paths, and Homebrew formulas (`Language::Python::Virtualenv`) always use stdlib `venv` + `pip`. Requiring uv blocks a future Homebrew formula in `aidanns/homebrew-tools` from following the standard pattern.

**Reference shape — `aidanns/systems-engineering`:**

- Builds one wheel + `.sha256` per release, uploads both as GitHub release assets.
- `install.sh` requires only `python3 ≥ 3.10` with the stdlib `venv` module. It downloads the wheel + checksum, verifies the checksum, creates `~/.local/share/<tool>/venv` with `python3 -m venv`, runs `<venv>/bin/pip install <wheel>`, and symlinks `<venv>/bin/<tool>` into `~/.local/bin/`.
[...]

PR #370, opened 2026-04-26
feature(ci): add release-build dry-run and post-publish asset smoke

### LINKED ISSUE #325: feat(ci): pre-tag dry-run and post-publish smoke test for release-publish.yml
## Context

The current `release-publish.yml` has been failing on every release since the workspace split (`v0.11.0` through `v0.13.0` — see #324 *Context* for the failure log). This was undetected for ~13 releases because:

- semantic-release in `release.yml` tags + writes `CHANGELOG.md` + opens the GitHub release *before* `release-publish.yml` runs, so the release looks successful from the outside.
- `packages/<svc>/install.sh` installs from `git+...#subdirectory=...`, never reading release assets, so a missing wheel produces no install-time symptom.
- `release-publish.yml` only runs `on: push: tags:` — there is no PR-time trigger that would have caught the regression in the workspace-split PR.
- No post-publish step verifies that the expected assets actually landed on the release.

`SECURITY.md` § *Supply-chain artifacts* documents a verification recipe (cosign + slsa-verifier) that has been silently unverifiable on every recent release.

This issue is **independent of #324**: regardless of whether the install path is `uv tool install git+...` or wheel+venv, we want CI checks that catch the publish workflow breaking. #324 fixes the current breakage; this issue prevents the next one.

### 1. PR-time dry-run of the build step

Add a job (either in `check.yml` or a new `release-dryrun.yml`) that runs on every PR and on push to `main`:

- Checks out the repo.
- Runs whatever build command `release-publish.yml` would run (today `uv build`; after #324, the per-package build loop) into a temp `dist/` directory.
- Asserts the expected artefact set is present (one `.whl` + `.tar.gz` per workspace member).

The job does **not** sign, generate SBOMs, or attempt to upload — it only proves the build doesn't error and produces the expected outputs. This is the highest-value, lowest-cost check: it would have caught the workspace-split regression in the PR that introduced it.

To keep the check honest, the build commands in `release-publish.yml` and the dry-run job MUST be sourced from a single script (e.g. `scripts/build-release-artifacts.sh`) so they cannot drift. Both jobs invoke the script; only the publish job runs the sign/upload steps that follow.

### 2. Post-publish smoke test on every tag

Add a final job to `release-publish.yml` (after `provenance`) that runs on the same `on: push: tags:` trigger:

- Waits for the release-asset upload to be visible via `gh release view <tag> --json assets`.
- Asserts the expected per-asset set is present (wheel, sdist, sha256, SBOM, cosign bundles, SLSA provenance) — currently one each, after #324 one set per workspace member.
- Fails loudly if any expected asset is missing.

This catches the case where the build succeeds but uploads silently drop artefacts (e.g. a `gh release upload` flag change, a permissions regression, a typo in the asset list).

### 3. Branch protection on the dry-run job

Once (1) is live and stable, add it to required status checks for `main`. After that point, no PR can merge with a broken release-build step.

This requires the dry-run job to be reliable (no flake from network errors, etc.) — gate the rule on a soak window of at least a couple of weeks of green runs.

### 4. Optional: end-to-end install smoke test

Going further (and overlapping with #324's acceptance), a job that runs `packages/<svc>/install.sh <tag>` against the just-published tag and asserts the binary works (`<bin> --version` returns a non-empty version). Highest fidelity, slowest, and the most likely to flake.

Probably out of scope for this issue if #324 already adds a `--local`-mode equivalent — leaving it called out so we can decide whether the local-mode test in #324 is enough or whether we want a true end-to-end test against the published release.

## Out of scope

- Fixing the *current* breakage of `release-publish.yml`. That's #324.
- Changes to install.sh or the install pattern. That's #324.
- Reworking the SLSA provenance or supply-chain story. ADR 0016 stands; this issue verifies it works, doesn't change it.

## Acceptance

- A PR-time job runs the release-build step (via a shared script) and fails the PR if `uv build` (or its replacement) errors or produces an unexpected artefact set.
- A post-publish job in `release-publish.yml` enumerates expected release assets and fails the workflow run if any are missing on `gh release view`.
- The shared build script is the single source of truth for build commands — `release-publish.yml` and the dry-run job both invoke it; no duplicated `uv build` invocations.
- Once stable, the PR-time job is added to branch-protection required checks (separate config-only PR is fine).
- Backfill: open a follow-up issue if any release between this work landing and now needs its assets re-published (i.e. tags published while the pipeline was broken).

## Related

- #324 — fixes the current `release-publish.yml` breakage and switches install.sh to wheel+venv. The dry-run job here verifies whatever build shape #324 lands on.
- ADR 0016 — `design/decisions/0016-release-supply-chain.md` (signing, SBOM, SLSA). This issue ensures the artefacts ADR 0016 promises actually exist on every release.
- `SECURITY.md` § *Supply-chain artifacts* — verification recipe consumers point at; relies on these CI gates to remain honest.

aidanns:
==COMMIT_MSG==
feature(ci): add release-build dry-run and post-publish asset smoke

Adds two CI gates that catch broken release-publish pipelines before
they ship. The PR-time dry-run lives in `release-dryrun.yml` and
runs the same build commands the publish path will run, via the new
shared `scripts/build-release-artifacts.sh`. That script is the
single source of truth — `release-publish.yml`'s `Build sdist and
wheel` step now invokes it too, so the two cannot drift.

The second gate is a post-publish `verify-assets` job in
`release-publish.yml` that enumerates assets via `gh release view
--json assets` and fails the workflow if any expected category
(wheel, sdist, SBOM, cosign bundle, SLSA provenance) is missing.

Closes the gap that hid the workspace-split release-build
regression for ~13 consecutive releases. The dry-run is
informational (`continue-on-error: true` on its build + assert
steps) until issue #324 unbreaks the build; a follow-up
config-only PR will move it to required status checks once the
soak window passes.

Closes #325

==COMMIT_MSG==

### Test plan

- [x] `treefmt --ci` passes locally on the new files.
- [x] `shellcheck scripts/build-release-artifacts.sh` clean.
- [x] `scripts/build-release-artifacts.sh --help` renders the usage block.
- [x] Smoke-tested the `count` jq function inline in the verify-assets
      step against a fabricated 9-asset list (1 wheel, 1 sdist, 2 SBOMs,
      4 cosign bundles, 1 SLSA provenance) — all five floors satisfied.
- [x] Confirmed the script faithfully reproduces today's release-publish
      build behaviour: `uv build` at the workspace root, which currently
      fails with the workspace-split setuptools error. That failure is
      the dry-run's intended PR-time signal.

### Internal decisions (resolved per the issue body)

- **Dry-run lives in `release-dryrun.yml`, not grafted onto `check.yml`.**
  Project convention is workflow-per-purpose (`release-pr.yml`,
  `release-publish.yml`, `release-tag.yml` are all separate); a release
  dry-run belongs on the same axis. `check.yml` exists for lint/format/
  ripsecrets only.
- **Shared script path: `scripts/build-release-artifacts.sh`** as
  suggested in the issue body. There is already a `scripts/build.sh`
  that runs `uv run ... python -m build` for dev-time use; the new
  script is intentionally separate because it captures the
  release-publish-specific build (`uv build`), and conflating the two
  would re-introduce drift between dev and release.
- **Asset verification logic is inline bash in `release-publish.yml`,**
  per the issue's explicit "the simplest option" guidance. It uses
  `gh release view --json assets` + `jq endswith(...)` per category;
  no extra script.
- **`continue-on-error: true` on the dry-run's build + assert steps
  (not at the job level).** A job-level toggle leaves the *job's*
  check-run reporting FAILURE in the PR rollup even when the
  workflow's overall conclusion is success — that's enough to wedge
  the merge-bot's gate. Step-level toggles let a failed step show as
  red inside the run while the job's check-run conclusion stays
  green. The issue body anticipates the dry-run as informational
  before being added to required status checks ("the soak window of
  at least a couple of weeks of green runs"); on `main` today the
  build is broken (#324 fixes it), so a hard gate would make every
  PR unmergeable until #324 lands. Comments at the call-site say
  removing both `continue-on-error` lines is the explicit "now
  required" signal.

### What this PR does NOT do (scope guards)

- Does not touch branch-protection settings — out of scope per the
  issue's acceptance criterion 3 ("separate config-only PR is fine").
- Does not edit unrelated `.github/` files. The merge-bot's
  `workflow_run` trigger list is intentionally not updated for the
  new `Release Dryrun` workflow; with the step-level
  `continue-on-error: true` the rollup status is success and the
  existing label-apply / commit-push triggers handle the retrigger
  path.
- Does not fix the existing release-publish breakage — that is #324.

### Backfill follow-up

Issue body acceptance #5 calls for a backfill issue if any release
between #325 landing and `release-publish.yml` going green needs its
assets re-published. Any release tagged while the build is broken
(today's state) has zero assets and can be re-published once #324
lands by re-tagging or re-running the publish workflow on the tag.
Tracking that should be done as a follow-up issue against #324's
landing PR; flagging here so it doesn't fall through.

### Self-review only

The `Task` / `general-purpose` review subagent type is unavailable in
this harness, so this PR did not receive an automated reviewer pass.
I self-reviewed the diff against the project's standard checklist:
scope creep (none — only the four files the issue body calls for),
undocumented decisions (all four internal choices justified above),
missing tests (CI infra; the dry-run job itself is the test surface),
dead code (none), unclear naming (`build-release-artifacts.sh` /
`verify-assets` / `Release Dryrun` mirror existing
`release-publish.yml` shapes), unflagged breaking changes (the only
behavioural change is `release-publish.yml`'s build step now invokes
a shell script instead of `run: uv build` — documented at the
call-site and in the changelog), security (new workflow uses
`permissions: contents: read` only; verify-assets uses default
`GITHUB_TOKEN` with read-only release access), convention
violations (SPDX headers, `set -euo pipefail`, conventional commit
prefix `feature(ci):`, DCO sign-off, manual changelog entry, title
under 72 chars all present).


changelog/@unreleased/pr-370-release-build-checks.yml
```diff
@@ -0,0 +1,21 @@
...
+    `verify-assets` job in `release-publish.yml` that fails the
+    workflow if any expected asset (wheel, sdist, SBOM, cosign
+    bundle, SLSA provenance) is missing from the GitHub release.
+    Closes the gap that hid the workspace-split release-build
+    regression for ~13 consecutive releases. The dry-run is
+    informational (`continue-on-error: true`) until #324 unbreaks
+    the build; a follow-up config-only PR will move it to required
+    status checks once the soak window passes.
+  links:
+    - https://github.com/aidanns/agent-auth/issues/325
```
aidanns, line 21: Claude Reviewer: Acceptance #5 of issue #325 says to **open a follow-up issue** if any release between this PR landing and the build going green needs its assets re-published. The PR body acknowledges this in "Backfill follow-up" but defers the issue creation ("should be done as a follow-up issue against #324s landing PR"). This is the kind of follow-up that quietly falls through the cracks; recommend opening the placeholder issue now (e.g. "Re-publish assets for v0.11.0 .. v0.14.0 once #324 lands") and linking it from this PR body so it survives the merge.

.github/workflows/release-dryrun.yml
```diff
@@ -0,0 +1,99 @@
...
+name: Release Dryrun
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
```
aidanns, line 31: Claude Reviewer: This workflow has no `concurrency:` block, so a force-push or rapid `git push` cycles will queue full duplicate runs of the dry-run on the same PR head. Other PR-triggered workflows in this repo cancel-in-progress on `pull_request` (compare e.g. how check.yml is shaped). Suggest adding:

```yaml
concurrency:
  group: release-dryrun-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

Minor / soft observation — saves runner minutes during the soak window when this fires on every PR.

.github/workflows/release-publish.yml
```diff
@@ -142,3 +144,74 @@ jobs:
...
+          echo "  wheels=${wheels} sdists=${sdists} sboms=${sboms}"
+          echo "  cosign_bundles=${cosign_bundles} slsa=${slsa}"
+          echo "Asset list:"
+          jq -r '.[]' <<<"${assets}" | sed 's/^/  /'
+
+          fail=0
+          # `wheel + sdist` = 2 source artefacts; one SBOM each = 2;
+          # cosign signs all 4 of those = 4 bundles; SLSA = 1 file.
+          [[ "${wheels}" -ge 1 ]] || { echo "::error::no .whl asset on release ${TAG}"; fail=1; }
+          [[ "${sdists}" -ge 1 ]] || { echo "::error::no .tar.gz sdist asset on release ${TAG}"; fail=1; }
```
aidanns, line 232: Claude Reviewer: The thresholds use `-ge` (e.g. `[[ "${cosign_bundles}" -ge 4 ]]`) which is the right floor for "at least N", but it also means the check passes silently if the publish job uploads MORE than expected (e.g. a future change accidentally double-uploads, or a flag regression duplicates an asset under a slightly-renamed pattern). Per ADR 0016 the artefact set is exact (1 wheel + 1 sdist + 2 SBOMs + 4 cosign bundles + 1 SLSA today), so an exact-equality assert (with the same per-package multiplier comment for the post-#324 future) would catch upload-duplication regressions too. Soft observation — if the rationale for `-ge` is "post-#324 the count multiplies" then please add a comment justifying that explicitly; otherwise tighten to `-eq`.

changelog/@unreleased/pr-370-release-build-checks.yml
aidanns: Claude: Done — opened #372 as the placeholder so it survives the merge. PR body updated to link it.

.github/workflows/release-dryrun.yml
aidanns: Claude: Done! Added a `concurrency:` block keyed on `github.event.pull_request.number || github.ref` with `cancel-in-progress: true`, with an inline comment explaining why the dry-run differs from the release-publish/-tag workflows (no mid-release state to protect).

.github/workflows/release-publish.yml
aidanns: Claude: Took the comment-justification path rather than tightening to `-eq`. The `-ge` floor is load-bearing for the #324 transition (per-package multiplier flips every count), so tightening to `-eq` here would make this gate fail the moment #324 lands. Expanded the inline rationale to call this out explicitly and left a "tighten once #324 settles" follow-up note for once the per-package count is stable.

PR #378, opened 2026-04-26
feature(release): per-package wheels + venv install (no uv)

### LINKED ISSUE #324: feat(release): build per-package wheels and switch install.sh to wheel+venv install (no uv prereq)

aidanns:
==COMMIT_MSG==
The release pipeline has produced zero artefacts since the
ADR 0032 workspace split: ``uv build`` from the repo root hits
the multiple-top-level-packages flat-layout error, so every
release tag (v0.13.0, v0.12.x, v0.11.0, ...) carries no wheels,
no SBOMs, no signatures, and no SLSA provenance. The breakage
went silent because each ``packages/<svc>/install.sh`` used
``uv tool install git+...`` rather than reading any release
asset. ADR 0043 walks through the per-package distribution
fix in full.

Build: ``uv build --all-packages --out-dir dist/`` produces
one wheel + one sdist per workspace member. Per-artefact loops
emit one .sha256, one SPDX SBOM, and one cosign keyless
signature bundle (plus a bundle for the SBOM) for every wheel
and sdist. SLSA Build L3 provenance subjects enumerate every
wheel + sdist in the release.

Install: each ``packages/<svc>/install.sh`` is a from-scratch
rewrite. End-user prereq drops from ``uv`` to Python 3.11+
with the stdlib ``venv`` module on PATH. Each script downloads
the service wheel + every workspace-dep wheel listed in a
hard-coded array, verifies sha256, installs into
``~/.local/share/<svc>/venv`` via ``python3 -m venv`` and
``pip install``, symlinks every console-script entrypoint into
``~/.local/bin``, and verifies with ``<bin> --version``.
``--uninstall`` removes both. ``--local <dir>`` lets CI and
local-bench work install from a directory of locally-built
wheels without a published release.

Defence in depth: a new ``install-from-wheels`` matrix job in
``test.yml`` (services: things-cli, gpg-bridge) builds every
workspace wheel via ``uv build --all-packages``, generates
.sha256 sidecars, and runs ``install.sh --local dist/`` in a
shell with every UV_* env var unset and uv stripped from PATH.
The job asserts the entrypoint binary is on ``~/.local/bin``,
``--version`` exits 0, the venv pip list carries both the
service and ``agent-auth-common``, and ``--uninstall`` cleans
up. This catches the silent-breakage failure mode that produced
the current state.

The CI integration test lives in ``test.yml`` rather than
``release-publish.yml``: adding a ``pull_request`` trigger to a
workflow with ``id-token: write`` + ``contents: write`` would
widen its trigger surface to every PR, while ``test.yml``
already runs on PRs with ``contents: read`` only.

Workspace-dep resolution is hard-coded per service (today every
service's runtime workspace closure is just
``{agent-auth-common}``). The audit gate is ADR 0036's
``scripts/verify_workspace_deps.py``; each install.sh's array
points at it inline.

Refs ADR 0016 (release supply chain), ADR 0032 (workspace
split), ADR 0035 (single workspace train), ADR 0036 (workspace
dep allowlist).

Closes #324
==COMMIT_MSG==

### Test plan

- [ ] CI ``install-from-wheels`` matrix passes (things-cli + gpg-bridge).
- [ ] Locally: ``uv build --all-packages --out-dir /tmp/aa-dist/``,
  generate ``.sha256`` for each, run
  ``packages/things-cli/install.sh --local /tmp/aa-dist/``, verify
  ``things-cli --version`` works and ``~/.local/share/things-cli/venv/bin/pip list``
  shows both ``things-cli`` and ``agent-auth-common``.
- [ ] ``--uninstall`` cleanup checked for at least one multi-entrypoint
  service (``agent-auth`` ships ``agent-auth`` + ``agent-auth-notifier``).
- [ ] All PR-gating checks green before applying ``automerge``.

### Notes for reviewer

- The integration-test job is in ``test.yml``, not
  ``release-publish.yml``. ADR 0043 captures the rationale: keeping
  ``release-publish.yml``'s trigger surface tag-only avoids widening
  a workflow that holds ``id-token: write`` and ``contents: write``.
- ``release-publish.yml`` retains its tag-only trigger and SHA-pinned
  third-party actions (cosign, sbom-action, setup-uv, slsa-generator
  exception called out in the existing comment block).
- All install scripts are mechanically generated from a single
  template (the renderer was a throwaway under ``/tmp`` and is not
  committed). Per-service variables (description, entrypoints, wheel
  prefix) are at the top of each script.
- The hand-authored changelog entry under
  ``changelog/@unreleased/`` will follow this PR open.

### Self-review only

The Task / general-purpose subagent tool is not available in this
environment, so I self-reviewed the diff against the same checklist
the orchestrator gave the spawn-a-reviewer step:

- ``release-publish.yml``: per-package build loop produces every
  expected asset (wheel, sdist, .sha256, SBOM, cosign sig); SLSA
  hashes enumerate every wheel + sdist subject (mapfile + sha256sum
  on the sorted basename list); no remaining ``uv build`` at repo
  root.
- Each ``packages/<svc>/install.sh``: drops uv prereq; uses
  python3+venv; downloads service wheel + every workspace-dep wheel
  + .sha256; verifies checksums (sha256sum on Linux, shasum -a 256
  -c on macOS); installs with workspace deps first; symlinks every
  entrypoint; calls ``<bin> --version``; ``--uninstall`` removes
  symlinks + venv; ``--local <dir>`` works (manually validated all
  six install scripts against ``uv build --all-packages`` output).
- CI integration test exists in ``test.yml`` and exercises the
  install path end-to-end with uv stripped from PATH.
- ``SECURITY.md`` § Supply-chain artifacts updated (per-package
  layout, verification recipe loops over every wheel + sdist).
- New ADR 0043 covers the distribution model and is linked from
  ``design/decisions/README.md``.
- Hand-authored ``changelog/@unreleased/pr-378-...yml`` with
  ``feature:`` type.
- Palantir prefix (``feature(release):``), ``==COMMIT_MSG==`` block
  shape validated by ``scripts/validate-commit-msg-block.py``, DCO
  sign-off in commit trailers, no ``(#<issue>)`` in subject.


aidanns (commented):
Claude Reviewer: Stale comments + loose post-publish assert in release-publish.yml `verify-assets` (lines ~189-193). The header comment says "Today the expected set is one entry per category … Issue #324 multiplies the wheel/sdist/SBOM/cosign categories per workspace member; **when that lands, tighten the asserts below**". This PR IS #324 landing — the per-package build is now committed in `scripts/build-release-artifacts.sh` (`uv build --all-packages`). After merge the asserts `wheels >= 1`, `sdists >= 1`, `sboms >= 2`, `cosign_bundles >= 4` no longer reflect the actual contract (7 wheels + 7 sdists + 14 SBOMs + 28 cosign bundles for the seven workspace members today). The thresholds still pass for any future regression that drops a single package's artefact set silently — the gate will green on 6/7 wheels even though one package shipped nothing. Same situation in `.github/workflows/release-dryrun.yml` lines ~75-84 ("Today the workspace-root `uv build` produces one wheel + one sdist; after #324 the per-package loop produces one pair per packages/<svc>/. … Tighten this to per-package enumeration once #324 lands"). Recommend either (a) tightening both asserts to enumerate per-package names in this PR, or (b) opening a follow-up issue and updating the comments to point at it (per the project rule that deferred work must be tracked in the issue tracker, not just in code comments). I checked `gh issue list` and no such follow-up exists today.

aidanns:
Claude: Tracked in #382 rather than addressed in this PR. The asserts and stale comments live in `.github/workflows/release-publish.yml` and `release-dryrun.yml`; #324 (this PR's issue) called for the per-package build, not the assertion tightening, and the deferred-work convention says the cleanup belongs in its own tracker entry. Issue #382 captures the proposed exact-count assertions plus the "tighten when #324 lands" comment cleanup, with origin link back to the review comment.

PR #383, opened 2026-04-27
improvement(merge-bot): auto-update behind PRs and re-trigger after CI

### LINKED ISSUE #373: operations: grant merge-bot App the workflows: write permission
## Context

PR #368 (which modifies `.github/workflows/release-pr.yml`, `release-tag.yml`, and `merge-bot.yml`) cannot be merged by merge-bot. The merge API call returns:

```
Repository rule violations found

refusing to allow a GitHub App to create or update workflow `.github/workflows/release-pr.yml` without `workflows` permission

12 of 12 required status checks are expected.

(HTTP 405)
```

The `12 of 12 required status checks are expected` line is a separate `BEHIND` issue (tracked in a sibling issue). This issue covers the first line: the `agent-auth-merge-bot` GitHub App is missing the `workflows` write permission, so it cannot author squash-merge commits that touch any file under `.github/workflows/`.

This is a structural blocker — every PR that edits a workflow file (which includes most release-automation, CI tightening, and dependabot workflow bumps) will hit this until the App is granted the permission.

## Proposal

This is a **manual GitHub App configuration change**, not a code change.

1. Open the `agent-auth-merge-bot` GitHub App settings: https://github.com/settings/apps/agent-auth-merge-bot → **Permissions & events** → **Repository permissions**.
2. Set **Workflows** to **Read and write**.
3. Save. GitHub will ask the installation owner to **review and accept** the new permission grant on the `aidanns/agent-auth` installation: https://github.com/settings/installations → agent-auth-merge-bot → **Configure** → accept the new permission.

No `contents: write` is required for this fix — `workflows` write is sufficient because the bot only writes via `PUT /pulls/{n}/merge` (which inherits the App's permissions when authoring the squash commit).

## Verification

After granting:
- Re-trigger merge-bot against PR #368 (`gh workflow run merge-bot.yml -f pr_number=368`) once the BEHIND issue is also resolved, and confirm the merge succeeds.
- The next workflow-file-touching PR after #368 should land via merge-bot without the `workflows` permission error.

## Out of scope

- The `BEHIND` / strict-required-status-checks behaviour — separate issue.
- Granting any other permissions (`contents`, `issues`, etc.) — only `workflows` write is needed for this failure mode.

## Related

- PR #368 — first PR to surface the failure.
- `docs/release/merge-bot-setup.md` — should be updated with the `workflows` permission requirement once granted (fold into the auto-update issue's PR or a follow-up doc patch).
- ADR 0038 — merge-bot via GitHub App.

### LINKED ISSUE #374: improvement(merge-bot): auto-update behind PRs and re-trigger after CI
## Context

The `main` ruleset (id `15002184`) has `strict_required_status_checks_policy: true` — required status checks must run on the post-rebase head SHA before a PR can merge. When a PR falls behind `main`, `mergeStateStatus` becomes `BEHIND` and the merge API rejects the call:

```
Repository rule violations found
12 of 12 required status checks are expected.
(HTTP 405)
```

Today merge-bot (`.github/workflows/merge-bot.yml`) treats this as a hard failure: it posts a `Claude: Cannot merge — GitHub merge API rejected the call: ...` comment and the PR sits stuck with the `automerge` label until a human rebases by hand. PR #368 is the immediate trigger — it has been blocked on this for several days while CI was busy on `main`.

## Proposal

Teach merge-bot to **auto-update behind PRs** and rely on the existing `workflow_run.completed` retrigger to do the second-pass merge:

1. After the `Re-check PR is still open` step (line ~437) and before the merge call, fetch `mergeStateStatus`.
2. If `mergeStateStatus == \"BEHIND\"`, call `PUT /repos/{owner}/{repo}/pulls/{n}/update-branch` (default merge-method is fine — it produces a merge commit, which is what GitHub's \"Update branch\" button does), post a one-line `Claude: Branch was behind main — updated; will retry merge after CI completes.` comment, and exit cleanly. The branch update creates a new head commit, CI retriggers, and `workflow_run.completed` (already wired in lines 59-75) fires merge-bot again. The second pass sees a green, up-to-date PR and merges normally.
3. **Loop guard** — track the number of times merge-bot has updated the same PR via the `automerge` label timeline (or a dedicated PR comment marker). If the PR has been auto-updated more than N times (suggest **3**) without merging, refuse and post `Claude: Auto-update loop exceeded — main is moving too fast or this PR keeps falling behind. Investigate manually.` This stops a flapping `main` from rebasing the same PR forever.
4. The `update-branch` API requires the App to have `contents: write`. Currently merge-bot has `contents: read` (line 89). Grant `contents: write` in the workflow `permissions:` block — the App installation already needs it for this to work; verify the App permission grant separately if the call returns 403.

## Tradeoffs

- **Each auto-update burns a full CI cycle.** A PR that sits behind through three rebases will run all 12 required checks 4× total. Acceptable cost for hands-off merging; the loop guard caps the worst case.
- **Merge commits in PR history.** `update-branch` defaults to a merge commit on the PR head; the squash-merge collapses it on landing, so `main` history stays linear. No semantic change vs. a manual rebase.
- **Race with concurrent pushes.** If a contributor pushes to the PR head between the BEHIND-check and the update-branch call, GitHub returns 422 (`expected_head_sha`). Treat as a clean exit — the contributor's push will retrigger the bot anyway.

## Acceptance

- merge-bot detects `BEHIND` after the green-check gate, calls `PUT /pulls/{n}/update-branch`, comments, and exits 0.
- The subsequent `workflow_run.completed` trigger after CI rebuilds against the new head fires merge-bot again, which then merges (assuming all checks pass on the new SHA).
- Loop guard caps auto-updates at 3 per PR; the 4th BEHIND state hard-fails with the operator-facing comment.
- `contents: write` added to the workflow's `permissions:` block.
- `docs/release/merge-bot-setup.md` updated with the auto-update behaviour and the `contents: write` requirement.

## Out of scope

- The `workflows: write` App-permission grant (separate issue — both block PR #368 but they're orthogonal failure modes).
- Switching the ruleset off `strict_required_status_checks_policy` — the strict policy is the right default for a release-bumping `main`; the bot should adapt to it, not the other way round.
- Generalising to other rule violations (signed commits, missing reviews, etc.) — `BEHIND` is the only one the bot can recover from autonomously.

## Related

- PR #368 — first PR to wedge on this; will land once both this and the workflows-permission issue are resolved.
- ADR 0038 — merge-bot via GitHub App.
- `docs/release/merge-bot-setup.md` — documents the bot's required permissions; needs an update with this change.
- Memory: `feedback_merge_bot_quirks.md` already notes the `workflow_run.completed` retrigger pattern this proposal builds on.

aidanns:
==COMMIT_MSG==
improvement(merge-bot): auto-update behind PRs and re-trigger after CI

The `main` ruleset enforces `strict_required_status_checks_policy:
true`, so a PR whose head sits behind `main` cannot merge until
its required checks have re-run on the post-rebase head SHA. Until
now merge-bot treated this as a hard failure: it surfaced the
405 from `PUT /pulls/{n}/merge` as a `Claude: Cannot merge —
GitHub merge API rejected the call: ...` comment and the PR sat
stuck with the `automerge` label until a human rebased by hand.

Add a BEHIND-handling step between the green-check / DCO gate and
the merge call. The step re-fetches `mergeStateStatus` (the value
in the earlier `meta` step can stale) and, when the PR is behind
`main`, calls `PUT /pulls/{n}/update-branch` with
`expected_head_sha` pinned. The `update-branch` call fast-forwards
the PR head onto `main` (default merge-method, equivalent to
clicking GitHub's "Update branch" button); the resulting new head
SHA retriggers every PR-gating CI workflow, and the existing
`workflow_run.completed` trigger re-fires merge-bot for the
second-pass merge once those workflows complete.

A loop guard caps the worst case at three auto-updates per PR.
The step counts prior `Claude: Branch was behind main — updated;`
comments authored by the bot's App identity (matched on
`user.login == "${APP_LOGIN}"` plus the literal comment-body
prefix, where `${APP_LOGIN}` is the App's `<slug>[bot]` resolved
from the token-mint step's `app-slug` output) and refuses the
fourth update with `Claude: Auto-update loop exceeded — main is
moving too fast or this PR keeps falling behind. Investigate
manually.` so a flapping `main` does not let the bot rebase the
same PR forever.

Race with concurrent contributor pushes is handled by the
`expected_head_sha` pin: a push between the BEHIND-check and the
update-branch call produces a documented 422
(`expected_head_sha`); the bot treats that as a clean exit
because the contributor's push will retrigger the bot via
`workflow_run.completed` once CI runs against the new head.

The new step needs `contents: write` on both the workflow's
top-level `permissions:` block and the App installation. Bump
the workflow permission and document both grants — plus the
separately-required `workflows: write` App grant that was the
subject of the parallel issue (the App installation needs it
because every PR that edits any file under `.github/workflows/`
otherwise refuses to merge through `PUT /pulls/{n}/merge`).
The user has already granted `workflows: write` on the App via
the GitHub UI; the doc update is the only remaining piece for
that issue.

Closes #373
Closes #374

==COMMIT_MSG==

### Self-review only

The general-purpose subagent (`Task` / `Agent` tool) was unavailable
in this environment, so this PR was self-reviewed against the
project conventions checklist rather than reviewed by a separate
subagent. Self-review checklist outcome:

- BEHIND-handling step exits cleanly without falling through to the
  merge call. The new step sets `behind=true` on the auto-update
  success path and on the `expected_head_sha` 422 race exit; the
  merge call's `if:` picked up an additional
  `&& steps.behind.outputs.behind == 'false'` so any non-`'false'`
  value (including the empty string when the step `exit 1`s on the
  loop-guard or update-branch failure paths) skips the merge.
- Loop-guard query targets the App's own comments specifically, not
  arbitrary `Claude: ` posts. Filter is `user.login == APP_LOGIN`
  where `APP_LOGIN` is `<app-slug>[bot]` resolved from the
  `actions/create-github-app-token` step's `app-slug` output, plus
  a literal comment-body prefix match.
- `contents: write` granted at the workflow's top-level
  `permissions:` block (was `contents: read`). Workflow's release-
  path-adjacent SHA pinning policy already applied to the three
  `uses:` references — no further hardening needed.
- Doc and changelog accurately describe the new behaviour. Doc
  also documents the `workflows: write` App grant (the #373
  piece) and explains why it is required.
- Trigger surface is unchanged — the new logic is an additional
  step inside the existing job, not a new trigger or
  permission scope. CodeQL `js/actions/command-injection` rules
  pass: every interpolation into the `run:` body uses `env:`
  bindings rather than direct `${{ }}` interpolation.

### Implementation choices

- **Loop-guard marker** is a literal comment-body prefix
  (`Claude: Branch was behind main — updated;`) plus the
  `<app-slug>[bot]` author filter. The prefix is specific enough
  that no other Claude-prefixed comment collides; the author
  filter narrows it to *this* App's auto-updates as a defence-in-
  depth.
- **`mergeStateStatus` re-fetched** in the new step rather than
  reused from the earlier `meta` step. The value can become stale
  over the lifetime of a run (someone could push to `main` between
  steps); a fresh read is the source of truth for the BEHIND
  decision.
- **`expected_head_sha` pinned** on the `update-branch` call so a
  contributor push that races us produces a documented 422 rather
  than a silent fast-forward past their commit. The 422 is treated
  as a clean exit — the push will retrigger merge-bot anyway.
- **Step ordering**: BEHIND-handling sits between the
  `Re-check PR is still open` step and the merge call, gated on
  the same `signoff.outputs.ok == 'true'` and
  `recheck.outputs.skip != 'true'` predicates. The merge call
  picked up an additional `&& steps.behind.outputs.behind == 'false'`
  predicate so a successful auto-update or a 422-clean-exit skips
  it cleanly.
- **`gh api ... --paginate --slurp`** wraps every page into a
  single outer JSON array so the comment-counting `jq` pass sees
  every comment regardless of page boundary; the marker is passed
  via `--arg` to keep its UTF-8 em dash and `;` intact through
  shell quoting.

### Test plan

- [ ] Static: `yq .github/workflows/merge-bot.yml > /dev/null` parses cleanly.
- [ ] Static: top-level `permissions:` shows `contents: write` (was `contents: read`).
- [ ] Static: the new `behind` step's `if:` chain is identical to the prior `merge_call` `if:` chain so the BEHIND check runs whenever the merge call would have run.
- [ ] Static: `merge_call.if` now includes `&& steps.behind.outputs.behind == 'false'` so a successful auto-update skips the merge step.
- [ ] Static: `shellcheck` on the new step's `run:` body parses clean (locally verified).
- [ ] Static: `jq --arg marker '<prefix>' --arg login '<app-slug>[bot]' '[.[] | .[] | select(.user.login == $login) | select(.body | startswith($marker))] | length'` returns the expected count against a synthetic two-page comments fixture (locally verified).
- [ ] Post-merge: open a PR, let it fall behind `main` (push something to `main`), apply `automerge`, confirm the bot posts the auto-update comment and CI retriggers; confirm the second `workflow_run.completed` re-fires the bot and the PR merges.
- [ ] Post-merge: confirm `Claude: Auto-update loop exceeded` fires after three prior auto-updates on the same PR (most easily smoke-tested by lowering the threshold to 1 in a follow-up branch and reverting before merge).

### DIFF
--- a/.github/workflows/merge-bot.yml
+++ b/.github/workflows/merge-bot.yml
@@ -525,6 +525,12 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           PR_NUMBER: ${{ steps.pr.outputs.pr }}
+          # `app-slug` is the App's slug (e.g. `agent-auth-merge-bot`);
+          # the matching bot user login is `<slug>[bot]`. Resolved
+          # from the token-mint step rather than at runtime so the
+          # filter doesn't depend on App identity endpoints that
+          # behave inconsistently with installation tokens.
+          APP_LOGIN: ${{ steps.app-token.outputs.app-slug }}[bot]
         run: |
           set -euo pipefail
           # Re-fetch mergeStateStatus + headRefOid so the BEHIND
@@ -541,20 +547,21 @@ jobs:
             exit 0
           fi
           # Count prior auto-update comments by this App. The App's
-          # comments are authored by `agent-auth-merge-bot[bot]` (App
-          # type = `Bot`); the prefix below is the literal first line
-          # of the success path's auto-update comment. `--paginate
+          # comments are authored as `${APP_LOGIN}` (the App slug
+          # from the token-mint step plus `[bot]`). `--paginate
           # --slurp` wraps every page into an outer JSON array so a
           # single `jq` pass sees every comment regardless of page
-          # boundary; flattening with `.[] | .[]` gives the per-comment
-          # iteration. The marker is passed via `--arg` rather than
-          # templated into the jq program so the embedded UTF-8 em
-          # dash and `;` survive shell quoting verbatim.
+          # boundary; flattening with `.[] | .[]` gives the
+          # per-comment iteration. The marker is passed via `--arg`
+          # rather than templated into the jq program so the
+          # embedded UTF-8 em dash and `;` survive shell quoting
+          # verbatim.
           marker="Claude: Branch was behind main — updated;"
           prior_updates="$(gh api \
             "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" \
             --paginate --slurp \
-            | jq --arg marker "${marker}" '[.[] | .[] | select(.user.type == "Bot") | select(.body | startswith($marker))] | length')"
+            | jq --arg marker "${marker}" --arg login "${APP_LOGIN}" \
+                '[.[] | .[] | select(.user.login == $login) | select(.body | startswith($marker))] | length')"
           echo "merge-bot: prior auto-update count=${prior_updates}"
           if [[ "${prior_updates}" -ge 3 ]]; then
             gh pr comment "${PR_NUMBER}" \
--- a/.github/workflows/merge-bot.yml
+++ b/.github/workflows/merge-bot.yml
@@ -81,12 +81,15 @@ on:
         type: number
 
 # `pull-requests: write` is needed to post the `Claude: ...` comment
-# and (via the App token) to call the merge endpoint. NO
-# `contents: write` — the bot does not push commits or tags. NO
-# `id-token: write` — the bot is not on the SLSA-attestation path.
+# and (via the App token) to call the merge endpoint. `contents: write`
+# is needed for the BEHIND-handling auto-update branch — the
+# `PUT /pulls/{n}/update-branch` endpoint requires write access to
+# the repo's contents to push the merge commit that fast-forwards
+# the PR head onto the latest `main`. NO `id-token: write` — the bot
+# is not on the SLSA-attestation path.
 permissions:
   pull-requests: write
-  contents: read
+  contents: write
   checks: read
 
 # Prevent two triggers (label + workflow_run) from racing on the same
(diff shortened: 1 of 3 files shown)


docs/release/merge-bot-setup.md
```diff
@@ -247,21 +267,87 @@ If any of those signals is wrong, the most likely cause is a setup
...
+  `contents: write` grant on the App installation, or a
+  ruleset rule that blocks the App from pushing the merge commit.
+- **Auto-update loop exceeded**: the bot has already updated this
+  PR three times without merging. `main` is moving faster than
+  CI completes for this PR, or the PR keeps re-falling-behind for
+  another reason (e.g. a conflict that `update-branch` can't
+  resolve). Investigate manually; once the underlying cause is
+  fixed, push a new commit to reset the count (the prior
+  auto-update comments stay on the PR but the `automerge` label
+  retrigger will start a fresh attempt).
```
aidanns: Claude Reviewer: Doc accuracy — the recovery path for "Auto-update loop exceeded" is incorrect. The doc says "push a new commit to reset the count (the prior auto-update comments stay on the PR but the `automerge` label retrigger will start a fresh attempt)". But the loop-guard query in `merge-bot.yml` counts ALL prior auto-update comments on the PR regardless of when they were posted (it filters by author + literal prefix, not by head SHA or timestamp). A new commit triggers a fresh CI cycle which re-fires the bot, the bot still sees `prior_updates >= 3`, and refuses again. The actual recovery requires deleting those auto-update comments by hand (or the maintainer running `update-branch` manually so the next bot run finds the PR no longer BEHIND). Suggest either fixing the doc to say "delete the prior auto-update comments to reset the count, then re-apply the `automerge` label" or filtering the count by `created_at >= <PR head ref push timestamp>` so a fresh push genuinely resets it.

.github/workflows/merge-bot.yml
```diff
@@ -490,9 +493,116 @@ jobs:
...
+          set -euo pipefail
+          # Re-fetch mergeStateStatus + headRefOid so the BEHIND
+          # decision is based on current state, not state from the
+          # earlier `meta` step.
+          state_payload="$(gh pr view "${PR_NUMBER}" \
+            --repo "${GITHUB_REPOSITORY}" \
+            --json mergeStateStatus,headRefOid)"
+          merge_state="$(jq -r '.mergeStateStatus' <<<"${state_payload}")"
+          head_sha="$(jq -r '.headRefOid' <<<"${state_payload}")"
+          if [[ "${merge_state}" != "BEHIND" ]]; then
```
aidanns, line 583: Claude Reviewer: `mergeStateStatus == UNKNOWN` race not handled. When `workflow_run.completed` fires very shortly after a push (or right after the `automerge` label is applied), GitHub may not yet have finished computing the merge state and `gh pr view ... --json mergeStateStatus` returns `UNKNOWN`. The new step at line 544 only special-cases the literal string `BEHIND`; an `UNKNOWN` value falls through and `behind=false` is set, so the merge call runs and (if the PR is in fact behind) is refused with a 405 → `Claude: Cannot merge — GitHub merge API rejected the call: ...`. Functionally non-fatal — the next `workflow_run.completed` retrigger will re-evaluate and probably hit a stable mergeStateStatus — but it produces a noisy "Cannot merge" comment that contradicts the auto-update story documented in the PR body. Consider polling `gh pr view` 1-2 times with a short backoff when `mergeStateStatus == "UNKNOWN"` (similar to the pattern other bots use for mergeability), or treating UNKNOWN as "exit cleanly and wait for the next retrigger" rather than as "proceed to merge call".

.github/workflows/merge-bot.yml
```diff
@@ -490,9 +493,116 @@ jobs:
...
+          # per-comment iteration. The marker is passed via `--arg`
+          # rather than templated into the jq program so the
+          # embedded UTF-8 em dash and `;` survive shell quoting
+          # verbatim.
+          marker="Claude: Branch was behind main — updated;"
+          prior_updates="$(gh api \
+            "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" \
+            --paginate --slurp \
+            | jq --arg marker "${marker}" --arg login "${APP_LOGIN}" \
+                '[.[] | .[] | select(.user.login == $login) | select(.body | startswith($marker))] | length')"
```
aidanns: Claude Reviewer: `gh api --paginate --slurp` is new to this workflow with no precedent in the repo (no other workflow uses `--slurp`). It is supported on gh ≥ 2.50 and GitHub-hosted runners ship a recent gh, but if the runner image ever pins an older gh — or someone runs this workflow on a self-hosted runner with a stale gh — the `--slurp` flag silently changes shape (older gh concatenated pages into a single array, so `.[] | .[]` would iterate strings inside comment bodies rather than comments inside a page array, and the count would be 0). Two tactical options: (a) add a brief comment in the YAML noting the gh-version dependency so the next reader does not get caught out, or (b) avoid `--slurp` entirely by using gh `--jq` to project each page to comment IDs and counting via `wc -l`, which has been the older pattern in this workflow file.

docs/release/merge-bot-setup.md
aidanns: Claude: Done! Adopted option (b): added a `created_at` filter anchored on the head commit's `committer.date` so the loop guard only counts auto-update comments posted since the most recent push. Updated the doc to explain that pushing a new commit resets the count.

.github/workflows/merge-bot.yml
aidanns: Claude: Done! Added a polling loop (10 attempts × 3s ≈ 30s) on `mergeStateStatus`. If it stays UNKNOWN after polling, the step posts `Claude: GitHub is still computing merge state — will retry on next workflow_run.completed.`, sets `behind=true` to skip the merge call, and exits 0.

.github/workflows/merge-bot.yml
aidanns: Claude: Done! Swapped `--paginate --slurp` for `gh api --paginate | jq -r ... | wc -l` to match the rest of the workflow and avoid the gh ≥ 2.50 dependency. Folded the new `created_at >= since` filter into the same single query rather than running a second one.

PR #393, opened 2026-04-27
improvement(ci): assert exact per-package release-asset counts

### LINKED ISSUE #382: Tighten release-build / dryrun assertions to per-package counts
## Context

Follow-up from review of #378 (https://github.com/aidanns/agent-auth/pull/378#issuecomment-4322199803).

`scripts/build-release-artifacts.sh` now produces per-package wheels via `uv build --all-packages`. The release pipeline emits one wheel + one sdist + one .sha256 + one SPDX SBOM + one cosign signature bundle per artefact and one cosign signature bundle per SBOM, for every workspace member.

Two CI gates still enforce the pre-#378 \"weakest meaningful\" thresholds:

1. \`.github/workflows/release-publish.yml\` — \`verify-assets\` job asserts \`wheels >= 1\`, \`sboms >= 2\`, \`cosign_bundles >= 4\`. With N packages the actual counts should be \`wheels = sdists = N\`, \`sboms = 2N\` (one per wheel + one per sdist), \`cosign_bundles = 4N\` (one per artefact + one per SBOM).
2. \`.github/workflows/release-dryrun.yml\` — assert step asserts at least one wheel + one sdist.

Both files carry comments saying \"tighten this once #324 lands\". #324 has now landed (#378). The current loose thresholds let a regression that drops one package's full asset set pass the gate.

## Proposed work

- Read the workspace member count from \`pyproject.toml\` \`[tool.uv.workspace].members\` (or grep \`packages/*/pyproject.toml\`) and assert exact counts in both jobs.
- Update / remove the \"tighten when #324 lands\" comments.
- Confirm the \`verify-assets\` job's per-package enumeration also catches the case where a wheel is built but its sidecar (.sha256, .spdx.json, .sig.bundle) is missing.

## Authorisation

This is a \`.github/\` change — it was deferred out of #378 because that PR's scope was already authorised under #324 and adding this would have widened the diff into territory the issue body did not call out. Tracking here so it doesn't get lost.

## Origin

Surfaced by the independent review of #378 — see https://github.com/aidanns/agent-auth/pull/378#issuecomment-4322199803.

aidanns:
==COMMIT_MSG==
improvement(ci): assert exact per-package release-asset counts

The post-publish `verify-assets` gate and the PR-time
`release-dryrun` gate both used `-ge 1` thresholds because the
workspace-root build that pre-dated #324 only produced one wheel
+ one sdist; tightening would have failed once the per-package
build (#324, ADR 0044) landed. With #324 now on `main` the soft
thresholds let a regression that drops one package's full asset
set pass silently.

Both gates now derive the workspace member count from
`packages/<svc>/pyproject.toml` (seven today, automatic on add)
and assert exact per-category counts — N wheels, N sdists, 2N
.sha256 sidecars, 2N SBOMs, 4N cosign bundles, 1 SLSA provenance
file. .sha256 sidecar enumeration is new; the previous gate did
not count them at all even though they have been uploaded since
#324. Adds an `actions/checkout` step to `verify-assets` so the
member count is read from the repo rather than hard-coded.

Closes #382

==COMMIT_MSG==

## Review notes

Origin: deferred from the independent review of #378 (https://github.com/aidanns/agent-auth/pull/378#issuecomment-4322199803). Issue #382 captured the deferred work; this PR addresses it.

Verification:
- `find packages -mindepth 2 -maxdepth 2 -name pyproject.toml | wc -l` → `7` on main today, matches `SECURITY.md` § Supply-chain artifacts ("seven packages today").
- The expected counts match the actual upload globs in `release-publish.yml`:
  - `*.whl` `*.whl.sha256` `*.whl.sig.bundle` → N + N + N
  - `*.tar.gz` `*.tar.gz.sha256` `*.tar.gz.sig.bundle` → N + N + N
  - `*.spdx.json` `*.spdx.json.sig.bundle` → 2N + 2N
  - SLSA generator → 1 `multiple.intoto.jsonl`
- Total: N wheels, N sdists, 2N `.sha256`, 2N SBOMs, 4N cosign bundles, 1 SLSA file. ✅
- `endswith` filters are mutually exclusive — `<wheel>.spdx.json.sig.bundle` ends with `.sig.bundle`, not `.spdx.json` or `.whl`, so no double-counting.

Choices made (no explicit issue guidance):

- Member count derivation by `find packages/*/pyproject.toml` rather than parsing the workspace glob from `pyproject.toml` — `pyproject.toml`'s `members = ["packages/*"]` is hard-coded today, and a future change to that glob would also need a corresponding find-pattern change here. Documenting the link in the workflow comment is enough.
- Added `actions/checkout` to `verify-assets` (new step). The job previously did no checkout because it only needed the GitHub release API. The shallow clone (`fetch-depth: 1`) is fine — only `packages/*/pyproject.toml` is read.
- Kept the `continue-on-error: true` on the dryrun assert step. The header comment explicitly calls removing it the trigger for moving the gate from informational to required, and that's a separate config-only follow-up tracked in the dryrun comment.

### DIFF
--- a/.github/workflows/release-dryrun.yml
+++ b/.github/workflows/release-dryrun.yml
@@ -84,6 +84,10 @@ jobs:
         run: |
           set -euo pipefail
           expected=$(find packages -mindepth 2 -maxdepth 2 -name pyproject.toml | wc -l)
+          if [[ "${expected}" -lt 1 ]]; then
+            echo "::error::no workspace members found under packages/* — dryrun count derivation is broken"
+            exit 1
+          fi
           # `nullglob` makes an unmatched glob expand to nothing
           # rather than the literal pattern, so the array stays empty
           # when the build step (above) failed and produced no dist/.
--- a/.github/workflows/release-dryrun.yml
+++ b/.github/workflows/release-dryrun.yml
@@ -72,21 +72,18 @@ jobs:
         continue-on-error: true
         run: scripts/build-release-artifacts.sh --out dist
 
-      # Today the workspace-root `uv build` produces one wheel + one
-      # sdist (when it works); after #324 the per-package loop
-      # produces one pair per `packages/<svc>/`. The assertion is
-      # "the build produced at least one wheel + one sdist", which
-      # is the weakest meaningful contract that still catches the
-      # workspace-split regression (build error → zero artefacts)
-      # and stays valid through the #324 transition without needing
-      # an update. Tighten this to per-package enumeration once #324
-      # lands (issue tracks the move from informational to required
-      # status check).
+      # Per-package release-asset layout (ADR 0044): one wheel + one
+      # sdist per workspace member. The expected count is derived from
+      # `packages/<svc>/pyproject.toml` so adding a new workspace
+      # member tightens the gate automatically with no edit here. A
+      # regression that drops one package's pair will fail the assert
+      # — the looser "at least one" gate this replaces could not.
       - name: Assert expected artefact set
         continue-on-error: true
         shell: bash
         run: |
           set -euo pipefail
+          expected=$(find packages -mindepth 2 -maxdepth 2 -name pyproject.toml | wc -l)
           # `nullglob` makes an unmatched glob expand to nothing
           # rather than the literal pattern, so the array stays empty
           # when the build step (above) failed and produced no dist/.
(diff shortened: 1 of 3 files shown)


aidanns (commented):
Claude Reviewer: Minor consistency gap between the two gates — `release-publish.yml` guards `members < 1` (lines 218-221) but `release-dryrun.yml` does not. If the workspace ever ends up restructured such that `find packages -mindepth 2 -maxdepth 2 -name pyproject.toml` returns 0 (renamed top-level dir, accidental delete, checkout race), the dryrun's asserts silently pass with `expected=0`, `wheels=0`, `sdists=0` because `0 -ne 0` is false. Adding the same guard to dryrun (`if [[ "${expected}" -lt 1 ]]; then echo '::error::no workspace members found ...'; exit 1; fi`) keeps the two gates symmetric and rules out the silent-zero failure mode that motivated this PR in the first place. Not a blocker — the actual failure mode in the wild would be `uv build --all-packages` skipping a member, where the find still returns N and the count mismatch fires correctly.

aidanns:
Claude: Done! Added the symmetric `expected -lt 1` guard to the dryrun assert step in fa7af1d.

Issue #408, opened 2026-04-27
fix(release): v0.16.0 wheels versioned 0.0.0+unknown instead of 0.16.0


aidanns:
## Symptom

`v0.16.0` (released 2026-04-27T12:42:17Z) shipped 70 supply-chain assets in the per-package shape introduced by #324/#378, but every wheel and sdist is versioned `0.0.0+unknown` instead of `0.16.0`.

## Impact

- The verification recipe in `SECURITY.md` § *Supply-chain artifacts* (around line 467) hardcodes `${PKG}-${VERSION}-py3-none-any.whl` patterns. Consumers running the documented `cosign verify` recipe against `v0.16.0` cannot match these names — the artefacts exist but with the wrong filenames.
- Anyone `pip install`-ing a wheel from this release sees PEP 440 local-version-segment dummy `0.0.0+unknown`, not `0.16.0`.
- `verify-assets` (post-#393) only counts categories, not name patterns; the dummy version did not trigger any gate.

## Hypothesis

`uv build --all-packages` is producing a placeholder version because the build step is reading the workspace state in some configuration where the git tag isn't being threaded into the package version. Likely candidates:

- Missing `setuptools-scm` / `hatch-vcs` config in `packages/*/pyproject.toml` (or its `uv` equivalent).
- The `Configure git identity` and tag-checkout sequence in `release-publish.yml` left the workspace in detached-HEAD without a discoverable tag for version derivation.
- `[tool.uv.sources]` workspace overrides masking the version inference.

Needs repro inside the publish workflow environment.

## Acceptance

- A subsequent release tag (e.g. `v0.16.1`) ships wheels named `${PKG}-0.16.1-py3-none-any.whl`, sdists named `${PKG}-0.16.1.tar.gz`, etc.
- The PR that fixes it includes a unit / dry-run check that asserts the version derived during `uv build` matches the git tag.
- Optional: a `release-dryrun.yml` assertion that catches this class of regression PR-time.

## Out of scope

- Backfilling `v0.16.0`'s assets — that's up to you (re-tagging or accepting the bug). Track separately if needed.
- The pre-#324 missing-assets gap on `v0.11.0..v0.15.3` — that's #372.

## Origin

Surfaced by triage on #372. Run inspection: `gh release view v0.16.0 --json assets --jq '.assets[].name'` shows `*-0.0.0+unknown-*.whl` patterns.

aidanns:
[...]

Issue #372, opened 2026-04-26
chore(release): re-publish missing assets for v0.11.0..v0.13.0 once #324 lands


aidanns:
## Context

Tracked from PR #370 (closes #325). Per the post-publish smoke test added in #325, every release between the workspace split and #324 landing has been silently failing to upload the expected per-asset set (wheels, sdists, SBOMs, cosign bundles, SLSA provenance). Specifically `v0.11.0`, `v0.12.0`, and `v0.13.0` (and any further releases that ship before #324 merges) are missing artefacts that `SECURITY.md` § *Supply-chain artifacts* tells consumers they can verify.

This issue is the placeholder for the cleanup pass once the publish pipeline is healthy again.

## Acceptance

- For each affected tag (`v0.11.0`, `v0.12.0`, `v0.13.0`, plus anything tagged before #324 merges), enumerate the missing assets via `gh release view <tag> --json assets` against the expected set the post-publish smoke test checks for (wheel, sdist, sboms, cosign bundles, SLSA provenance).
- Decide per-tag whether to:
  - Re-run the publish workflow against the existing tag (preferred — keeps tag identity stable), OR
  - Cut a patch release that subsumes the missing supply-chain artefacts and document the gap in the affected tags' release notes.
- Update `SECURITY.md` § *Supply-chain artifacts* with a one-line note for any tag where the verification recipe will not work, so consumers don't try to verify against missing assets.

## Out of scope

- The fix to `release-publish.yml` itself — that's #324.
- The CI gates that prevent this from happening again — that's #325 (this PR's parent).

## Related

- #325 — adds the PR-time dry-run + post-publish smoke test that catches this class of regression going forward.
- #324 — fixes the current `release-publish.yml` breakage.
- ADR 0016 — `design/decisions/0016-release-supply-chain.md`.

aidanns:
Claude: Blocked — this issue is explicitly gated on "once #324 lands". PR #378 (which closes #324) is still OPEN, so the per-tag missing-asset enumeration (the precondition for deciding re-run vs. patch-release) cannot start yet. Parked until #378 merges; resume via `/workflow:implement 372` once that lands.

aidanns:
Claude: Blocker cleared — PR #378 (closing #324) merged 2026-04-27T12:21:39Z, which was the precondition for this work. The `blocked` label is left in place per the parent run choosing to defer this issue to a fresh session; re-run `/workflow:implement 372` when ready to pick it up.

[...]

Issue #270, opened 2026-04-24
refactor: move tests into per-package tests/ trees


aidanns:
Tests currently live in a monolithic `tests/` at the repo root, mixing every package's tests together. With the workspace split (#257) each package should own its own `tests/` tree so per-package runs, coverage floors, and CI matrices become straightforward.

### Target layout
- `packages/agent-auth/tests/` — tokens, crypto, keys, scopes, store, server, approval, rate-limit, audit, migrations, etc.
- `packages/agent-auth-common/tests/` — `tests_support` tests, http client tests, `things_models` tests, `server_metrics` tests.
- `packages/things-bridge/tests/`
- `packages/things-cli/tests/`
- `packages/things-client-cli-applescript/tests/`
- `tests/` at repo root — workspace-wide only: `test_openapi_spec.py`, `test_release_semver.py`, `test_pip_audit_to_sarif.py`, `test_report_scan_failure.py`.

### Required updates
- `pyproject.toml` `[tool.pytest.ini_options]` — per-package pythonpath / coverage targets.
- `[tool.mutmut] tests_dir` entries.
- `conftest.py` splits where helpers are package-specific.
- Any `scripts/verify-*.sh` that greps for test paths.
- Integration tests under `tests/integration/` — decide per-file which package owns them.

### Acceptance
- `pytest packages/<svc>/tests` passes in isolation for each package.
- Root `task test` still runs the full suite.
- Coverage remains at or above the current floor.

### Unblocks
- Per-package coverage floors (separate issue in this milestone).
- Per-package CI matrix work.

PR #410, opened 2026-04-27
improvement(release): add workflow_dispatch to publish for tag re-runs

aidanns:
==COMMIT_MSG==
improvement(release): add workflow_dispatch to publish for tag re-runs

The post-publish asset-count smoke test added in #325 surfaced
that every release between the workspace split (#270) and #324
landing produced an incomplete asset set on the GitHub release
page. Twelve tags (`v0.11.0..v0.15.3`) need a re-publish run to
upload the per-package wheels, sdists, SBOMs, cosign bundles, and
SLSA provenance the verification recipe in SECURITY.md tells
consumers they can verify against. Today `release-publish.yml`
only fires on `push: tags:`, so once a tag is pushed there is no
in-band way to re-run the workflow against it.

Add a `workflow_dispatch` trigger with a required `tag` input so
a maintainer can run `gh workflow run release-publish.yml -f
tag=<TAG>` against an existing tag from the default-branch copy
of the workflow file. Thread `inputs.tag || github.ref_name`
through the concurrency group key, both `actions/checkout` ref:
fields, and the `gh release upload` / `gh release view` env
blocks so a dispatched run binds its build / sign / SBOM / SLSA
chain to the tag's commit (not `main`'s tip) and `--clobber`
overwrites partial assets idempotently. The release-bot App's
`Contents: Read & write` permission gates the API surface, not
the trigger type, so the dispatched run mints installation tokens
exactly the same way as the tag-push path.

Add SECURITY.md notes under § Supply-chain artifacts for tag
ranges where the per-package verification recipe will not match
the release page: `v0.6.0..v0.10.0` ship the legacy single-package
shape (no per-package PACKAGES loop match); `v0.11.0..v0.15.3`
have incomplete asset sets pending the post-merge re-publish ops;
`v0.16.0` carries `0.0.0+unknown`-versioned wheels per #408 so the
recipe's `agent_auth-0.16.0-...` filenames do not exist on the
release.

Re-publish ops against the 12 affected tags are post-merge
operational follow-up, not part of this PR. Issue #372 stays
open across those ops; the orchestrator closes it after the
re-publish loop succeeds.

Refs #372
Refs #408

==COMMIT_MSG==

### Background

`#372` is the cleanup pass for the wheel-less releases between
the workspace split (`#270`) and `#324` (per-package wheels +
venv install). Twelve tags need a re-publish to populate the
per-package asset set the verification recipe in SECURITY.md
documents. `release-publish.yml` today only fires on
`push: tags`, so once a tag is pushed there is no in-band way to
re-run the workflow against it. This PR closes that gap; the
actual `gh workflow run` invocations are post-merge operational
follow-up.

### `${{ github.ref* }}` references touched in `release-publish.yml`

Run `grep -n 'github\.ref\|inputs\.tag' .github/workflows/release-publish.yml`
on the PR branch to cross-check. Adapted sites (post-edit line
numbers):

- **L43 — `concurrency.group`**: now keys on the bare tag name
  via `inputs.tag || github.ref_name`. `github.ref` would resolve
  to `refs/heads/main` for `workflow_dispatch` and split the
  group from the original tag-push run.
- **L62 — `actions/checkout` (publish job)**: new `ref:`
  defaulting to `inputs.tag || github.ref` so the dispatched run
  binds its build / sign / SBOM / SLSA chain to the tag's commit
  rather than `main`'s tip.
- **L163 — `Upload artefacts` env `TAG`**: now
  `inputs.tag || github.ref_name`. Threaded into
  `gh release upload "${TAG}" ...` via the `env:` block (the
  CodeQL `js/actions/command-injection` safe pattern is
  preserved).
- **L233 — `actions/checkout` (verify-assets job)**: same
  `inputs.tag || github.ref` ref so the workspace member count
  read from `packages/<svc>/pyproject.toml` reflects the tag's
  layout.
- **L237 — `Verify expected release assets are uploaded` env
  `TAG`**: now `inputs.tag || github.ref_name`. Same shape as
  L163.

### Test plan

`workflow_dispatch` only fires off the default-branch copy of the
workflow file, so this PR's change is **not** dispatchable
pre-merge.

- [ ] `git diff main...HEAD -- .github/workflows/release-dryrun.yml`
      is empty (only `release-publish.yml` is touched).
- [ ] `grep -n 'github\.ref\|inputs\.tag' .github/workflows/release-publish.yml`
      shows the five adapted sites listed above.
- [ ] `task format -- --check` clean.
- [ ] `task lint` clean.
- [ ] `mdformat --check SECURITY.md plans/republish-missing-assets.md`
      clean.
- [ ] `yq . .github/workflows/release-publish.yml > /dev/null` (YAML
      well-formed).
- [ ] **Post-merge smoke (orchestrator)**: pick an existing tag
      whose release page is correct (e.g. any future green
      release; pre-merge `v0.16.0` works as a smoke target since
      its asset shape is per-package even if the wheel versions
      are wrong) and run
      `gh workflow run release-publish.yml -f tag=<TAG>`. Expect
      the run to go green and the release-page asset set to be
      unchanged (uploads use `--clobber`, idempotent).

### Post-merge operational follow-up (orchestrator runs these)

After this PR merges, run `gh workflow run release-publish.yml -f
tag=<TAG>` against each of the 12 post-split-but-empty tags. The
assets-upload step uses `--clobber` so re-running is safe even
against a release page that already has some (but not all)
assets.

```bash
for TAG in \
  v0.11.0 \
  v0.12.0 v0.12.1 v0.12.2 \
  v0.13.0 v0.13.1 \
  v0.14.0 v0.14.1 \
  v0.15.0 v0.15.1 v0.15.2 v0.15.3
do
  gh workflow run release-publish.yml -f tag="${TAG}"
done
```

After every run completes green and `verify-assets` confirms the
asset count, close `#372`.

The pre-split range (`v0.6.0..v0.10.0`) and `v0.16.0` are **not**
re-publishable through this workflow — pre-split releases use a
different build shape, and `v0.16.0`'s versioning bug needs the
fix in `#408`. SECURITY.md captures that for consumers.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -526,16 +526,116 @@ today. It will not match every historical release tag:
   release-publish pipeline silently produced incomplete asset sets
   before [#324](https://github.com/aidanns/agent-auth/issues/324)
   landed. Re-publish is being driven through
-  [#372](https://github.com/aidanns/agent-auth/issues/372); until
-  each affected release page enumerates the full per-package asset
-  set the recipe's `cosign verify-blob` and `slsa-verifier verify-artifact` steps will fail at the missing-asset-download
-  step.
+  [#372](https://github.com/aidanns/agent-auth/issues/372) by firing
+  `release-publish.yml` via `workflow_dispatch` from `main`. While
+  re-publish is in flight, an affected release page will not yet
+  enumerate the full per-package asset set, and the recipe above
+  fails at the missing-asset-download step. Once re-publish
+  completes, the asset set is present but the signature / provenance
+  identities are bound to `refs/heads/main` rather than
+  `refs/tags/<TAG>`; consumers must use the re-published-tag recipe
+  in [Re-published tag applicability](#re-published-tag-applicability)
+  below.
 - **`v0.16.0`** — wheels and sdists are present in the per-package
   layout but versioned `0.0.0+unknown` instead of `0.16.0` (tracked
   as [#408](https://github.com/aidanns/agent-auth/issues/408)). The
   recipe builds filenames like `agent_auth-0.16.0-py3-none-any.whl`,
   which do not exist on this release. Use the next non-affected
-  release (`v0.16.1` or later) once shipped.
+  release (`v0.16.1` or later) once shipped. If `v0.16.0` is later
+  re-published via `workflow_dispatch` to fix #408, the same
+  branch-bound identity caveats from the re-published-tag recipe
+  below apply.
+
+### Re-published tag applicability
+
+Tags re-published via `gh workflow run release-publish.yml -f tag=<TAG>` (the post-merge ops driven by
+[#372](https://github.com/aidanns/agent-auth/issues/372)) inherit
+their cosign keyless identity and their SLSA provenance source-ref
+from the **dispatch ref** — `refs/heads/main` — rather than from the
+historical tag the workflow checks out. This is a property of GitHub
+OIDC's `job_workflow_ref` claim and the SLSA generator's
+`external_parameters.workflow.ref` encoding for `workflow_dispatch`
+runs. Firing `--ref <TAG>` would re-bind both to the tag, but the
+12 affected tags (`v0.11.0..v0.15.3`) predate the addition of
+`workflow_dispatch:` to `release-publish.yml`, so the tag-ref
+dispatch path is not viable for them.
+
+For these re-published assets, the verification recipe above fails
+at the cosign / SLSA signature check. Use this alternate recipe
+instead — it is identical to the recipe above except for the
+identity / source-ref expressions:
+
+```bash
+: "${TAG:?set TAG=vX.Y.Z to the re-published release tag}"
+
+VERSION="${TAG#v}"
+# Re-published tags carry an identity bound to refs/heads/main, not
+# refs/tags/<TAG>. The match must be exact to avoid the regex /
+# unescaped-`.` widening risk that the primary recipe also avoids.
+IDENTITY="https://github.com/aidanns/agent-auth/.github/workflows/release-publish.yml@refs/heads/main"
+
+PACKAGES=(
+  agent_auth
+  agent_auth_common
+  gpg_bridge
+  gpg_cli
+  things_bridge
+  things_cli
+  things_client_cli_applescript
+)
+
+SDISTS=()
+WHEELS=()
+for pkg in "${PACKAGES[@]}"; do
+  SDISTS+=("${pkg}-${VERSION}.tar.gz")
+  WHEELS+=("${pkg}-${VERSION}-py3-none-any.whl")
+done
+
+# 1. cosign verify-blob with the main-bound identity.
+for f in "${SDISTS[@]}" "${WHEELS[@]}"; do
+  for target in "${f}" "${f}.spdx.json"; do
+    cosign verify-blob \
+      --bundle "${target}.sig.bundle" \
+      --certificate-identity "${IDENTITY}" \
+      --certificate-oidc-issuer "https://token.actions.githubusercontent.com" \
+      "${target}"
+  done
+done
+
+# 2. slsa-verifier with --source-branch main instead of
+#    --source-tag "${TAG}". The provenance encodes the dispatch ref
+#    (refs/heads/main), so a tag-pinned check fails closed.
+for f in "${SDISTS[@]}" "${WHEELS[@]}"; do
+  slsa-verifier verify-artifact \
+    --provenance-path multiple.intoto.jsonl \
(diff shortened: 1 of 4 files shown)


SECURITY.md
```diff
@@ -512,6 +512,31 @@ enforce the same property for the SLSA provenance: an attestation
...
+  release-publish pipeline silently produced incomplete asset sets
+  before [#324](https://github.com/aidanns/agent-auth/issues/324)
+  landed. Re-publish is being driven through
+  [#372](https://github.com/aidanns/agent-auth/issues/372); until
+  each affected release page enumerates the full per-package asset
+  set the recipe's `cosign verify-blob` and `slsa-verifier verify-artifact` steps will fail at the missing-asset-download
+  step.
+- **`v0.16.0`** — wheels and sdists are present in the per-package
+  layout but versioned `0.0.0+unknown` instead of `0.16.0` (tracked
+  as [#408](https://github.com/aidanns/agent-auth/issues/408)). The
```
aidanns, line 541: Claude Reviewer: **High — `workflow_dispatch` re-publish will produce signatures the SECURITY.md verification recipe rejects.**

When a workflow signs with cosign keyless, the cert SAN is bound to the OIDC `job_workflow_ref` claim — for `workflow_dispatch` runs this is the **ref the dispatch was fired from** (i.e. `refs/heads/main` by default), not `actions/checkout`'s `ref:`. The same applies to the SLSA generator: the provenance encodes the workflow run's `github.ref` as the source ref. So the planned post-merge ops (`gh workflow run release-publish.yml -f tag=v0.13.0`) will produce:

- cosign certs with identity `.../release-publish.yml@refs/heads/main`, while SECURITY.md L450 builds `IDENTITY=".../release-publish.yml@refs/tags/${TAG}"` and uses **exact match** (no regexp). `cosign verify-blob` will reject.
- SLSA provenance with `external_parameters.workflow.ref = refs/heads/main`, while SECURITY.md L494 calls `slsa-verifier --source-tag "${TAG}"`. `slsa-verifier` will reject.

The plan acknowledges this in the threat-model section as a defense-in-depth feature, but the practical consequence is that re-publishing the 12 affected tags via `workflow_dispatch` from `main` **does not solve #372** — consumers running the documented recipe still cannot verify those releases, just for a different reason.

Two ways out (pick one before running the post-merge ops sequence):

1. **Fire the dispatch against the tag**: `gh workflow run release-publish.yml -f tag=v0.13.0 --ref v0.13.0`. GitHub will then bind both the cosign OIDC identity and the SLSA provenance ref to the tag. Caveat: this runs the **tag's copy** of `release-publish.yml`, which on pre-#324 tags lacks the per-package build, the `verify-assets` job, and the `workflow_dispatch` trigger itself — so this option is not viable for `v0.11.0..v0.15.3` without re-tagging those tags onto a workflow that supports it.
2. **Document the identity drift in SECURITY.md**: extend the new `### Tag applicability` block to call out that re-published tags carry a cosign identity of `...@refs/heads/main` and a SLSA `source-tag` of `main` rather than the tag, and either provide an alternate verification recipe or accept that the per-tag verification recipe is not applicable to re-published assets.

Either way, the PR's SECURITY.md note for `v0.11.0..v0.15.3` ("until each affected release page enumerates the full per-package asset set the recipe's `cosign verify-blob` and `slsa-verifier verify-artifact` steps will fail at the missing-asset-download step") is incomplete — even after re-publish populates the asset set, those two steps will still fail, just at the signature/provenance check step instead.

Recommend resolving this before running the post-merge ops in the PR body. Worth a tracking issue if option 2 is chosen, since it materially changes the verification story for #372.

aidanns: Claude: Done! Verified the cosign / SLSA identity claim against `v0.16.0` (push-tag) — its cert SAN is `…/release-publish.yml@refs/tags/v0.16.0` and the SLSA `configSource.uri` is `git+…@refs/tags/v0.16.0`, both bound to the trigger ref. For `workflow_dispatch` from `main` (the only viable path for `v0.11.0..v0.15.3`, since those tags predate `workflow_dispatch:` in the workflow file) the same fields bind to `refs/heads/main`. Option 1 (`--ref <TAG>`) is therefore not viable for those tags; option 2 was the only mechanical fix. Implemented: added `### Re-published tag applicability` to SECURITY.md with an alternate recipe (cosign `--certificate-identity …@refs/heads/main`, slsa-verifier `--source-branch main`), updated the existing `v0.11.0..v0.15.3` and `v0.16.0` bullets to point at it, and explicitly stated the security implication — the branch-bound recipe proves *some* maintainer dispatched `release-publish.yml` from `main` and is strictly weaker than the per-tag-bound primary recipe (every re-published tag shares the same identity). Plan threat-model section also updated to acknowledge the verification gap.

PR #417, opened 2026-04-27
chore(ci): enforce two-tier scope allowlist (package or area)

### LINKED ISSUE #402: lint: standardise scope allowlist with two-tier rule (package-or-area)
## Problem

The set of scopes used in \`type(scope):\` is sprawling and drifting past the CONTRIBUTING.md allowlist. Recent main-branch examples:

- [\`b57eaf8\`](https://github.com/aidanns/agent-auth/commit/b57eaf8) — \`improvement(merge-bot): …\`
- [\`29fbac3\`](https://github.com/aidanns/agent-auth/commit/29fbac3) — \`improvement(pr-lint): …\`
- [\`bcfbc64\`](https://github.com/aidanns/agent-auth/commit/bcfbc64) — \`fix(build_release): …\`
- [\`aa7a23f\`](https://github.com/aidanns/agent-auth/commit/aa7a23f) — \`fix(release-pr): …\`

None of these scopes are on the CONTRIBUTING.md allowlist. The fact that \`requireScope: false\` is set in \`pr-lint.yml\` and the type allowlist is enforced separately means scope drift sails through.

## Proposed fix: two-tier rule

1. **Package-tier.** If the change is contained to one of the eight \`packages/*\` directories, the scope MUST be the package name. Mechanically derivable from the PR's changed-files list.

   Allowed: \`agent-auth\`, \`agent-auth-common\`, \`gpg-bridge\`, \`gpg-cli\`, \`things-bridge\`, \`things-cli\`, \`things-client-cli-applescript\`. (And \`gpg-backend-cli-host\` until decommissioned.)

2. **Area-tier.** Otherwise, the scope MUST be one of a small fixed set:

   \`release\`, \`ci\`, \`deps\`, \`deps-dev\`, \`docs\`, \`design\`, \`security\`, \`claude\`.

   Drop everything else: \`(merge-bot)\` becomes \`(ci)\`; \`(release-pr)\` / \`(release-publish)\` / \`(build_release)\` all become \`(release)\`; \`(pr-lint)\` becomes \`(ci)\`. Keep \`(release)\` distinct from \`(ci)\` because release tooling is its own concern with bump-affecting consequences.

3. **Bare scope.** Reserved for genuine cross-cutting changes (a sweeping rename, a repo-wide lint pass) — already covered by CONTRIBUTING.md.

## Out of scope: keep \`chore(release): X.Y.Z\`

The release commit subject should remain \`chore(release): 0.16.2\`, not \`chore: release v0.16.2\`. The \`(release)\` scope is doing useful work — \`git log --grep='chore(release)'\`, the GitHub release page, and the merge-bot's COMMIT_MSG extraction can all key off it. Dropping the scope makes those filters fragile. The version is part of the prose either way; keeping the scope costs ~9 chars and gains regex-friendliness.

## Mechanics

- Update \`pr-lint.yml\` to enforce the matrix. Two new pieces of state:
  - The package list (read from \`packages/\`, no manual list to drift).
  - The fixed area-set (one Python constant, imported by the validator).
- Surface clear error messages:
  > Scope \`(merge-bot)\` is not in the allowlist. The change touches \`.github/workflows/merge-bot.yml\`, which is CI tooling — use \`(ci)\`. See CONTRIBUTING.md → \"Allowed scopes\".

- The package-tier check requires reading the PR's changed-files list. Available via \`gh pr view <N> --json files\` or via the existing checkout.

## Acceptance criteria

- \`pr-lint.yml\` rejects scopes outside the two-tier allowlist with a clear remediation hint.
- A PR touching only \`packages/agent-auth/\` requires \`(agent-auth)\` (and rejects e.g. \`(server)\`).
- A PR touching only \`.github/workflows/\` requires \`(ci)\` (and rejects e.g. \`(merge-bot)\`).
- A PR touching multiple packages can use bare scope or fall back to a representative scope (define behaviour explicitly).
- CONTRIBUTING.md is the source-of-truth doc; the allowlist matches the Python constant.

## Related

Pairs naturally with #401 (type x scope matrix). Both should share the same Python module (#C).

## Origin

Self-review of recent main-branch commit log. The original suggestion was \"package names only\"; this two-tier form is the recommended refinement because it accommodates non-package changes (CI workflows, design docs, agent instructions) without forcing them bare.

aidanns:
==COMMIT_MSG==
Recent main-branch commits drifted past the CONTRIBUTING.md scope
allowlist (`improvement(merge-bot):`, `improvement(pr-lint):`,
`fix(build_release):`, `fix(release-pr):`). The pr-lint type-check
ignores the scope contents, so invented names sail through.

Add a two-tier rule in scripts/validate-pr-title.py:

1. Package tier — when every changed file lives under one
   packages/<name>/, the scope MUST equal <name>.
2. Area tier — otherwise, the scope MUST be in
   commit_taxonomy.AREA_SCOPES (release, ci, deps, deps-dev, docs,
   design, security, claude).
3. Bare scope — always accepted; reserved for cross-cutting
   changes per CONTRIBUTING.md.

Multi-package PRs fall through to area-tier (and therefore can use
bare scope or pick a representative area name) — the conservative
answer per the issue's "define behaviour explicitly" prompt.

The validator gains a `--changed-files-from <path>` flag; the
pr-title-style job in pr-lint.yml fetches the PR's changed-files
list via `gh pr view --json files` and pipes it in. Co-locating the
new step with the existing prose-style validator (rather than
adding a new job) keeps related checks in one place. The trigger
surface stays on `pull_request_target` — `gh` reads PR metadata via
the default GITHUB_TOKEN's `pull-requests: read` permission, no
additional checkout needed.

Error messages quote the offending scope and point at the suggested
replacement: `(merge-bot)` -> `(ci)`, `(release-pr)` -> `(release)`,
`(server)` on a packages/agent-auth/ PR -> `(agent-auth)`. Each
ends with the CONTRIBUTING.md cross-reference so a contributor
landing on the error has a recoverable path.

CONTRIBUTING.md "Allowed scopes" documents the two-tier rule next
to the existing "Type x scope matrix" subsection. The matrix table
prunes `typecheck`, `verify-standards`, `python`, `setup-toolchain`,
`vscode` from its prose enumeration since the two-tier rule now
rejects them outright; commit_taxonomy.INTERNAL_ONLY_SCOPES retains
those names as a belt-and-braces matrix guard. Eight new self-test
fixtures cover the new branches: package-tier accept/reject,
area-tier accept/reject (with both `(merge-bot)` and `(release-pr)`
suggestions), multi-package fallback, and root-level cross-cutting.

Closes #402

==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

Implementation choices resolved in-flight (per the orchestrator's
"resolve internal decisions yourself" rule):

- **Workflow shape**: option (a) from the brief — extend the
  existing `pr-title-style` job with a "List changed files" step
  rather than spinning up a new job. Keeps related checks
  colocated and avoids a fresh `actions/checkout` round-trip; the
  `gh pr view --json files` path doesn't need a checkout at all.
- **Multi-package fallback**: bare scope (matches CONTRIBUTING.md's
  existing "Bare scope is reserved for cross-cutting changes"
  convention without forcing a contributor to invent an area name).
- **CLI shape**: a new `--changed-files-from <path>` flag on the
  validator. The list is read from disk so paths with shell
  metacharacters survive cleanly. When the flag is omitted (e.g.
  the validator's `--self-test` mode or a local invocation), the
  two-tier check is silently skipped — keeps the validator usable
  outside CI without lying about acceptance.
- **Error-message wording**: matches the issue body's worked
  example. The remediation hint includes a suggested replacement
  for the historical drift cases (`merge-bot` / `pr-lint` / `dco` /
  `changelog-bot` / `scorecard` -> `ci`; `release-pr` /
  `release-publish` / `build_release` / `build-release` ->
  `release`); other unknown scopes get a generic "pick from
  AREA_SCOPES" hint.

### Test plan

- [ ] `python3 scripts/validate-pr-title.py --self-test` — 23 cases
      pass locally (pre-existing 15 + 8 new).
- [ ] `task lint` — clean.
- [ ] `task format -- --check` — clean.
- [ ] `task typecheck` — clean.
- [ ] `task changelog:test` — 212 passes.
- [ ] `bash scripts/verify-standards.sh` — clean.
- [ ] CI: `pr-title-self-test` job re-runs the inline cases inside
      the workflow.
- [ ] CI: `pr-title-style` job runs against this PR's actual
      `chore(ci):` title and the multi-area diff (validator,
      workflow, contributing) — area-tier accept path.
- [ ] CI: `pr-title-types-self-test` (taxonomy in-sync check)
      remains green — no `ALLOWED_TYPES` change in this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -307,33 +307,59 @@ stay browsable. Adding a new scope is a CONTRIBUTING edit, not an
 ad-hoc invention in a PR — if none of the scopes below fit, raise a
 PR to add one. The Python lint surface mirrors these lists in
 [`scripts/lint/commit_taxonomy.py`](scripts/lint/commit_taxonomy.py):
-`PACKAGE_SCOPES` is discovered at import time from `packages/*/`, and
-`AREA_SCOPES` currently mirrors a conservative subset of the area
-scopes documented below (the scopes already pinned by existing tests
-and tooling). #402 (the two-tier internal-only scope rule) will
-refine the constant against the full prose set; until it lands,
-expect the constant to be narrower than this section's enumeration.
+`PACKAGE_SCOPES` is discovered at import time from `packages/*/`,
+and `AREA_SCOPES` is the fixed cross-cutting-concern list documented
+below. The two surfaces are kept in lockstep by the
+`pr-title-types-self-test` job and by code review against this
+section.
 
-- **Subsystem scopes** — the module or surface the commit touches.
-  `agent-auth`, `things-bridge`, `things-cli`,
-  `things-client-cli-applescript`, `server`, `audit`, `store`,
-  `keys`, `scopes`, `rate-limit`, `tls`, `notifier`, `metrics`, `api`,
-  `benchmark`, `observability`.
-- **Area scopes** — cross-cutting concerns.
-  `release` (release automation, CHANGELOG, versioning), `ci` (CI
-  workflow / composite action changes), `deps` (runtime dep bumps),
+The validator at
+[`scripts/validate-pr-title.py`](scripts/validate-pr-title.py)
+applies a **two-tier rule** (#402) when the PR's changed-files list
+is available:
+
+1. **Package tier.** If every changed file lives under a single
+   `packages/<name>/` directory, the scope MUST be `(<name>)` —
+   `feature(server):` on a PR contained to `packages/agent-auth/`
+   is rejected with a remediation hint pointing at `(agent-auth)`.
+   Mechanically derived from the diff, so adding a new package and
+   using its name as the scope works on day one.
+2. **Area tier.** Otherwise (the diff spans multiple packages, or
+   touches files outside `packages/`), the scope MUST be in
+   `AREA_SCOPES`. Historical drift like `(merge-bot)`, `(pr-lint)`,
+   `(release-pr)`, `(release-publish)`, `(build_release)` is
+   rejected; the validator suggests `(ci)` for workflow-tweak
+   scopes and `(release)` for release-tooling scopes.
+3. **Bare scope** (no `(scope)`) is always accepted — reserved for
+   genuinely cross-cutting changes (a sweeping rename, a repo-wide
+   lint pass). Multi-package PRs that don't fit a single area name
+   should go bare too: it is the conservative answer per #402's
+   "define behaviour explicitly" rule.
+
+- **Package scopes** (the package-tier branch of the rule) — the
+  workspace package the commit is contained to. Auto-discovered
+  from `packages/*/` so adding a new package never drifts from the
+  lint: `agent-auth`, `agent-auth-common`, `gpg-bridge`, `gpg-cli`,
+  `things-bridge`, `things-cli`, `things-client-cli-applescript`.
+- **Area scopes** (the area-tier branch of the rule) — cross-cutting
+  concerns. The fixed set is `release` (release automation,
+  CHANGELOG, versioning), `ci` (CI workflow / composite action
+  changes — covers what historical drift called `(merge-bot)`,
+  `(pr-lint)`, `(dco)`, etc.), `deps` (runtime dep bumps),
   `deps-dev` (dev dep bumps), `docs` (README, CONTRIBUTING,
   CLAUDE.md, SECURITY.md, SUPPORT.md), `design` (`design/**`, ADRs),
-  `security` (SECURITY.md content, security controls, scorecard, dco,
-  supply-chain), `typecheck` (mypy / pyright), `verify-standards`
-  (`scripts/verify-standards.sh`), `python` (ruff, pytest, project
-  Python config), `setup-toolchain` (`.github/actions/setup-toolchain`),
-  `docker` (Dockerfiles, compose, image pipeline), `vscode`
-  (`.vscode/`), `claude` (CLAUDE.md / `.claude/instructions/`).
+  `security` (SECURITY.md content, security controls, scorecard,
+  dco, supply-chain), `claude` (CLAUDE.md / `.claude/instructions/`).
+  Other internal-only names that the type-x-scope matrix references
+  (`typecheck`, `verify-standards`, `python`, `setup-toolchain`,
+  `vscode`) are not first-class area scopes for the two-tier rule —
+  fold those changes under the closest area name (typically `(ci)`
+  for tooling wired into the workflow, or `(deps-dev)` for the
+  underlying dependency bump).
 
 Bare (no-scope) commits are reserved for changes that genuinely
-don't fit a single scope — a sweeping rename, a repo-wide lint pass.
-Prefer a scope when one applies.
+don't fit a single scope — a sweeping rename, a repo-wide lint pass,
+or a multi-package PR. Prefer a scope when one applies.
 
 #### Type × scope matrix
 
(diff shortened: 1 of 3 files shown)


scripts/validate-pr-title.py
```diff
@@ -228,76 +371,94 @@ def validate(title: str) -> None:
...
     (
         # Package scope — release-bumping types are fine here; only
         # the internal-only scopes are restricted.
         "feature(agent-auth): add JIT approval flow for prompt-tier scope",
+        None,
         True,
         "passing-feature-package-scope",
     ),
     (
         "fix(tokens): use constant-time HMAC comparison",
```
aidanns: Claude Reviewer: This passing self-test (`fix(tokens):`) keeps `changed_files=None` so the new two-tier check is skipped — the test still passes as-is. But if this exact title were used on a real PR (with a real `gh pr view` output), the new rule would reject it: there is no `packages/tokens/`, and `tokens` is not in `AREA_SCOPES`. The fixture has become a fiction relative to CI behaviour. Either provide a representative `changed_files` (e.g. `["packages/agent-auth/src/.../tokens.py"]` plus rename the title to `fix(agent-auth):`), or drop the fixture. As-is, a contributor reading this passing case as "valid PR title" will be misled.

scripts/validate-pr-title.py
```diff
@@ -312,37 +473,102 @@ def validate(title: str) -> None:
...
+    (
+        # Multi-package fallback: PR spans two packages -> falls
+        # through to area-tier; bare scope is accepted.
+        "chore: bump shared HTTP client default timeout",
+        [
+            "packages/agent-auth/src/agent_auth/http.py",
+            "packages/gpg-bridge/src/gpg_bridge/http.py",
+        ],
+        True,
+        "passing-multi-package-bare-scope",
```
aidanns, line 585: Claude Reviewer: Missing fixture for the multi-package + scoped-and-rejected case. The dispatch / issue body explicitly calls out that `feature(agent-auth):` on a PR touching both `packages/agent-auth/` AND `packages/gpg-bridge/` should be REJECTED (it is no longer "contained to agent-auth", and `agent-auth` is not in `AREA_SCOPES`). Today the only multi-package fixture is `passing-multi-package-bare-scope` (line 555). Adding the rejected counterpart guards against a regression where someone "fixes" the multi-package fall-through by accidentally accepting any package-named scope.

scripts/validate-pr-title.py
```diff
@@ -218,7 +229,139 @@ def check_type_scope_matrix(title: str) -> None:
...
 
-def validate(title: str) -> None:
+def _single_package_for(changed_files: list[str]) -> str | None:
+    """Return the package name if every changed file lives in one
+    ``packages/<name>/`` directory, else ``None``.
+
+    Path normalisation: ``gh pr view --json files --jq '.files[].path'``
+    yields forward-slash POSIX paths regardless of runner OS, so we
+    split on ``/`` rather than ``os.sep``. Empty / whitespace-only
+    lines (the ``mktemp`` rounding error from a no-op PR) are skipped
```
aidanns: Claude Reviewer: The phrase "the `mktemp` rounding error from a no-op PR" does not describe what is happening. `mktemp` is not used on this code path (the workflow writes to `${RUNNER_TEMP}/changed-files.txt`), and "rounding error" is the wrong term for a trailing-newline edge case. Suggest rewording to something like: "Empty / whitespace-only lines are skipped so a trailing newline in the file written by `gh pr view` does not collapse the check." A future maintainer searching for `mktemp` to understand this comment will land on the wrong code path.

CONTRIBUTING.md
```diff
@@ -347,13 +373,24 @@ workflow tweak would surface in `CHANGELOG.md` as a feature and bump
...
+| **Internal-only scopes:** `ci`, `deps-dev`, `claude`, `design`, `docs`          | `chore`, `fix` only                                                           |
+
+`commit_taxonomy.INTERNAL_ONLY_SCOPES` carries a few extra names
+(`typecheck`, `verify-standards`, `python`, `setup-toolchain`,
+`vscode`) inherited from #401 for the matrix check. After #402,
+those names are no longer first-class area scopes — the two-tier
+rule above rejects them. Pick `(ci)` for tooling wired into the
+workflow, `(deps-dev)` for the underlying dev-dep bump, or drop the
+scope. The matrix entry for those names is preserved as a
+belt-and-braces guard so a contributor who tries them gets the
```
aidanns, line 391: Claude Reviewer: The "belt-and-braces" exception keeps `typecheck`, `verify-standards`, `python`, `setup-toolchain`, `vscode` in the matrix-table prose AND in `commit_taxonomy.INTERNAL_ONLY_SCOPES`, but the new two-tier rule rejects them outright (none are in `AREA_SCOPES`, none are package names). The end result for a contributor is confusing: `chore(python):` survives the matrix check (because `chore` is allowed on internal-only scopes) but then fails the two-tier check with a generic "pick from AREA_SCOPES" message that does not mention `python` was a known internal-only name. Either drop these names from the matrix row in this PR (since they are now unconditionally rejected by the two-tier rule and the matrix entry serves no purpose for them), or extend `_area_scope_suggestion` to map them to a useful target (e.g. `python` -> `deps-dev` or `ci`, depending on what the contributor was actually doing). The current state preserves the contradiction and pushes resolution to "the same issue thread" — better to resolve it now while the PR is in flight.

scripts/validate-pr-title.py
```diff
@@ -218,7 +229,139 @@ def check_type_scope_matrix(title: str) -> None:
...
+            return None
+        candidate = parts[1]
+        if package is None:
+            package = candidate
+        elif package != candidate:
+            return None
+    if package is None:
+        return None
+    if package not in PACKAGE_SCOPES:
+        return None
```
aidanns: Claude Reviewer: When `_single_package_for` discovers a single package but the name is not in `PACKAGE_SCOPES`, it returns `None` and the validator falls through to the area-tier branch. A contributor in this situation (presumably a stale workspace snapshot or a brand-new package added in the same PR before `PACKAGE_SCOPES` is reseeded) will get the generic "scope is not in the allowlist; pick from AREA_SCOPES" hint, which is wrong — the issue is that the package itself was not recognised. Either raise a distinct, explanatory error here ("scope `(<X>)` matches the contained directory `packages/<X>/` but `<X>` is not in `PACKAGE_SCOPES`; if you just added the package, rebase to refresh the auto-discovery"), or accept the package-tier match unconditionally and let the area-tier suggestion logic stay the fallback. The defensive-None path silently swaps useful diagnostics for a misleading hint.

scripts/validate-pr-title.py
aidanns: Claude: Done!

scripts/validate-pr-title.py
aidanns: Claude: Done!

scripts/validate-pr-title.py
aidanns: Claude: Done!

CONTRIBUTING.md
aidanns: Claude: Picked option (b) — extended `_area_scope_suggestion` to map all five matrix-only names (`typecheck`, `verify-standards`, `python`, `setup-toolchain`, `vscode`) to `(ci)`. They all name tooling/config surfaces wired into CI, so a uniform mapping keeps the suggestion table tight (~5 entries) and matches the rationale in `commit_taxonomy.py`. The CONTRIBUTING.md belt-and-braces note still stands; a contributor adjusting a dev-dep pin underneath one of those names should pick `(deps-dev)` by hand.

scripts/validate-pr-title.py
aidanns: Claude: Done!

Issue #401, opened 2026-04-27
lint: forbid feature/improvement on internal-only scopes (ci, deps-dev, claude, ...)


aidanns:
## Problem

Internal repo plumbing (CI, lint, dev-deps, editor configs, agent instructions) is currently being labelled \`feature(…)\` or \`improvement(…)\`, which:

1. Pollutes \`CHANGELOG.md\` and the GitHub release notes with churn that doesn't ship to users.
2. Triggers the wrong release-impact bump in \`scripts/changelog/version_logic.py\`, which derives bumps from the \`type:\` field. A \`feature(ci):\` PR triggers a minor version bump for an internal-only change.

## Examples

- [\`166f55c\`](https://github.com/aidanns/agent-auth/commit/166f55cf445e16c7a8eb99ba35250672f2b66e76) — \`feature(ci): add release-build dry-run …\`. Pure CI plumbing. Surfaced in 0.15.0 release notes as a Feature and contributed to the minor bump.
- [\`b07fe58\`](https://github.com/aidanns/agent-auth/commit/b07fe588ca7eda6a45be27bbb9aef566ff027865) — \`improvement(ci): wire release-publish.yml to release-bot App\`. Bot identity refactor, surfaced in 0.14.0.

Compare to [\`61aa0d7\`](https://github.com/aidanns/agent-auth/commit/61aa0d7) — \`chore(ci): fail-loudly on missing bot App secrets\`. Same release, correctly labelled \`chore\` and (correctly) absent from the release notes.

## Proposed fix

Add a lint rule (in \`scripts/validate-pr-title.py\` or as a new \`pr-lint.yml\` job) that constrains the \`type(scope)\` combination matrix:

| Scope | Allowed types |
|---|---|
| Subsystem scopes (per CONTRIBUTING.md \"Subsystem scopes\") | \`feature\`, \`improvement\`, \`fix\`, \`chore\`, \`deprecation\`, \`migration\`, \`break\` |
| \`release\` | \`feature\`, \`improvement\`, \`fix\`, \`chore\` |
| **Internal-only scopes:** \`ci\`, \`deps-dev\`, \`typecheck\`, \`verify-standards\`, \`python\`, \`setup-toolchain\`, \`vscode\`, \`claude\`, \`design\`, \`docs\` | **\`chore\`, \`fix\` only** |
| \`deps\` (runtime) | \`feature\`, \`improvement\`, \`fix\`, \`chore\` (runtime dep bump can be release-worthy) |
[...]

PR #418, opened 2026-04-27
chore(ci): add release-impact-comment job to pr-lint

### LINKED ISSUE #406: pr-lint: surface predicted release-impact bump as a PR comment
## Problem

Mislabelled changelog entries (e.g. a CI plumbing change with \`type: feature\`) silently trigger the wrong public version bump. The mistake is invisible at PR review time — it surfaces only when the next release goes out and the version jumps unexpectedly.

#401 (type×scope matrix) catches the worst category of this (forbids \`feature(ci):\` etc). But the type itself is still author-controlled in \`changelog/@unreleased/*.yml\`, and a contributor can label a CI-adjacent-but-package-touching change as \`feature\` when \`improvement\` or \`fix\` is more appropriate. The matrix lint won't catch that — only a human reviewer will.

## Proposed fix

A PR-time bot comment that surfaces the *predicted release-impact bump*:

> **Release impact:** This PR will trigger a **MINOR** version bump because of \`type: feature\` in \`changelog/@unreleased/pr-NNN-slug.yml\`.

Updated whenever the changelog YAML changes. Computed by importing \`scripts/changelog/version_logic.py\` and feeding it the unreleased YAMLs that would be present after this PR merges.

## Mechanics

- New job in \`pr-lint.yml\`: \`release-impact-comment\`.
- Reads the PR's \`changelog/@unreleased/*.yml\` (existing entries + the one this PR adds).
- Calls \`version_logic.compute_bump(...)\` to get the predicted bump.
- Posts (or edits, idempotently) a comment via the \`agent-auth-merge-bot\` App or a dedicated impact-bot identity.
- Comment includes the next predicted version (\`current + bump\`) so the reviewer sees \`v0.16.0\` vs \`v0.15.4\`.

## Catches that this surfaces

- \`feature(server)\` for a tiny refactor — the comment shows MINOR; reviewer asks \"should this be \`improvement\` or \`fix\`?\"
- The first \`feature:\` PR in a quiet release cycle — surfaces \"this is the bump driver\" so the reviewer knows it's load-bearing.
- A \`break:\`-prefixed PR — surfaces the MAJOR bump, prompting the reviewer to make sure the \`BREAKING CHANGE:\` trailer is present and the migration notes are written.

## Acceptance criteria

- A new comment appears (or updates) on every PR that touches \`changelog/@unreleased/\`.
- The comment shows the predicted \`current → next\` version transition.
- The comment is idempotent — re-running the workflow does not stack duplicate comments.
- The bot identity matches the existing convention (App-based, prefixed \"Claude: \" per the project CLAUDE.md if it's the Claude-mediated bot).

## Out of scope

Changing the bump computation itself. This is a surfacing/communication issue, not a logic issue. \`version_logic.py\` is already authoritative.

## Origin

Self-review of recent main-branch commit log; defensive complement to #401 (catches mislabels that the matrix can't reject mechanically).

aidanns:
==COMMIT_MSG==
chore(ci): add release-impact-comment job to pr-lint

Mislabelled changelog YAML entries silently trigger the wrong
public version bump. The matrix lint (#401) catches the worst
category -- `feature(ci):` and friends -- but the YAML's `type:`
field is still author-controlled and a contributor can label a
CI-adjacent change as `feature` when `improvement` or `fix` is
more appropriate. The matrix won't catch that; only a human
reviewer will.

Adds a `release-impact-comment` job to `pr-lint.yml` that posts
an idempotent `Claude:` comment on every PR touching
`changelog/@unreleased/*.yml`, surfacing the predicted
`current -> next` version transition so a mislabel is visible at
PR-review time. The comment uses an HTML-comment marker
(`<!-- claude-release-impact -->`) for idempotent matching and
is PATCHed in place on subsequent runs to avoid notification
noise. PRs that touch no `@unreleased/*.yml` files are skipped
entirely so the bot does not post on every PR.

Bump computation is a thin wrapper over the existing
`build_release.compute_release(...)` so the prediction surfaced
on the PR is byte-identical to the plan the release workflow
will execute. The new
`scripts/changelog/predict_release_impact.py` owns the wrapper
and the JSON envelope shape; `predict-release-impact-self-test`
exercises the bump matrix on every run.

The job runs under the existing `pull_request_target` trigger
(not widening the trigger surface). The PR head's YAMLs are
fetched as data only -- the predict script and the
`version_logic` parser both come from the base ref, so a fork
PR cannot ship a malicious predictor in the same PR. The
comment is posted via the default `GITHUB_TOKEN` with
`pull-requests: write` at the job level; no additional bot-token
plumbing is required.

Closes #406
==COMMIT_MSG==

## Review notes

This PR is plumbing -- the new comment IS user-visible to PR
contributors but is internal tooling, not a runtime behaviour
change. Apply `no changelog` after opening per the project
convention for `chore(ci):` PRs.

### Bot identity

Posts as `github-actions[bot]` via the default `GITHUB_TOKEN`
(`pull-requests: write` at the job level). The merge-bot /
release-bot / changelog-bot App path was considered for symmetry,
but minting an App token requires provisioning new secrets and
the comment surface here is purely informational (no
protected-branch or `contents:` writes). Using the default token
keeps the bot fully operational on the PR that initially adds
the workflow without depending on a separate one-time setup.
The `Claude:` prefix per the project convention still identifies
the comment as agent-authored. Documented inline in the workflow
comment block.

### Test plan

- [ ] `scripts/changelog/tests/test_predict_release_impact.py` exercises the bump matrix (`feature` → MINOR, `improvement` / `fix` / `deprecation` / `migration` → PATCH, `break` (in 0.x) → MINOR, `break` (in 1.x) → MAJOR, mixed → highest, empty `@unreleased/` → no impact, malformed YAML → `ChangelogValidationError`). Run via `uv run pytest scripts/changelog/tests/test_predict_release_impact.py -v`.
- [ ] All 227 existing changelog tests still pass.
- [ ] `task test` (full suite, 753 tests) passes.
- [ ] `task lint`, `task typecheck`, `python3 scripts/lint/commit_taxonomy.py check-pr-lint-yaml`, `python3 scripts/validate-pr-title.py --self-test`, `bash scripts/verify-standards.sh` all pass.
- [ ] On this PR itself, the new `release-impact-comment` job runs and EARLY-EXITs (this PR doesn't touch `changelog/@unreleased/*.yml`, so no comment is posted). The job log line `release-impact-comment: no changelog/@unreleased/ files touched; nothing to predict.` confirms the no-noise branch.
- [ ] The `predict-release-impact-self-test` job runs and passes.

### Out of scope

- Changing the bump computation itself. This is a surfacing / communication issue, not a logic issue. `version_logic.py` (and its `build_release.compute_release` wrapper) is already authoritative.

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -491,13 +491,18 @@ jobs:
           # filters in a single pass so we don't depend on a particular
           # gh version's output shape. `--arg` keeps the marker out of
           # the jq program text so its `<`, `!`, `-` characters survive
-          # shell quoting verbatim.
+          # shell quoting verbatim. `first(...)` (rather than a trailing
+          # `head -n 1`) caps the result inside jq itself: under
+          # `set -euo pipefail`, closing the pipe early via `head` can
+          # deliver SIGPIPE upstream and fail the pipeline on
+          # long-lived PRs whose comment list spans multiple pages.
+          # `// empty` keeps the assignment empty (rather than emitting
+          # the literal string "null") when no matching comment exists.
           existing_id="$(gh api \
             "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" \
             --paginate \
             | jq -r --arg marker "${MARKER}" \
-                '.[] | select((.body // "") | startswith($marker)) | .id' \
-            | head -n 1)"
+                'first(.[] | select((.body // "") | startswith($marker)) | .id) // empty')"
           # `-F body=@<file>` reads the body from disk as a JSON
           # string field — identical semantics to inlining
           # `body="$(cat <file>)"` but binary-safe and free of any
@@ -545,4 +550,12 @@ jobs:
           # `conftest.py` under scripts/changelog/tests/ puts
           # scripts/changelog/ and scripts/lint/ on sys.path so the
           # bare-name imports in the test file resolve.
-          python3 -m pytest scripts/changelog/tests/test_predict_release_impact.py -v
+          # `-o addopts=` overrides the workspace `pyproject.toml`'s
+          # `[tool.pytest.ini_options].addopts` (which injects
+          # `--cov=packages --cov-branch --cov-report=term-missing`).
+          # Without this override, bare pytest would reject those
+          # flags with `error: unrecognized arguments` and exit 4
+          # because this isolated job intentionally does not install
+          # `pytest-cov`; the job exercises one script in isolation,
+          # not the workspace's full coverage matrix.
+          python3 -m pytest -o addopts= scripts/changelog/tests/test_predict_release_impact.py -v
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -363,33 +363,51 @@ jobs:
           fi
 
       - name: Fetch PR head's @unreleased YAMLs
-        # Bring the head ref's `changelog/@unreleased/*.yml` files into
-        # the workspace WITHOUT checking out the head's code. The
-        # predict script + the `version_logic` parser both come from
-        # the base ref (pinned by the earlier `actions/checkout`); only
-        # the YAML data files come from the head. PyYAML's `safe_load`
-        # parses YAML as data — no code-execution surface — so this
-        # is safe even on a fork PR.
+        # Materialise the head ref's `changelog/@unreleased/*.yml` into
+        # a sandbox under `${RUNNER_TEMP}` — NOT into the workspace.
+        # The predict script + `version_logic` parser both come from
+        # the base ref (pinned by the earlier `actions/checkout`); the
+        # YAML data files are read from the sandbox by passing
+        # `--repo-root ${RUNNER_TEMP}/predict-sandbox` to the predict
+        # invocation. PyYAML's `safe_load` parses YAML as data — no
+        # code-execution surface — so this is safe even on a fork PR.
+        #
+        # We deliberately AVOID `git checkout "${HEAD_SHA}" -- path/`
+        # here. CodeQL's `actions/untrusted-checkout` query treats any
+        # `git checkout <head-sha>` in a `pull_request_target` workflow
+        # as a code-execution surface, regardless of the `-- <path>`
+        # constraint that scopes it to data-only files. Routing through
+        # the GitHub Contents API keeps the head fetch off CodeQL's
+        # flagged channel and onto the unflagged HTTP-fetch channel.
         if: steps.touched.outputs.touched == 'true'
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
           set -euo pipefail
-          # `pull_request_target` checkout pinned to the base SHA does
-          # not fetch the head SHA into the local clone. Fetch it
-          # explicitly, then materialise just the @unreleased dir from
-          # that commit. `--no-tags --depth=1` keeps the fetch tiny.
-          git fetch --no-tags --depth=1 origin "${HEAD_SHA}"
-          # Wipe whatever `@unreleased/` looks like at the base ref
-          # before overlaying the head's version, so files removed by
-          # the PR (e.g. the release-PR's mass rename) don't linger.
-          rm -rf changelog/@unreleased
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-lint.yml
```diff
@@ -266,3 +266,283 @@ jobs:
...
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "3.11"
+
+      - name: Install runtime deps (PyYAML, pytest)
+        run: pip install --no-cache-dir 'pyyaml>=6' 'pytest>=8'
```
aidanns, line 582: Claude Reviewer: **BLOCKING:** the `predict-release-impact-self-test` job will fail in CI as written. The install step only installs `pyyaml` and `pytest`, but the workspace `pyproject.toml` `[tool.pytest.ini_options].addopts` includes `--cov=packages --cov-branch --cov-report=term-missing`. Without `pytest-cov` installed, bare pytest rejects those flags with `error: unrecognized arguments: --cov=packages --cov-branch --cov-report=term-missing` and exits 4 before collecting any tests.

Reproduced locally with the exact CI command:

```
$ python3 -m venv /tmp/v && /tmp/v/bin/pip install --no-cache-dir "pyyaml>=6" "pytest>=8"
$ /tmp/v/bin/python -m pytest scripts/changelog/tests/test_predict_release_impact.py -v
ERROR: usage: __main__.py [options] [file_or_dir] ...
__main__.py: error: unrecognized arguments: --cov=packages --cov-branch --cov-report=term-missing
  inifile: /tmp/.../pyproject.toml
$ echo $?
4
```

Three options to fix, in increasing order of robustness:

1. Add `pytest-cov` to the install line — minimal change but you also need `tests_support.integration.plugin` resolvable via the workspace `pythonpath`, which it is, so this should be enough.
2. Override addopts at invocation time with `python3 -m pytest -o addopts= scripts/changelog/tests/test_predict_release_impact.py -v` to bypass the workspace-level pytest config entirely. Cleanest in spirit since the job is meant to test only the predict script in isolation.
3. Run via `uv run pytest ...` like every other test invocation in the repo. Heavier (full `uv sync`) but most consistent with how other workflows shape the test step.

Option 2 is the smallest change and most accurately reflects what the job is doing (running an isolated unit test of one script, not the workspace test suite). Recommend that.

Note that the existing `validator-self-test` job in this same workflow does NOT hit this because it uses bash, not pytest.

.github/workflows/pr-lint.yml
```diff
@@ -266,3 +266,283 @@ jobs:
...
+          # filters in a single pass so we don't depend on a particular
+          # gh version's output shape. `--arg` keeps the marker out of
+          # the jq program text so its `<`, `!`, `-` characters survive
+          # shell quoting verbatim.
+          existing_id="$(gh api \
+            "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" \
+            --paginate \
+            | jq -r --arg marker "${MARKER}" \
+                '.[] | select((.body // "") | startswith($marker)) | .id' \
+            | head -n 1)"
```
aidanns: Claude Reviewer: minor robustness concern. `gh api ... --paginate | jq ... | head -n 1` runs under `set -euo pipefail`. When `head -n 1` finds a match on the first line and closes the pipe, jq (and possibly gh) get SIGPIPE. With `pipefail`, an upstream non-zero exit fails the whole pipeline. This will not normally trigger on a small PR (single page, few comments), but on a long-lived PR with many comments it can become flaky.

Two cheap mitigations:

- Drop `head -n 1` and let jq do the limiting: `... | jq -r first(.[] | select((.body // \"\") | startswith($marker)) | .id) // empty`. jq emits at most one line and exits cleanly.
- Or: capture into a temp file and slice with shell: `gh api ... --paginate > \"${RUNNER_TEMP}/comments.json\"; existing_id=\"$(jq -r ... < ...)\"`.

Not blocking, but worth a one-line tweak while the surrounding code is being touched.

.github/workflows/pr-lint.yml
aidanns: Claude: Done!

.github/workflows/pr-lint.yml
aidanns: Claude: Done!

PR #419, opened 2026-04-27
chore(ci): enforce trailer-block contiguity in commit-msg validator

### LINKED ISSUE #400: commit-msg: trailer-block contiguity (Closes / Signed-off-by must not be split)
## Problem

There is inconsistent visual treatment of the trailer block in commit messages. Some commits separate \`Closes #N\` and \`Signed-off-by:\` with a blank line for visual breathing room — but **a blank line between trailers breaks \`git interpret-trailers --parse\`**, which treats the blank as the body/trailer boundary and then sees only the lines after it as trailers.

## Example

[\`c3c7136\`](https://github.com/aidanns/agent-auth/commit/c3c71364139805a6de0e3da95e972d371446ee46) ends:

\`\`\`
Closes #338
\`\`\`

…with no blank line. That's actually **correct** kernel-style — the trailer block must be contiguous.

## Why the rule exists

RFC 5322 / kernel convention / cbea.ms all say trailers form one contiguous block. \`git interpret-trailers --parse\` will not see \`Closes:\` as a trailer if a blank line precedes \`Signed-off-by:\`. Tooling downstream of \`git interpret-trailers\` (release-note generators, audit-trail extractors, GitHub's own \"linked issues\" inference for some shapes) silently misses the \`Closes:\` reference.

## Proposed fix

Document and enforce **two rules**, not one:

1. **Trailer block contiguity.** No blank lines *between* trailers. \`Closes:\`, \`Co-authored-by:\`, \`Signed-off-by:\` etc. all stack with no blanks.
2. **One blank line *before* the trailer block.** That's where the visual separation goes. The body ends, blank line, then the contiguous trailer stack.

So the canonical shape is:

\`\`\`
<body paragraph>

Closes #NNN
Signed-off-by: <name> <email>
\`\`\`

## Implementation

- \`scripts/validate-commit-msg-block.py\` already partially enforces this (the \`SIGNOFF_RE\` and trailer parsing). Extend it to:
  - Require **at least one blank line** between the last body paragraph and the first trailer.
  - **Reject** a blank line between any two consecutive trailers.
- Update \`CONTRIBUTING.md\` → \"Trailers\" subsection to document both rules with a worked example.

## Acceptance criteria

- The validator rejects blank lines between trailers with a clear error message and remediation hint.
- The validator requires a blank line before the trailer block.
- Existing PR-template fixtures and the inline self-test loop in \`pr-lint.yml\` cover both new cases.
- A representative sample of recent commits (\`c3c7136\`, \`b07fe58\`, \`166f55c\`) continues to pass.

## Origin

Self-review of recent main-branch commit log. The original suggestion (add blank line between Closes and Signed-off-by) was inverted in review because it would break \`git interpret-trailers\`.

aidanns:
==COMMIT_MSG==
chore(ci): enforce trailer-block contiguity in commit-msg validator

Two new layout rules in scripts/validate-commit-msg-block.py keep the
==COMMIT_MSG== block aligned with `git interpret-trailers --parse`'s
contract: no blank lines between two consecutive trailers (the parser
treats the first blank line above a candidate trailer as the
body/trailer boundary, so a `Closes #N` separated from
`Signed-off-by:` by a blank silently drops out of the trailer set),
and one blank line between the body and the first trailer (the visual
separator goes there, not between trailers).

The trailer-shape predicate is gated on KNOWN_TRAILER_TOKENS rather
than any `[A-Za-z0-9-]+:` shape so a stray body line like
`Bug-123: see ticket` does not extend the trailer-block region across
a body/trailer blank and produce a misleading "blank between
trailers" diagnostic. `check_trailers` (the unknown-token gate) runs
before the new layout checks so a `Cosed:` typo still surfaces as the
more actionable "unknown trailer token" error.

Coverage: two new fixtures
(invalid-blank-between-trailers.md, invalid-no-blank-before-trailers.md)
plug into the existing fixture loop in pr-lint.yml's
validator-self-test job, so both new failure paths exercise on every
PR. CONTRIBUTING.md gains an explicit "two layout rules" subsection
under "Trailers" with the canonical shape and the
`git interpret-trailers --parse` rationale; the CI-vs-convention
table grows two new rows. The PR template's inline help calls out
the contiguity rule alongside the existing trailer-format rule.

Closes #400
==COMMIT_MSG==

==NO_CHANGELOG==

### Test plan

- [ ] `python3 scripts/validate-commit-msg-block.py .github/workflows/tests/pr-lint-fixtures/invalid-blank-between-trailers.md` exits 1 with the contiguity error.
- [ ] `python3 scripts/validate-commit-msg-block.py .github/workflows/tests/pr-lint-fixtures/invalid-no-blank-before-trailers.md` exits 1 with the blank-line-before error.
- [ ] Every existing `valid-*.md` fixture continues to pass (the canonical kernel-style shape — body, blank, contiguous trailers — was already the convention; nothing valid regresses).
- [ ] `python3 scripts/validate-commit-msg-block.py --self-test` still passes the title-aware self-test cases.
- [ ] CI's `validator-self-test` job iterates every fixture; the two new fixtures plug straight into the loop without workflow changes.

### Implementation notes

- `_is_trailer_shape_line` uses `KNOWN_TRAILER_TOKENS` rather than the bare `TRAILER_RE` shape so a body line like `Bug-123: see ticket` does not falsely extend the trailer-block region across a body/trailer blank. The cost is that an unknown-token typo (`Cosed: #1`) is no longer trailer-shape, which would swap the failure mode under the new layout checks for the misleading "no blank line before trailer block". `check_trailers` is therefore reordered to run *before* the two new checks so the `Cosed:`-style typo still surfaces as the more actionable "unknown trailer token" error.
- The `_trailer_block_start_index` helper deliberately walks across blank lines so the broken `Closes #N` / blank / `Signed-off-by:` pattern is captured as a single (broken) trailer block; the contiguity check then fails on any blank inside the captured region. Without that walk-across-blanks step the broken pattern would be classified as a body line and a separate trailer block, hiding the bug the validator is meant to catch.
- An issue-body discrepancy worth flagging: the issue's acceptance criteria says `c3c7136`, `b07fe58`, and `166f55c` "continue to pass". `c3c7136` (the canonical positive example in the issue body) does pass under the new rules. `b07fe58` and `166f55c` *do not* — they are exactly the broken pattern this PR is closing (a blank line between `Closes #N` and `Signed-off-by:`), and `git log --format=%B` confirms it. They are commits already merged on `main`; the validator only runs against PR bodies, so the merged history is unaffected. Treating "continues to pass" as referring only to the canonical sample (`c3c7136`) reads as the consistent interpretation of the issue.

### DIFF
--- a/scripts/validate-commit-msg-block.py
+++ b/scripts/validate-commit-msg-block.py
@@ -102,6 +102,22 @@ GITHUB_KEYWORD_RE = re.compile(
     r"^(Closes|Fixes|Resolves)\s+(#\d+|[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+#\d+)\.?$"
 )
 
+# Kernel-style `Fixes: <sha> ...` trailer line — the FULL line including
+# the `Fixes:` prefix and a 7-40 char hex SHA value (CONTRIBUTING.md →
+# "Trailers" documents this shape). This tighter check distinguishes a
+# real `Fixes:` trailer pointing at an introducing commit from a body
+# paragraph whose heading happens to be `Fixes:` — e.g. the release-PR
+# body's `Fixes: <prose description>.` section heading rendered by
+# `scripts/changelog/build_release.py:render_commit_msg_block`. Without
+# the value-shape distinction, `_is_trailer_shape_line` would absorb
+# the section heading into the trailer-block region and
+# `check_trailer_block_contiguity` / `check_blank_line_before_trailers`
+# would fire on an otherwise well-formed body. (Distinct from
+# ``FIXES_SHA_TRAILER_RE`` defined further below, which matches only
+# the value portion as part of `check_fixes_trailer_shape`'s shape
+# check on already-parsed trailer values.)
+FIXES_SHA_TRAILER_LINE_RE = re.compile(r"^Fixes:[ \t]+[0-9a-fA-F]{7,40}\b")
+
 # Recognised trailer tokens in this project. Other tokens parse as
 # trailers structurally but warrant a stricter check (we want to fail
 # closed on typos like `Cosed: #1` — see is_trailer_token).
@@ -318,6 +334,17 @@ def _is_trailer_shape_line(line: str) -> bool:
     """
     trailer_match = TRAILER_RE.match(line)
     if trailer_match is not None and is_trailer_token(trailer_match.group(1)):
+        # `Fixes:` only counts as a trailer when its value is a SHA
+        # (kernel `Fixes: <sha> ("subject")` form). When the value is
+        # prose, the line is a body paragraph whose heading happens to
+        # be `Fixes:` (e.g. release-PR body sections rendered by
+        # `build_release.render_commit_msg_block`); treating it as a
+        # trailer would absorb the body paragraph into the trailer-block
+        # region and trigger false-positive contiguity / blank-before
+        # failures. The `Fixes #N` GitHub-keyword form is unaffected —
+        # it falls through to the `GITHUB_KEYWORD_RE` branch below.
+        if trailer_match.group(1).lower() == "fixes":
+            return FIXES_SHA_TRAILER_LINE_RE.match(line) is not None
         return True
     if GITHUB_KEYWORD_RE.match(line) is not None:
         return True
--- a/scripts/validate-commit-msg-block.py
+++ b/scripts/validate-commit-msg-block.py
@@ -43,10 +43,12 @@
 #      `Signed-off-by:` makes it see only the latter as a trailer, so
 #      release-note generators and audit-trail extractors silently
 #      lose the `Closes:` reference.
-#  10. Exactly one blank line sits between the last body paragraph and
-#      the first trailer line. The body ends, blank line, then the
-#      contiguous trailer stack — that blank is where the visual
-#      separation lives, not between trailers.
+#  10. At least one blank line sits between the last body paragraph
+#      and the first trailer line. The body ends, blank line(s),
+#      then the contiguous trailer stack — that blank is where the
+#      visual separation lives, not between trailers. This matches
+#      `git interpret-trailers --parse` semantics, which treats any
+#      run of one or more blanks as the body/trailer boundary.
 #
 # The PR title (subject) has its own prose-style rules — length cap,
 # trailing period, past-tense imperative — enforced by the sibling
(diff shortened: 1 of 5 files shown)


scripts/validate-commit-msg-block.py
```diff
@@ -308,6 +387,81 @@ def parse_trailer_block(lines: list[str]) -> list[tuple[int, str, str]]:
...
+        "contiguous — `git interpret-trailers --parse` treats a blank "
+        "line as the body/trailer boundary, so a `Closes #N` "
+        "separated from `Signed-off-by:` by a blank drops out of the "
+        "trailer set and downstream tooling silently loses the "
+        "reference. Stack the trailers with no blanks between them."
+    )
+
+
+def check_blank_line_before_trailers(lines: list[str]) -> None:
+    """Require exactly one blank line between body and the trailer block.
```
aidanns: Claude Reviewer: Docstring/implementation drift. The header rule list at line 46 says "Exactly one blank line sits between the last body paragraph and the first trailer line", and `check_blank_line_before_trailers` at line 434 docstrings the same. But the implementation only checks that the line immediately preceding the trailer block is non-blank — two or more blank lines pass silently. I confirmed this locally: a body with two blank lines before `Closes #N` does not raise. Either tighten the implementation to also reject `>= 2` consecutive blanks above the trailer block (and update the error message), or relax the docstring/rule-comment to "at least one blank line". The latter matches `git interpret-trailers --parse` semantics (>=1 blank suffices), so the doc-only fix is the lower-risk one.

scripts/validate-commit-msg-block.py
```diff
@@ -308,6 +387,81 @@ def parse_trailer_block(lines: list[str]) -> list[tuple[int, str, str]]:
...
+    blank_offsets = [
+        first_trailer_idx + offset for offset, line in enumerate(region) if not line.strip()
+    ]
+    if not blank_offsets:
+        return
+    # 1-based line numbers for the error message, consistent with the
+    # rest of the validator's diagnostics.
+    blank_line_numbers = [idx + 1 for idx in blank_offsets]
+    raise ValidationError(
+        f"`{MARKER}` block has blank line(s) between trailers "
```
aidanns, line 472: Claude Reviewer: Error-message text is illustrated only with `Closes #N` / `Signed-off-by:`, but the contiguity rule fires equally on a `BREAKING CHANGE:` / `Signed-off-by:` split (verified locally — the trailer-shape predicate accepts `BREAKING CHANGE:` so the contiguity check rejects the blank between it and `Signed-off-by:`). The diagnostic is still actionable, but a contributor reading "a `Closes #N` separated from `Signed-off-by:`" who hit the failure on a `BREAKING CHANGE` footer may briefly think the message does not apply. Consider generalising the example, e.g. "two consecutive trailers (e.g. `Closes #N` and `Signed-off-by:`, or `BREAKING CHANGE:` and `Signed-off-by:`)".

scripts/validate-commit-msg-block.py
aidanns: Claude: Took option (b) — relaxed the docstring + header rule and the matching CONTRIBUTING.md prose to "at least one blank line", matching `git interpret-trailers --parse` semantics and the issue body wording. Implementation already accepts >= 1 blank, so no behaviour change. Pushed in 3545a23.

scripts/validate-commit-msg-block.py
aidanns: Claude: Generalised the diagnostic two ways: the per-line breakdown now names the actual neighbour trailers straddling each blank (e.g. `between `BREAKING CHANGE: …` and `Signed-off-by: …``), and the explanatory text now says "any two consecutive trailers (e.g. `Closes #N` and `Signed-off-by:`, or `BREAKING CHANGE:` and `Signed-off-by:`)" so the example reads as illustrative. Pushed in 3545a23.

PR #422, opened 2026-04-28
chore(ci): warn on verbose ==COMMIT_MSG== bodies

### LINKED ISSUE #395: commit-msg: tighten verbose-body guidance + add soft length lint
## Problem

Several recent commits have bodies that re-narrate the diff in prose rather than explaining *why* the change was made. The cbea.ms / Linux-kernel guidance (which `pr-lint.yml` already enforces structurally) says **why, not what** — but length is currently unconstrained, so verbose what-narrations slip through.

Example: [`7ab4c6a`](https://github.com/aidanns/agent-auth/commit/7ab4c6a90a1b283dbd12fce0a89af3642d1ca259) — \"chore: standardise YAML extension on .yml\". The body is ~360 words across six paragraphs, most of which restate the renamed paths, the regex tightenings, and the new gate. All of that is reachable via `git show`.

## What good looks like

The convention is **\"why, not what\"**, not \"be terse for terse's sake.\" Some commits in this project legitimately need long bodies — ADR-grade decisions, supply-chain shifts, wedge-bug fixes where the failure mode matters more than the diff. A blunt length cap would over-compress those.

## Proposed change

1. **Tighten the prose guidance** in `CONTRIBUTING.md` → \"Writing release-worthy commits.\" Add an explicit \"don't restate the diff\" rule with a counter-example pulled from `7ab4c6a`. Spell out when a long body *is* warranted (ADR-grade decision; non-obvious failure mode; supply-chain or release-pipeline change with audit-trail value).

2. **Add a soft length lint** to `scripts/validate-commit-msg-block.py` (or a sibling): warning (not error) when the COMMIT_MSG block exceeds N lines or M words, where the threshold is tuned to flag the worst offenders without false-positiving the legitimately-long bodies. Surface as a PR-comment annotation, not a CI failure.

3. **Reviewer playbook**: when the warning fires, the reviewer pushes back on the PR description rather than the contributor relitigating after merge.

## Acceptance criteria

- `CONTRIBUTING.md` documents the why-not-what rule with a concrete counter-example.
- The lint emits a warning (not a failure) on the worst offenders in current `git log`.
- A representative sample of legitimate long bodies (`b07fe58`, `c3c7136`, `166f55c`) does **not** trip the warning.

## Origin

Self-review of recent main-branch commit log. See PR review comment on [`7ab4c6a`](https://github.com/aidanns/agent-auth/commit/7ab4c6a90a1b283dbd12fce0a89af3642d1ca259) and discussion in this repo.

aidanns:
==COMMIT_MSG==
chore(ci): warn on verbose ==COMMIT_MSG== bodies

Bodies that re-narrate the diff in prose ("renamed these files,
tightened these regexes, added this gate") slip past the existing
==COMMIT_MSG== lint because length is unconstrained — the cbea.ms
"why, not what" rule is convention-only. Add a soft, warning-only
length lint so the worst offenders surface at PR-author time and
the reviewer can push back on the description before merge,
without false-positiving the bodies whose length is justified
(ADR-grade decisions, non-obvious failure modes, supply-chain or
release-pipeline changes with audit-trail value).

The thresholds (32 non-blank body lines, 250 words) were sized
against the historical commit log: the offender called out in the
issue (7ab4c6a) trips the line cap; the three legitimate long
bodies the issue named as must-not-trip (b07fe58, c3c7136,
166f55c) all stay under both. The line cap is the active
separator — 7ab4c6a sits beneath the legitimate c3c7136 on word
count, so a pure word threshold cannot tell them apart.

The warning prints to stderr and the validator still exits 0; the
pr-body-commit-msg job surfaces the line in its workflow log
without going red. CONTRIBUTING.md grows a counter-example that
shows the diff-restating shape against a one-paragraph rewrite,
plus the three categories where a long body is warranted, so the
reviewer playbook is documented next to the prose rules it
enforces.

Closes #395
==COMMIT_MSG==

==NO_CHANGELOG==

### Test plan

- [ ] `python3 scripts/validate-commit-msg-block.py --self-test` passes all 6 cases (4 verbose-body cases including the two boundary tests + 2 existing title-aware cases).
- [ ] Validator against every fixture under `.github/workflows/tests/pr-lint-fixtures/` returns the expected exit code (no regression in existing checks).
- [ ] Manual: feed the validator the body of `7ab4c6a` and confirm the warning fires; feed it `b07fe58`, `c3c7136`, and `166f55c` and confirm the warning is silent.
- [ ] CI: `pr-body-commit-msg` and `pr-body-title-aware-self-test` jobs are green; my own PR body does not trip the warning (210 words / 23 lines).

### DIFF
--- a/scripts/validate-commit-msg-block.py
+++ b/scripts/validate-commit-msg-block.py
@@ -960,6 +960,33 @@ _VERBOSE_BODY_SELF_TEST_CASES: tuple[tuple[str, bool, str], ...] = (
         True,
         "boundary-one-over-max-words",
     ),
+    (
+        # Boundary: non-blank body line count exactly at
+        # VERBOSE_BODY_MAX_LINES, word count well under the word cap.
+        # ``words_per_line=1`` makes line count == word count, so the
+        # body has 32 non-blank lines / 32 words — at the line
+        # threshold, well under the word threshold. The line check
+        # uses strict `>` so "exactly at the threshold" must NOT fire.
+        # Pairs with ``boundary-one-over-max-lines`` to lock the
+        # strict-greater-than contract on the line cap (the line cap
+        # is the active separator for the worst offender flagged in
+        # #395, so it deserves its own boundary coverage parallel to
+        # the word-cap pair above).
+        _build_verbose_body(VERBOSE_BODY_MAX_LINES, words_per_line=1),
+        False,
+        "boundary-exact-max-lines",
+    ),
+    (
+        # Boundary: one non-blank body line over VERBOSE_BODY_MAX_LINES,
+        # word count well under the word cap. 33 lines / 33 words
+        # isolates the line check from the word check, so a regression
+        # in the line check (off-by-one on the strict `>`, miscounting
+        # blank lines, slicing the body region wrong) would surface
+        # here even if the word check still works.
+        _build_verbose_body(VERBOSE_BODY_MAX_LINES + 1, words_per_line=1),
+        True,
+        "boundary-one-over-max-lines",
+    ),
 )
 
 
--- a/scripts/validate-commit-msg-block.py
+++ b/scripts/validate-commit-msg-block.py
@@ -49,6 +49,18 @@
 #      visual separation lives, not between trailers. This matches
 #      `git interpret-trailers --parse` semantics, which treats any
 #      run of one or more blanks as the body/trailer boundary.
+#  11. (Soft, warning-only — issue #395.) The body region (block
+#      content with the subject line and trailer block removed) is
+#      not unusually long. A body that exceeds either of the soft
+#      thresholds in `VERBOSE_BODY_MAX_LINES` /
+#      `VERBOSE_BODY_MAX_WORDS` prompts a stderr `Warning:` line
+#      pointing at CONTRIBUTING.md → "Writing release-worthy
+#      commits" → "Body" → "Lead with why, not what". This rule
+#      does NOT fail CI — bodies whose length is justified
+#      (ADR-grade decisions, non-obvious failure modes, supply-chain
+#      or release-pipeline changes with audit-trail value) are
+#      legitimately long and the project would rather surface the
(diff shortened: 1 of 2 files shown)


scripts/validate-commit-msg-block.py
```diff
@@ -776,6 +884,85 @@ def validate(body: str, title: str | None = None) -> None:
...
+    ),
+    (
+        # Boundary: one word over VERBOSE_BODY_MAX_WORDS — must fire.
+        # Pairs with the case above to lock the strict-greater-than
+        # contract: anything past the threshold trips the warning.
+        _build_verbose_body(VERBOSE_BODY_MAX_WORDS + 1),
+        True,
+        "boundary-one-over-max-words",
+    ),
+)
```
aidanns, line 990: Claude Reviewer: The four `_VERBOSE_BODY_SELF_TEST_CASES` cover the word-cap boundary (`boundary-exact-max-words` at 250 vs `boundary-one-over-max-words` at 251) but do not cover the **line-cap boundary**. The `verbose-body-trips` case uses `words_per_line=5` so its 300-word body also produces 60 lines, tripping both thresholds simultaneously — a regression that broke the line check (e.g. accidentally counting `lines[1:]` instead of non-blank lines, or off-by-one on the strict `>` comparison) while leaving the word check intact would not be caught by any case here. Verified empirically that the line check itself works (32 non-blank body lines passes, 33 fires the warning), but the line cap is exactly the threshold the issue accepts as the *active* separator for the worst offender (7ab4c6a — 216 words / 34 lines), so leaving it untested is the bigger of the two gaps. Suggest two more cases parallel to the word ones: `boundary-exact-max-lines` (32 non-blank lines, well under the word cap, expect no warning) and `boundary-one-over-max-lines` (33 lines, expect warning). `_build_verbose_body` already accepts `words_per_line` so the fixture wiring is already there.

aidanns: Claude: Done!

Issue #399, opened 2026-04-27
merge-bot: append (#N) to squash-merge commit titles


aidanns:
## Problem

Squash-merge commits made by \`merge-bot\` lose the \`(#N)\` PR-number suffix that GitHub's web UI \"Squash and merge\" button auto-appends. Without the suffix, GitHub does not render the commit subject as a clickable link to the originating PR in the commit log.

## Example

Compare:
- Pre-merge-bot (UI squash): [\`f72b2a2\`](https://github.com/aidanns/agent-auth/commit/f72b2a2) — \`feature(gpg-cli): implement refresh + reissue token workflow (#335)\` ← clickable.
- Post-merge-bot: [\`7ab4c6a\`](https://github.com/aidanns/agent-auth/commit/7ab4c6a90a1b283dbd12fce0a89af3642d1ca259) — \`chore: standardise YAML extension on .yml\` ← no link.

## Root cause

\`.github/workflows/merge-bot.yml:737\`:

\`\`\`bash
gh api --method PUT \\
  \"repos/\${GITHUB_REPOSITORY}/pulls/\${PR_NUMBER}/merge\" \\
  -f merge_method=squash \\
  -f commit_title=\"\${PR_TITLE}\" \\
  …
\`\`\`

The \`PUT /pulls/{n}/merge\` REST endpoint does **not** auto-append \`(#N)\` the way the UI does. The bot passes the title verbatim.

## Proposed fix

Append the PR number explicitly:

\`\`\`bash
-f commit_title=\"\${PR_TITLE} (#\${PR_NUMBER})\"
\`\`\`

## Pre-flight: title-length cap

\`scripts/validate-pr-title.py\` enforces a 72-char hard cap on the un-suffixed title. Adding \`(#NNNN)\` is up to ~7 extra chars. Two options:

1. **Lint un-suffixed, append in merge-bot.** Keep the validator as-is; accept that the final commit subject can exceed 72 chars by ~7. Pro: contributors aren't penalised for high PR numbers. Con: the rendered subject violates the documented cap.
2. **Lint with projected suffix.** Update the validator to compute \`len(title) + len(f\" (#{pr_number})\")\` against the cap. Pro: the documented limit holds. Con: requires the PR number at lint time (available via \`github.event.pull_request.number\`).

Prefer option (2) — keeps the cap honest.

## Acceptance criteria

- New merge-bot squash commits show \`(#N)\` in the subject.
[...]

PR #424, opened 2026-04-28
fix(release): derive wheel versions from workspace git root

### LINKED ISSUE #408: fix(release): v0.16.0 wheels versioned 0.0.0+unknown instead of 0.16.0

aidanns:
==COMMIT_MSG==
v0.16.0 shipped 70 supply-chain assets versioned ``0.0.0+unknown``
because each ``packages/<svc>/pyproject.toml`` declared
``[tool.setuptools_scm]`` without setting ``root``. With ``uv build
--all-packages``, the build runs from the per-package directory
which is not a git root, so setuptools-scm fell back to
``fallback_version = "0.0.0+unknown"`` instead of reading the
workspace's git tags. The ``cosign verify`` recipe in SECURITY.md
hardcodes ``${PKG}-${VERSION}-py3-none-any.whl``, so consumers
running the documented verification path could not match a single
asset on the v0.16.0 release page.

Pin ``[tool.setuptools_scm].root = "../.."`` in every per-package
pyproject.toml so setuptools-scm walks two levels up to the
workspace's ``.git``. Reproduced locally that wheels are now
versioned ``0.16.2.devN+g<sha>`` on a non-tag commit and would be
``0.16.1`` on the tag commit.

Add two regression gates:

- ``tests/test_release_wheel_version.py`` runs in ``task test``
  (which the ``unit`` CI job invokes; ``release-dryrun.yml``
  itself only invokes the build script and a count gate, so it
  does NOT run pytest). One case statically asserts every package
  pins ``root = "../.."``; the other invokes the build script and
  asserts every produced wheel + sdist shares one git-derived
  version segment matching a strict PEP 440 regex (rejecting any
  ``unknown`` substring so shallow-checkout fallback strings like
  ``0.0.1.dev1+unknown.gsha`` cannot pass vacuously).
- ``release-publish.yml`` gains a fail-fast step between ``uv
  build`` and ``cosign sign`` that asserts every artefact filename
  contains the release-tag's version. Belt-and-braces for the
  publish path itself, so any future SCM-checkout regression that
  slips past the PR-time dry-run cannot still produce a signed
  release with the wrong filenames.

Closes #408
==COMMIT_MSG==

### Internal decisions

- **Picked Option A from the issue's hypothesis (configure
  setuptools-scm) rather than B (thread version via env var) or C
  (workflow-level checkout fix).** The workflow already does
  ``fetch-depth: 0`` and the tag IS visible in the checkout — I
  reproduced the bug locally with full git history in place. The
  problem is that setuptools-scm reads ``pyproject.toml``'s parent
  as its SCM root by default; for a uv workspace that's the
  per-package dir which has no ``.git``. ``root = "../.."`` fixes
  this without touching workflow YAML or build scripts.
- **Kept ``fallback_version = "0.0.0+unknown"``.** Removing it
  would explode loudly on any non-git build context (sdist
  tarball, vendored extract). The new tests + the
  publish-workflow guard catch the only failure mode that matters
  (release builds shipping the fallback).
- **No changes to ``SECURITY.md``.** The fix restores the
  ``${PKG}-${VERSION}-py3-none-any.whl`` shape the recipe already
  documents; the v0.16.0 advisory note added in #410 stays
  accurate.

### Test plan

- [ ] ``task test`` passes locally — includes the new
  ``tests/test_release_wheel_version.py``.
- [ ] ``uv build --all-packages --out-dir /tmp/dist-test`` produces
  filenames containing a real PEP-440 version, not
  ``0.0.0+unknown``.
- [ ] CI ``release-dryrun.yml`` build succeeds and produces correct
  filenames (visible in the job log).
- [ ] CI ``Lint / pr-title`` accepts ``fix(release): …``.
- [ ] CI ``Lint / commit-msg-block`` accepts the ==COMMIT_MSG==
  block.
- [ ] CI ``DCO sign-off check`` passes.
- [ ] CI ``Changelog Lint`` passes once the
  ``changelog/@unreleased/pr-<N>-*.yml`` entry is added.

### Self-review only

The Agent / general-purpose subagent tool was unavailable in the
isolated worktree environment used for this implementation, so the
PR-author-side review was performed inline against `CLAUDE.md`,
`.claude/instructions/*.md`, and `CONTRIBUTING.md` rather than by a
spawned review subagent. The orchestrator's Phase 5 monitor should
dispatch a review subagent at its level to fill this gap.

Inline self-review findings (all addressed):

- PR title `fix(release):` matches the project allowlist; `release`
  is the correct two-tier scope for a multi-package change.
- ==COMMIT_MSG== block formatted per `pr-lint.yml` rules: no
  headings, lines wrap <=72 chars, `Closes #408` + `Signed-off-by:`
  contiguous trailers with one blank line above.
- Commit signed off; subject carries no `(#408)` suffix (the merge
  bot appends `(#<pr#>)` itself per #399).
- Regression test built into both `task test` (PR-time) and the
  publish workflow (release-time); confirmed the static-config
  assertion would have failed pre-fix.
- No security-sensitive surfaces touched; `setuptools-scm` is
  already in every package's `[build-system].requires`.
- Restores the "Version string from VCS tags" standard documented
  in `.claude/instructions/release-and-hygiene.md`.


tests/test_release_wheel_version.py
```diff
@@ -0,0 +1,145 @@
...
+  workspace member. A new package added without the right ``root``
+  fails here without needing to run ``uv build``.
+- ``test_uv_build_derives_version_from_git`` actually shells out to
+  ``scripts/build-release-artifacts.sh`` and asserts every produced
+  wheel + sdist filename shares one non-fallback version segment. The
+  ``release-publish.yml`` workflow gate cross-checks the version
+  against the release tag at publish time; this PR-time smoke test
+  only needs to assert the build is reading git, not falling back to
+  ``0.0.0+unknown`` — which is what would have caught the v0.16.0
+  regression at PR time.
```
aidanns: Claude Reviewer: The PR body and the commit message both claim this test runs "via `release-dryrun.yml` on every PR", but `release-dryrun.yml` only invokes `scripts/build-release-artifacts.sh` and asserts the artefact-count gate — there is no pytest step in that workflow. The new test runs only via `task test` (the `unit` job in `test.yml`). Consider either updating the description to remove the inaccurate claim, or adding a step to `release-dryrun.yml` that re-runs the basename version check against the temp `dist/`.

tests/test_release_wheel_version.py
```diff
@@ -0,0 +1,145 @@
...
+    (version,) = versions
+    # Reject the setuptools-scm fallback verbatim — that string is the
+    # signature of the regression this test exists to catch. The
+    # ``release-publish.yml`` guard cross-checks the version against
+    # the release tag at publish time; this PR-time test only needs
+    # to assert the build is reading git, not falling back.
+    assert "0.0.0+unknown" not in version, (
+        f"artefacts versioned as setuptools-scm fallback ({version!r}); "
+        "[tool.setuptools_scm].root is not pointing at the workspace git root"
+    )
```
aidanns, line 144: Claude Reviewer: This assertion only catches the literal `0.0.0+unknown` string. The unit-test job in `test.yml` checks out with default `fetch-depth: 1` (no tags), and on a tag-less shallow clone setuptools-scm produces a version like `0.0.1.dev1+unknown.gfdefe9c26` — which contains `unknown` but not the literal `0.0.0+unknown`, so this assertion passes vacuously. That means this PR-time test exercises the `subprocess.run` plumbing but does not actually verify "tags are visible to the build" in CI. Two ways to tighten it: (1) give the test a stricter assertion such as "version string starts with a digit and matches `\d+\.\d+\.\d+`" so it rejects any `0.0.x.devN+unknown.*` variant, or (2) have the test xfail / skip when no git tags are reachable from HEAD so it is honestly informational rather than silently green.

.github/workflows/release-publish.yml
```diff
@@ -74,6 +74,39 @@ jobs:
...
+          # so the comparison matches setuptools-scm's PEP-440 output.
+          expected_version="${TAG#v}"
+          fail=0
+          for f in dist/*.whl dist/*.tar.gz; do
+            base="$(basename "${f}")"
+            if [[ "${base}" != *"-${expected_version}-"* && "${base}" != *"-${expected_version}.tar.gz" ]]; then
+              echo "::error::artefact ${base} does not carry expected version ${expected_version} (regression of #408)"
+              fail=1
+            fi
+          done
```
aidanns, line 119: Claude Reviewer: Minor: this fail-fast loop runs after `set -euo pipefail`, but if `dist/` is empty (e.g. a future regression where the build silently produces zero artefacts) the globs `dist/*.whl dist/*.tar.gz` expand to the literal patterns and the inner test fails the basename match → exits 1. That happens to be the desired outcome, but it is incidental rather than intended. Consider an explicit early sanity check (`shopt -s nullglob; files=(dist/*.whl dist/*.tar.gz); [[ ${#files[@]} -gt 0 ]] || { echo "::error::no artefacts found in dist/"; exit 1; }`) so the failure mode and message are explicit.

tests/test_release_wheel_version.py
aidanns: Claude: Done! Updated docstring + PR body commit-message block to say the test runs in `task test` (the unit CI job).

tests/test_release_wheel_version.py
aidanns: Claude: Done! Tightened to a regex (`^\d+\.\d+\.\d+(\.dev\d+)?(\+g[0-9a-f]+(\.d\d+)?)?$`) that rejects any `unknown` substring and requires a real PEP 440 numeric core. Accepts the legitimate `X.Y.Z[.devN][+gSHA[.dDATE]]` shape the fix produces post-tag, so shallow checkouts where the tag IS reachable still pass.

.github/workflows/release-publish.yml
aidanns: Claude: Done! Added `shopt -s nullglob` plus an explicit empty-array guard before the loop, so an empty `dist/` produces a loud workflow failure with a clear `::error::` message instead of either silently passing or running with bogus literal filenames.

Issue #294, opened 2026-04-25
test(perf): compose_stop sits at the 5s grace ceiling — SIGTERM handlers don't short-circuit teardown


aidanns:
## Symptom

After #288 (harness drops `-t 30` so compose-file `stop_grace_period: 5s` applies), every per-test compose teardown clusters tightly at ~5.7–6.7 s:

```
phase=compose_stop elapsed_seconds=5.758  service=agent-auth
phase=compose_stop elapsed_seconds=5.779  service=agent-auth
phase=compose_stop elapsed_seconds=5.781  service=agent-auth
phase=compose_stop elapsed_seconds=5.786  service=agent-auth
phase=compose_stop elapsed_seconds=6.019  service=things-bridge
phase=compose_stop elapsed_seconds=6.023  service=things-bridge
phase=compose_stop elapsed_seconds=6.026  service=things-bridge
phase=compose_stop elapsed_seconds=6.456  service=things-bridge   # things-cli stack
phase=compose_stop elapsed_seconds=6.527  service=things-bridge   # things-cli stack
phase=compose_stop elapsed_seconds=6.661  service=things-bridge   # things-cli stack
```

The clustering is way too tight to be anything other than \"docker hits the 5 s `stop_grace_period` ceiling and SIGKILLs / the in-process 5 s watchdog calls `os._exit(1)`.\" #154's acceptance criterion was \"<2 s per test.\"

## Why this matters

- We're shipping dirty shutdowns in tests today: agent-auth's SQLite WAL doesn't checkpoint cleanly, the audit log may lose buffered writes, and the things-bridge in-flight subprocess (the fake Things CLI) is killed mid-stream rather than draining. None of it surfaces as a test failure but the production path goes through the same code, so the integration suite is effectively *not* exercising graceful shutdown end-to-end.
- The unit-level subprocess shutdown tests added in #163 (commit `e496217`) pass — so the handlers are wired, but something about the in-container, in-compose context is keeping the process alive past the watchdog deadline.
- The harness budget gate from #288 is set at 10 s; once this is fixed we can ratchet it down to ~3 s and lock in the win.

## Likely root cause

[...]

PR #426, opened 2026-04-28
improvement(release): append (#N) PR-link to release-note entries

### LINKED ISSUE #411: release: append (#N) PR-link to each release-note entry
## Problem

Release-note entries currently carry no link back to the originating PR. With #407 (terse \`description:\`) landing, the entry becomes information-lossy: a reader who wants the full \"why\" or implementation detail has no path from CHANGELOG.md / the release page to the originating PR's \`==COMMIT_MSG==\` block (where the verbose prose lives per the audience separation).

#399 fixes the same problem for non-release commit *titles* on \`main\` (squash-merge subjects get \`(#N)\`). This issue applies the same idea to release-note *entries*.

## Proposed fix

Append \`(#N)\` to the end of each rendered release-note entry, in all three audiences:

1. The release-PR squash-merge commit body on \`main\` (\`==COMMIT_MSG==\` block, via \`render_commit_msg_block\`).
2. \`CHANGELOG.md\` (via \`render_release_notes\`).
3. The GitHub release page.

Example after this lands (assuming #397 and #407):

\`\`\`
Improvements:
- \`merge-bot\` now auto-updates a PR whose head sits behind \`main\` instead of treating the merge API's 405 as a hard failure. (#383)
- \`pr-lint\` allows plain bullet lists in COMMIT_MSG block. (#384)
\`\`\`

GitHub auto-renders \`#NNN\` to a clickable PR link in commit bodies, CHANGELOG.md (when viewed on github.com), and the release page — no markdown link syntax needed.

## Implementation notes

**Source the PR number from the filename, not \`links:\`.** The convention \`pr-<N>-<slug>.yml\` is enforced by \`scripts/changelog/add.py\` and ADR 0037. The filename is single-PR, deterministic, and machine-readable. The \`links:\` array is human-authored and may include multiple PRs, issues, or be missing entirely. Parse with a regex like \`^pr-(\\d+)-.+\\.yml$\`.

**Single-point append.** Add the suffix in \`_flatten_description\` (or its caller in \`build_release.py\`) so all three audiences inherit it from one place. Re-runs \`render_release_notes\` and \`render_commit_msg_block\` simultaneously.

**Format.** \`(#NNN)\` at end of the entry. Bare; let GitHub auto-link it.

**Validator interaction (\`render_commit_msg_block\` only).** \`_wrap_paragraph\` has a no-numbered-list-opener rule that rejects \`<digits>.\` or \`<digits>)\` at line start. \`(#383)\` doesn't trigger that rule because it starts with \`(\`. But if the wrap algorithm splits the line such that \`(#383)\` lands alone on a wrapped line, it should not wrap. Treat the PR-suffix as bound to the preceding whitespace-separated token (i.e. never insert a line break immediately before \`(#NNN)\`). One-line change to \`_wrap_paragraph\` or its tokeniser.

**Filename without a PR number.** Some legacy entries may not match the \`pr-<N>-\` pattern (e.g. backfilled entries or entries authored by tooling). Two reasonable handlings:
- Skip the suffix for entries whose filename doesn't match the pattern (fail-soft).
- Lint that every \`changelog/@unreleased/*.yml\` matches \`pr-<N>-\` at PR-time, with a clear error pointing at the offender.

Prefer the second — \`scripts/changelog/add.py\` already produces conformant filenames; an entry that doesn't match likely indicates a bypass that shouldn't have happened.

## Acceptance criteria

- New release commits, CHANGELOG.md updates, and GitHub release notes show \`(#N)\` at the end of each entry.
- GitHub renders the suffix as a clickable link in all three audiences.
- The \`==COMMIT_MSG==\` validator continues to pass on the new output.
- The wrap algorithm never leaves \`(#N)\` alone at the start of a wrapped line.
- A test fixture covers a long single-bullet description that wraps; assert the suffix stays bound to the preceding token.
- Filename-pattern lint (or fail-soft skip) decided and documented.

## Out of scope

- Backfilling \`(#N)\` into already-published \`CHANGELOG.md\` entries on \`main\`. Forward-only.
- Changing the YAML schema. The PR number is derived from the filename, not added as a new field.

## Related

- #397 (always-bullets) — bullets are the natural per-entry shape that the suffix attaches to.
- #399 (merge-bot \`(#N)\` on commit titles) — same convention applied to PR titles.
- #407 (terse \`description:\`) — closes the information-loss gap that this issue compensates for.

## Origin

Self-review of recent main-branch CHANGELOG / release notes. Surfaces from the same audience-separation thinking as #407 — terse description + PR-link suffix together restore information access without bloating the release notes.

aidanns:
==COMMIT_MSG==
Each rendered release-note entry now ends with `(#N)` so a reader of
CHANGELOG.md, the GitHub release body, or the release-PR
==COMMIT_MSG== block can click straight through to the originating
PR for the verbose context that the terse `description:` field
deliberately omits. Single-point append in `_pr_link_suffix` —
sourcing the PR number from the YAML filename via
`ENTRY_FILENAME_PATTERN` — so all three audiences inherit the
suffix from one helper. The wrap algorithm gains a sibling rule to
the existing numbered-reference one: `(#N)` tokens are bound to the
preceding word so a soft-wrap never visually divorces the link from
the entry it points at. A new PR-time lint check
(`check_present_file_naming`) fails any
`changelog/@unreleased/*.yml` whose filename doesn't match
`pr-<N>-<slug>.yml`, so the renderer can always derive the PR
number — fail-soft skipping the suffix at release time would
silently drop the link from published entries.

Closes #411
==COMMIT_MSG==

## Review notes

Implements #411. Three rendered audiences now carry the per-entry
PR-link suffix: `CHANGELOG.md` bullets, the GitHub-release body
shared with the release PR's preview, and the `==COMMIT_MSG==`
block pasted as the squash-merge commit body. The PR number is
derived from the YAML filename pattern that
`scripts/changelog/add.py` already enforces, not the
human-authored `links:` array (which may carry multiple URLs or
none).

A new PR-time lint check (`check_present_file_naming`) fails any
non-conforming `changelog/@unreleased/*.yml`, so renderer
fail-soft is reachable only by legacy entries that pre-date this
PR — the existing `pr-99999-release-pr-link-suffix.yml` placeholder
on this branch will be renamed to the actual PR number after
opening (so this PR's own entry conforms in turn).

### Test plan

- [ ] `task changelog:test` (227 baseline + 6 new tests pass)
- [ ] `uv run --frozen ruff check scripts/changelog/` (clean)
- [ ] `uv run --frozen ruff format --check scripts/changelog/` (clean)
- [ ] Smoke-test the rendered output by building a `compute_release` plan against
      a fixture matching the issue's example (two improvement YAMLs) and
      asserting the bullets end with `(#383)` / `(#384)`.


scripts/changelog/tests/test_build_release.py
```diff
@@ -410,6 +410,137 @@ def test_render_commit_msg_block_wraps_normally_around_hyphen_joined_id(
...
+    visually divorced from the entry it links to defeats the
+    audience-link-back motive (#411). The fixture's prose is sized so
+    the buggy wrapper would land the suffix at the start of the
+    second wrapped line of the Improvements paragraph; the fix moves
+    the preceding token down with it.
+    """
+    # Crafted so the Improvements paragraph header + this prose pushes
+    # `(#411)` past the 72-char limit: forcing the wrap algorithm to
+    # decide where the suffix belongs.
+    description = (
```
aidanns, line 563: Claude Reviewer: This regression-test fixture does not actually exercise the bug it claims to test. With the assembled prose `Improvements: Auto-update a PR whose head sits behind main instead of treating the merge API's 405 as a hard failure (#411).` (124 chars), a naive wrapper without the `PR_SUFFIX_RE` rule produces the same two lines as the fixed wrapper — line 1 fills to 69 chars (`Improvements: ... instead of`), line 2 is `treating the merge API's 405 as a hard failure (#411).` (54 chars). The `(#411).` suffix fits naturally on line 2 because there is no token-binding decision to make. I confirmed this by monkey-patching `PR_SUFFIX_RE` to a never-match regex — `render_commit_msg_block` produces byte-identical output. Removing the suffix-binding branch from `_wrap_paragraph` would not fail this test. To turn this into a real regression test, lengthen line 2's prose so that `(#411)` would land alone at the start of a wrapped third line under the buggy wrapper (e.g. add ~12 more chars of prose between "failure" and the natural wrap point). The acceptance-criteria intent in #411 is that this be a true regression guard, not just a smoke test.

scripts/changelog/lint.py
```diff
@@ -289,19 +315,26 @@ def run_lint(
...
     check_file_presence(pr_number, added_files, labels, report)
 
     # Check 2: file naming on every added file.
     check_file_naming(pr_number, added_files, report)
 
+    # Check 2b: filename pattern on every file already present in
+    # `@unreleased/`. Guards the release renderer's `(#N)` PR-link
+    # suffix derivation (#411) against legacy / hand-edited entries
+    # that bypass `task changelog:add`.
+    check_present_file_naming(present_files, report)
```
aidanns: Claude Reviewer: When a PR adds a non-conforming filename, both `check_file_naming` and `check_present_file_naming` will flag the same file (the just-added file is in both `added_files` and `present_files`), producing two near-duplicate "filename must match" lines in the rendered report. Verified by running `run_lint` against a `wrong-name.yml` added in a single PR diff:
```
changelog/@unreleased/wrong-name.yml: filename must match `pr-<N>-<slug>.yml` (letters, digits, dashes, underscores in the slug)
changelog/@unreleased/wrong-name.yml: filename must match `pr-<N>-<slug>.yml` so the release renderer can append the `(#N)` PR-link suffix. Rename the file or recreate it via `task changelog:add`.
```
The existing `test_run_lint_fails_when_filename_does_not_match_pattern` (line 247 in test_lint.py) only asserts `"filename must match" in report.render()` so it still passes, but the contributor experience is worse. Two options:
1. Skip files already covered by `check_file_naming`: pass `added_files` to `check_present_file_naming` and exclude them, or short-circuit when the path is already in `added_files`.
2. Fold the two checks into one and keep separate messages only when the offender is on `main` (i.e. not added in this PR).

Either is fine; option 1 is the smaller patch.

scripts/changelog/tests/test_build_release.py
aidanns: Claude: Done! Lengthened the prose so a naive (unbound) wrapper would orphan `(#411).` on a third wrapped line — verified by monkey-patching `PR_SUFFIX_RE` to never match (now produces `previously` on line 2 and `(#411).` alone on line 3, exactly the failure mode the assertions forbid). Also added a meta-test `test_pr_link_suffix_binding_fixture_actually_exercises_binding_rule` that pins this property so future fixture edits that shorten the prose below the threshold fail loudly rather than silently degrading into a smoke test.

scripts/changelog/lint.py
aidanns: Claude: Done! Took option 1 — `check_present_file_naming` now accepts `added_files` and skips paths already covered by `check_file_naming`. A non-conforming filename added in this PR now produces exactly one "filename must match" line. Added regression test `test_run_lint_emits_single_filename_error_when_added_file_is_non_conforming` that pins the dedup, and the existing legacy-on-`main` test (`test_run_lint_fails_when_unreleased_file_lacks_pr_prefix`) still passes — the present-files-only case is unchanged.

aidanns:
Claude: integration-agent-auth failed on `compose_stop` phase budget (7.2s vs 3.0s budget) — all 30 integration tests passed. This is a known timing flake in the test harness (see #292, #294); the diff is changelog-tooling only and does not touch Docker/compose. Re-running the failed job.

Issue #100, opened 2026-04-19
Consolidate structured audit log format across all services


aidanns:
All services (agent-auth, things-bridge) should emit audit events in a single, consistent structured JSON schema. Currently each service logs independently with no shared envelope format, which complicates SIEM integration and forensic analysis.

**Acceptance:**
- Define a shared audit-event JSON schema (service name, event type, timestamp, payload fields).
- Apply it to agent-auth's audit log and things-bridge's request log.
- Pin the schema with a test so field renames are caught as breaking changes.
- Document the schema in `design/DESIGN.md`.

Tracked from the audit surface section of `SECURITY.md`.

PR #1, opened 2026-04-12
Define system design, functional decomposition, and product breakdown

aidanns:
## Summary

- Add system design document covering token model (access/refresh pair with HMAC signing), three-tier authorization (allowed/prompt/denied), JIT approval flow, HTTP API, and network configuration for host and devcontainer usage.
- Add functional decomposition (37 leaf functions across 9 top-level functions) and product breakdown (11 configuration items across 3 components: agent-auth, things-bridge, things-cli) in systems-engineering CLI format.

## Test plan

- [x] `systems-engineering product verify -p product_breakdown.yaml -f functional_decomposition.yaml` passes
- [x] `systems-engineering function diagram functional_decomposition.yaml -o diagrams` generates diagrams
- [x] `systems-engineering product diagram product_breakdown.yaml -o diagrams` generates diagrams

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #423, opened 2026-04-28
chore(ci): submit dependency snapshot on pull_request

aidanns:
==COMMIT_MSG==
chore(ci): submit dependency snapshot on pull_request

Without a head-SHA snapshot, actions/dependency-review-action prints
"No snapshots were found for the head SHA <sha>" on every PR and
silently downgrades to native pyproject.toml parsing. The CI tools
declared in .github/tool-versions.yml (shellcheck, ruff, uv, ...) are
not exercised by the PR-time CVE gate, so a PR bumping one of those
to a CVE-bearing version slips past dep-review and only surfaces in
the next Monday cron-driven main-branch snapshot.

Add `pull_request` to dependency-submission.yml's trigger list so
the snapshot is also submitted for the PR head SHA. Snapshots are
keyed by SHA, so a PR-triggered submission cannot rewrite main's
snapshot — the original "no pull_request trigger" rationale was
outdated. Fork PRs receive a read-only GITHUB_TOKEN, so the POST
would 403; a job-level `if:` guard skips submission cleanly on fork
PRs while leaving push / cron / workflow_dispatch unaffected. The
condition is written in the skip-only-on-fork-PR direction
(`event_name != 'pull_request' || head.repo == repo`) so adding new
non-pull_request triggers later doesn't accidentally gate them on
the fork check.

Update ADR 0031 to reflect the new trigger and replace the
"no pull-request path — malicious PR cannot poison the snapshot"
rationale with the head-SHA-keying argument plus the fork-skip
mitigation. Update merge-bot.yml's exclusion comment so the
dependency-submission line is no longer claimed to "not run on
PRs"; it does run on PRs but is intentionally not a required gate
because making it required would block fork PRs from ever merging.

Closes #412
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

Internal implementation choices, resolved per the dispatch instructions:

- **Guard direction:** chose `if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository` (skip-only-on-fork-PR) over the equivalent `event_name == 'pull_request' && head.repo == repo` form. The chosen form means non-`pull_request` triggers (push / cron / workflow_dispatch) keep running unchanged regardless of the fork-context fields, which is the safer default for any future trigger added to the `on:` list.
[...]

PR #434, opened 2026-04-28
improvement: close linked issues after squash-merge

### LINKED ISSUE #429: merge-bot: close linked issues after squash-merge (App-token auto-close doesn't fire)
## Problem

The merge-bot squash-merges PRs via `PUT /pulls/{n}/merge` using the bypass-actor App token (`MERGE_BOT_APP_ID` / `MERGE_BOT_PRIVATE_KEY`). GitHub's auto-close-on-`Closes #N` **does not fire** for App-token-mediated API merges, even when the squash commit body contains the trailer correctly. This is deterministic, not intermittent — confirmed across multiple PRs:

- #350 (closing #347) — 2026-04-26
- #354 (closing #348) — 2026-04-26
- #423 (closing #412) — 2026-04-28

(see `feedback_merge_bot_quirks.md` in the auto-memory store for the original observation).

The current workaround is a manual `gh issue close <N>` step in every PR-shepherding workflow (Phase 6 step 2 of `/workflow:implement`, and the equivalent in `/work-issues`). If the orchestrator skips that step — or the user merges via the bot manually and forgets — the linked issue stays OPEN indefinitely. There's no other safety net: the issue tracker silently drifts out of sync with the actual state of the work.

## Proposed change

Have the merge-bot itself close linked issues after a successful squash-merge. After `PUT /pulls/{n}/merge` returns 200 (or equivalently after `merged_at` is set on the merge response), the bot:

1. Reads the squash commit body from the merge response.
2. Parses linked-issue references using the standard GitHub auto-close keywords — `closes`, `fixes`, `resolves` (case-insensitive) — followed by `#N` (same-repo) or `org/repo#N` (cross-repo). The set of recognised keywords matches what GitHub's UI auto-closer uses.
3. For each same-repo issue reference, calls `PATCH /repos/{owner}/{repo}/issues/{N}` with `state: closed` and posts a closing comment of the form: `Closed by merge of PR #<P> (squash commit <sha>).`
4. Skips cross-repo references (`org/repo#N`) — the App's installation token isn't scoped to other repos. Log a notice in the workflow output so it's discoverable.

The merge-bot's job changes from "merge PR" to "merge PR + close its linked issues", which matches what users already expect from `Closes #N`.

## Permissions impact

The merge-bot App's installation permissions need `issues: write` if not already granted. Currently the bot likely has `pull_requests: write` (for the merge call) and `contents: write` (for the squash commit). Adding `issues: write` is a small, deliberate scope widening — it lets the bot close issues but not delete them or modify other issue fields beyond `state` (the API endpoint we're calling is scoped). Worth calling out in the App's installation prompt for transparency.

## Alternatives considered

- **Switch the merge-bot to a personal token.** Auto-close would work, but defeats the bypass-actor pattern that lets the bot merge despite branch-protection rulesets. Rejected — the bypass-actor pattern is the entire reason the bot exists in its current form.
- **Add a separate workflow on `pull_request: closed` (merged: true) that closes linked issues.** Splits responsibility for no benefit — the bot already has the merge context, including the resolved squash commit body. Rejected as gratuitous workflow proliferation.
- **Wait for GitHub to fix the auto-closer for App tokens.** Confirmed deterministic as of 2026-04-28 across three sampled PRs; not a transient bug. No public GitHub issue tracking a fix is known to me. Rejected as indefinite.
- **Keep the manual workaround and rely on Phase 6 step 2 / `/work-issues` discipline.** This is the status quo. Rejected: a workflow-level invariant that depends on every orchestrator never skipping a step is fragile, and there's no observable signal when it does drift (an issue staying open looks the same as an in-progress issue).

## Acceptance

- A merge-bot squash-merge of a PR whose squash commit body contains `Closes #<N>` results in issue `#<N>` transitioning to `CLOSED` with a closing comment, without any orchestrator-side `gh issue close` step.
- Removing the manual `gh issue close` from Phase 6 step 2 of `/workflow:implement` no longer causes issues to drift OPEN. (Verified by intentionally skipping the manual close on a test PR and confirming the bot closed the issue.)
- Multiple linked issues in one PR (`Closes #N\nCloses #M`) are all closed.
- A cross-repo reference (`Closes other-org/other-repo#N`) is skipped without erroring the merge step, and the skip is logged.
- An already-closed issue (`Closes #N` where N is already CLOSED) is a no-op — the bot does not re-comment or error.

## Additional context

- Originating context: this orchestrator-side workaround was applied during the merge of PR #423 (issue #412 — `chore(ci): submit dependency snapshot on pull_request`), which is the most recent confirmation of the auto-close failure.
- Memory note: `feedback_merge_bot_quirks.md` (auto-memory store) documents this as quirk #1; quirk #2 (changelog-bot label-toggle race) is a separate concern.
- Once this issue ships, the manual `gh issue close` calls in `/workflow:implement` Phase 6 step 2 and the equivalent step in `/work-issues` should be removed (or downgraded to a defensive sanity-check that's expected to be a no-op). Either option is fine — slight preference for removal so the bot is the single source of truth for issue closure.

## Related work

- ADR 0038 (merge-bot design) — should be updated to reflect the new responsibility.
- `.github/workflows/merge-bot.yml` is the workflow this PR will modify.
- Issue #348 (`fix(merge-bot): trigger on workflow_run`) is unrelated but is a recent reference for the merge-bot's evolution.

aidanns:
==COMMIT_MSG==
improvement: close linked issues after squash-merge

GitHub's UI auto-close-on-`Closes #N` does not fire for App-token-
mediated `PUT /pulls/{n}/merge` calls. Confirmed deterministic
across PRs #350 / #354 / #423: the squash commit body carries the
trailer, the merge succeeds, but the linked issue stays OPEN. The
current workaround is a manual `gh issue close` step in every PR-
shepherding workflow; when an orchestrator skips that step the
tracker silently drifts out of sync with the actual state of the
work, with no observable signal.

Move closure into the merge bot: after the merge call returns
200, parse the squash commit body for GitHub's auto-close keyword
set (closes / closed / closing / fix / fixed / fixes / fixing /
resolve / resolved / resolves / resolving — case-insensitive,
matching what GitHub's UI accepts), then call
`PATCH /repos/.../issues/{N}` with `state: closed` and post a
`Closed by merge of PR #<P> (squash commit <sha>).` comment on
each same-repo issue. Cross-repo references are logged via
`::notice::` and skipped (the App's installation token is scoped
to this repo). Already-closed issues are skipped to avoid
duplicate comments. Per-issue close failures (404, 5xx) are
warning-only — the merge has already landed by this point;
failing the workflow would generate a red merge-bot run on a PR
that merged successfully without rolling back the merge, a
worse signal than a warning-only log line.

Implementation: a new `scripts/parse-close-keywords.py` helper
parses the body via `Path.read_text` (no shell pipeline so
backticks / `$(...)` in the body cannot be evaluated) and writes
same-repo issue numbers to stdout one per line; the workflow's
new `Close linked issues` step loops over them via `mapfile`.
HTML comments are stripped before scanning so template
scaffolding examples like `<!-- "Closes #N" is also accepted -->`
do not produce false-positive matches. The parser pins the full
keyword set in `AUTO_CLOSE_KEYWORDS`; tests cover every keyword
inflection, case sensitivity, trailing punctuation,
deduplication, and cross-repo separation.

This expands the merge-bot App's installation permissions to
include `issues: write`. The actual permissions update is an
operational step the maintainer performs; the setup doc lists
the new requirement in Step 1.

Closes #429
==COMMIT_MSG==

==CHANGELOG_MSG==
Merge-bot now closes issues referenced via auto-close keywords
(`Closes #N`, `Fixes #N`, `Resolves #N` and the inflections
GitHub's UI accepts) in the squash commit body, with a closing
comment of the form `Closed by merge of PR #<P> (squash commit
<sha>).`. Compensates for GitHub's auto-close-on-`Closes #N` not
firing for App-token-mediated merges. Requires `issues: write`
on the merge-bot App's installation; see
`docs/release/merge-bot-setup.md` Step 1.
==CHANGELOG_MSG==

### Test plan

- [ ] `uv run pytest tests/test_parse_close_keywords.py` — 39 unit
  tests covering every keyword inflection, case-insensitivity,
  trailing-punctuation tolerance, deduplication, cross-repo
  splitting, HTML-comment stripping, and the CLI's
  stdout/stderr contract.
- [ ] `uv run ruff check scripts/parse-close-keywords.py
  tests/test_parse_close_keywords.py` — clean.
- [ ] `uv run ruff format --check
  scripts/parse-close-keywords.py
  tests/test_parse_close_keywords.py` — clean.
- [ ] `uv run mypy scripts/parse-close-keywords.py` — clean.
- [ ] `uv run pyright scripts/parse-close-keywords.py` — clean.
- [ ] `shellcheck` on the new `Close linked issues` step body —
  clean (shellcheck against the workflow YAML's run block, with
  the existing `SC2154` exclusion the rest of the file uses).
- [ ] `uv run python3 -c "import yaml;
  yaml.safe_load(open('.github/workflows/merge-bot.yml'))"` —
  workflow YAML parses cleanly.
- [ ] Live verification (post-merge): once the merge-bot App
  installation has the new `issues: write` permission, the
  next PR with a `Closes #N` trailer in its `==COMMIT_MSG==`
  block lands with the linked issue transitioning to CLOSED
  with a `Closed by merge of PR #<P> (squash commit <sha>).`
  comment posted by `agent-auth-merge-bot[bot]` — without an
  orchestrator-side `gh issue close` step.

### Notes

- ADR 0038 is amended with a new "Decision 6" describing the
  issue-closing responsibility, the keyword set, the
  cross-repo skip behaviour, the warning-only failure
  handling, and the App-permissions widening.
- `docs/release/merge-bot-setup.md` is updated in three places:
  the "What the bot does" enumeration gets a step 6, Step 1's
  permissions list adds `Issues: Read & write` with a one-line
  justification, the verifying-end-to-end checklist mentions
  the new closing-comment signal, and a new "Closing linked
  issues" section walks through the operational details.
- The orchestrator-side `gh issue close` removal is explicitly
  out of scope for this PR — those workflow files live in the
  `claude-skills` repo, not here. The issue body's "Once this
  issue ships..." note is tracked separately.


tests/test_parse_close_keywords.py
```diff
@@ -0,0 +1,358 @@
...
+    assert expected == parser.AUTO_CLOSE_KEYWORDS
+
+
+# --- find_same_repo_issue_numbers --------------------------------------
+
+
+def test_simple_closes_reference_returns_issue_number() -> None:
+    """The canonical `Closes #N` git trailer must produce one match."""
+    body = "Subject line.\n\nCloses #429\nSigned-off-by: A <a@example.com>\n"
+    assert parser.find_same_repo_issue_numbers(body) == [429]
```
aidanns, line 82: Claude Reviewer: The parser does not match comma-separated multi-issue forms like `Closes #1, #2, and #3` — only `#1` is captured, because the regex requires a keyword before each `#N`. This matches GitHub UI auto-closer behaviour exactly (the UI also only closes #1 here), so the implementation is correct. However the issue body lists "Multiple linked issues in one PR (`Closes #N\nCloses #M`) are all closed" as an acceptance criterion, and the test suite covers newline-separated and per-keyword-repeated forms but not the comma-separated form. Recommend adding a regression test that pins the behaviour (`Closes #1, #2, and #3` returns `[1]`) to document that contributors who want all three closed must repeat the keyword. Optional: document this nuance in the helper docstring or in `docs/release/merge-bot-setup.md` "Closing linked issues" section.

scripts/parse-close-keywords.py
```diff
@@ -0,0 +1,204 @@
...
+# at a word boundary so prose like "the fixes are landing" doesn't
+# match without a `#N`. Trailing punctuation on the reference (`#N.` —
+# common in the validator's GITHUB_KEYWORD_RE shape) is consumed by the
+# `[.,!?;:]?` non-capturing tail; the issue number itself is what we
+# capture.
+#
+# Group `repo` captures `owner/repo` for cross-repo references; absent
+# for same-repo. Group `n` captures the integer issue number.
+_KEYWORD_GROUP = "|".join(re.escape(kw) for kw in sorted(AUTO_CLOSE_KEYWORDS))
+REFERENCE_RE = re.compile(
```
aidanns, line 101: Claude Reviewer: A body containing "Will close #100 once we ship" is treated as a same-repo close reference for issue #100. The regex anchors on `\b<keyword>\b` plus a `#N` follow-on, regardless of preceding context — so `Will close #100`, `If we close #100`, `did not close #100`, etc. all match. This is consistent with GitHub UI auto-closer behaviour (which also matches `Will close #100`), so the parser is correct. But the practical risk is that a PR body discussing future work (e.g. `Notes: future patches will close #500`) will inadvertently close #500 on merge. Recommend either (a) adding a regression test pinning the current behaviour as intentional, or (b) considering whether the parser should additionally require a beginning-of-line / sentence-leading anchor for a tighter match — though the latter would diverge from GitHub UI behaviour. The pragmatic path is (a) plus a one-line note in the docstring or setup doc warning contributors that any in-body `Closes #N` mention closes the issue.

tests/test_parse_close_keywords.py
aidanns: Claude: Done!

scripts/parse-close-keywords.py
aidanns: Claude: Done!

PR #471, opened 2026-04-29
chore(ci): create nightly.yml orchestrator shell

aidanns:
## Summary
- Adds nightly.yml orchestrator shell, daily cron at 04:00 UTC.
- required-checks-passed aggregator (no children yet — mutation testing migrates in a later issue).

### Test plan
- [ ] Workflow file is syntactically valid (workflow appears in the Actions tab).
- [ ] No existing workflows are modified.

==COMMIT_MSG==
chore(ci): create nightly.yml orchestrator shell

Establish the empty shell for daily-cadence checks under the new
uv-style nested workflow_call CI structure. Triggers on a 04:00 UTC
cron and workflow_dispatch; the required-checks-passed aggregator
has no children yet (mutation testing migrates in a follow-up issue).

Closes #442
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

Issue #440, opened 2026-04-29
ci: migrate to nested workflow_call CI structure (uv-style)


aidanns:
## Goal

Migrate agent-auth's CI from 19 standalone workflow files to a nested `workflow_call` structure modeled on [uv's CI](https://github.com/astral-sh/uv/tree/main/.github/workflows). Each PR-time check becomes a reusable workflow invoked from a single `ci.yml` orchestrator. Daily and weekly cadence checks live under separate `nightly.yml` and `weekly.yml` orchestrators.

## Target structure

```
ci
    plan
    build
    check-docs                       (always-passing placeholder)
    check-fmt
        spdx-license-headers
    check-lint
        python
        systems-engineering          (verify-design + verify-function-tests)
    check-security
        codeql-analyse
        dependency-review
            - dependency-check
        dependency-submission
        ripsecrets
    check-standards                  (absorbs verify-standards)
    check-pull-request
        - developer-certificate-of-origin
        - title-prefix
        - title-prose-style
        - commit-message
    check-changelog
        - exists
        - format
    check-publish                    (uv publish --dry-run + twine check)
    check-release                    (build dist + dry-run GH release upload)
    test-unit                        (+ verify-token-cli-http-parity as a test)
    test-integration
    test-smoke
    test-system

nightly
    mutation

weekly
    open-ssf-scorecard
    bench

merge-bot
    merge-pr
    sweep-after-merge
changelog-bot
    - release-impact-comment
release-bot
dependabot-adaptor-bot
```

## Approach

Step-by-step migration: establish empty orchestrator shells first, then bots, then migrate each existing workflow into the new structure one at a time. Each existing workflow's required-status-check stays green throughout the migration; the final cutover happens in Phase 5 when branch protection is switched to require only `required-checks-passed`.

## Decisions (locked in)

- Independent builds for `build` vs `check-release` (no shared artifact handoff).
- `verify-token-cli-http-parity` becomes a unit test, not a workflow.
- `bench` stays weekly (not PR-time).
- pr-lint validators are extracted into a standalone same-monorepo package; CI calls the published version.
- Pr-lint's `release-impact-comment` job moves to `changelog-bot.yml`.
- One repo-wide `required-checks-passed` aggregator at the top of `ci.yml` (no per-workflow aggregators).
- `check-docs` is a placeholder that always passes (no docs site today).

## References

- Originating audit: #414
- Modeled on: https://github.com/astral-sh/uv/tree/main/.github/workflows

[...]

PR #472, opened 2026-04-29
chore(ci): create weekly.yml orchestrator shell

### LINKED ISSUE #443: ci: create weekly.yml orchestrator
Part of #440.

## Scope
Create the `weekly.yml` orchestrator that will host weekly-cadence checks (OSSF Scorecard, bench). Empty shell on creation.

## Plan
- Create `.github/workflows/weekly.yml` with:
  - `on: schedule (cron 05:00 UTC Sunday), workflow_dispatch`
  - `required-checks-passed` aggregator job (no children yet)
- No existing workflows are modified or removed

## Dependencies
Independent — no other migration issues need to land first.

aidanns:
## Summary
- Adds weekly.yml orchestrator shell, weekly cron at 05:00 UTC Sunday.
- required-checks-passed aggregator (no children yet — OSSF Scorecard and bench migrate in later issues).

### Test plan
- [ ] Workflow file is syntactically valid (workflow appears in the Actions tab).
- [ ] No existing workflows are modified.

==COMMIT_MSG==
chore(ci): create weekly.yml orchestrator shell

Establish the empty shell for weekly-cadence checks under the new
uv-style nested workflow_call CI structure. Triggers on a 05:00 UTC
Sunday cron and workflow_dispatch; the required-checks-passed
aggregator has no children yet (OSSF Scorecard and bench migrate in
follow-up issues).

Closes #443
==COMMIT_MSG==

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -7,15 +7,20 @@ name: Weekly
 # Orchestrator shell for weekly-cadence checks (OSSF Scorecard, bench)
 # under the uv-style nested workflow_call CI structure (issue #440).
 # Currently an empty shell — the `required-checks-passed` aggregator
-# has no children yet. Scorecard and bench migrate under here in
-# follow-up issues. Triggers on a 05:00 UTC Sunday cron so weekly
-# results land early in the week, and on workflow_dispatch so a
-# maintainer can trigger an on-demand run.
+# runs alone and trivially passes so the empty shell can land
+# independently. Scorecard and bench migrate under here in follow-up
+# issues; the forward-looking `if: always()` + `needs.*.result`
+# aggregator pattern lands when the first child does. Triggers on a
+# 05:00 UTC Sunday cron so weekly results land early in the week, and
+# on workflow_dispatch so a maintainer can trigger an on-demand run.
 on:
+  workflow_dispatch:
   schedule:
     - cron: "0 5 * * 0" # 05:00 UTC every Sunday
-  workflow_dispatch:
 
+# Don't cancel an in-flight scheduled run when a manual dispatch arrives:
+# weekly checks are expensive and the in-flight result is still useful.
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly.yml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  required-checks-passed:
+    name: Required checks passed
+    runs-on: ubuntu-latest
+    # `if: always()` + `needs.*.result` is the pattern future child
+    # jobs will use; encoded now (with `needs: []`) so adding the
+    # first child is a one-line change to `needs:` rather than a
+    # restructuring of the aggregator.
+    if: always()
+    needs: []
```
aidanns: Claude Reviewer: Structural divergence from sibling shell PR #471 (`nightly.yml`). That PR's `required-checks-passed` aggregator is a trivial one-step `run: echo "all required checks passed"` with no `if: always()`, no `needs: []`, and no jq-based result aggregation; this PR encodes the full forward-looking aggregator (`if: always()` + `needs: []` + `toJSON(needs)` + jq) up front. Both are empty shells per the same parent issue (#440), so the divergence reads as accidental rather than deliberate. The block comment here justifies the choice on its own merits, but does not address the inconsistency between siblings — flagging per the explicit "consistency with PR #471" check. Suggest either (a) downgrade this aggregator to match nightly.yml so the empty shells are uniform, then upgrade both together when the first child lands; or (b) plan to upgrade nightly.yml to this richer pattern in a follow-up so the two converge. Either way, the inconsistency itself is the concern.

aidanns: Claude: Done — downgraded the aggregator to match nightly.yml`s trivial echo pattern. Both empty shells are now uniform; the forward-looking aggregator pattern will land in a follow-up when the first children migrate under weekly.yml.

PR #481, opened 2026-05-01
chore(ci): add check-docs as always-passing placeholder

### LINKED ISSUE #464: ci: add ci/check-docs as always-passing placeholder
Part of #440.

## Scope
Add a `check-docs.yml` (workflow_call) that is a placeholder always-passing job. agent-auth has no docs site today, but the slot is reserved in the new structure for when one is added.

## Plan
- Create `.github/workflows/check-docs.yml` (workflow_call) with a single job that exits 0
- Add a top-of-file comment explaining the placeholder, with a `// TODO(aidanns):` marker noting it should be replaced with an actual docs build when agent-auth gains a docs site
- Wire from `ci.yml`'s `required-checks-passed`

## Dependencies
- Depends on #441

aidanns:
==COMMIT_MSG==
chore(ci): add check-docs as always-passing placeholder

Reserve the `check-docs` slot in `ci.yml`'s nested workflow_call
structure (issue #440) so subsequent migrations land against a stable
shape. agent-auth has no docs site today, so the child workflow runs a
single no-op step that exits zero. A TODO(aidanns) marker in the file
header points at the future replacement (e.g. `mkdocs build --strict`)
once a docs site is introduced.

The aggregator in `ci.yml` treats `skipped` alongside failure for any
required check, so the job runs unconditionally rather than gated
behind an `if:` that could short-circuit to `skipped`. The caller job
is wired into `required-checks-passed`'s `needs:` list so the
aggregate status check covers it from day one.

Closes #464
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

CI-only placeholder; no user-visible behaviour change. The new child
workflow runs a single `echo` step so the aggregator sees a `success`
result rather than a `skipped` one (the aggregator treats `skipped` as
a failure). The caller job in `ci.yml` is wired into
`required-checks-passed`'s `needs:` list so the aggregate status check
covers it from day one.

### Test plan

- [ ] `Required checks passed` reports green on this PR (proves the
      caller's `needs: [plan, check-docs]` aggregator wiring works
      end-to-end).
- [ ] The `Check Docs / check-docs` job appears in the PR's checks
      list and reports `success` (not `skipped`).
- [ ] `reuse.yml` passes (proves the SPDX header on the new file is
      well-formed).

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
+#
+# SPDX-License-Identifier: MIT
+
+name: Check Docs
+
+# Placeholder docs-build child workflow for the uv-style nested
+# `workflow_call` CI structure (parent issue #440). agent-auth has no
+# docs site today, but the slot is reserved in `ci.yml`'s aggregator so
+# the structure stays uniform across migration issues — when a docs
+# site is introduced this workflow will run the real build, and no
+# wiring change in `ci.yml` will be needed.
+#
+# The aggregator in `ci.yml` treats `skipped` as a failure alongside
+# `failure` / `cancelled` / `timed_out`, so this job must actually run
+# and report `success`. Do not gate it behind any `if:` that could
+# resolve to a `skipped` outcome.
+#
+# TODO(aidanns): replace this no-op with a real docs build (e.g.
+# `mkdocs build --strict`) once agent-auth gains a docs site.
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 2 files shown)


aidanns:
Claude Reviewer: LGTM. Verified SPDX header (year + MIT), `TODO(aidanns):` marker, no `if:` on the new `check-docs` caller in `ci.yml` (no skip risk), no third-party `uses:` needing SHA pin, PR title 55 chars with `chore(ci):` prefix, `==COMMIT_MSG==` block has contiguous `Closes #464`/`Signed-off-by:` trailers and `==NO_CHANGELOG==` marker, and only the two intended files are touched (no scope creep).

Issue #511, opened 2026-05-02
ci: pr-lint.yml + composite action changes can't be tested in the PR that introduces them


aidanns:
## Context

`.github/workflows/pr-lint.yml` runs on `pull_request_target`, which is intentional: it is the only safe trigger that gives the lint job repo-secret access without granting fork PRs the ability to weaken the validator (see PR #506's commit message and `.claude/instructions/tooling-and-ci.md`).

The downside is that `pull_request_target` checks out the **base ref's** copy of the workflow + composite action at `.github/actions/install-pr-lint-validator/`. So a PR that modifies either of those two surfaces is tested against the *previous* (pre-PR) version — the new version is never exercised on its own PR.

## Concrete failure

PR #506 cut the pr-lint jobs over to a new local composite action (`install-pr-lint-validator`). The new composite action contained a load-time syntax error: a literal `${{ github.token }}` expression inside an input `description:` field, which GitHub rejects with `Unrecognized named-value: 'github'` because `action.yml` metadata only has the `inputs` context in scope at parse time.

PR #506 went green and merged, because its own CI ran the *old* (in-tree-script) pr-lint.yml from main. Once it landed, every PR opened against main (e.g. #510) had every PR-Lint job die at the "Install pr-lint-validator from release" step. The follow-up fix PR removes the offending expression, but the structural gap that hid the bug from #506's CI remains.

## What we want

Some form of CI coverage that runs on the *head ref's* copy of pr-lint.yml + the composite action, on every PR that touches either, so a load-time syntax error in those files fails CI in the same PR that introduces it.

Sketch options (not committing to any):

- A separate `actionlint` job triggered on `pull_request` (not `pull_request_target`), scoped to `.github/workflows/pr-lint.yml` + `.github/actions/install-pr-lint-validator/`. Forks won't get repo secrets but actionlint doesn't need them — it's pure-static.
- A small unit test that loads action.yml files via `yaml.safe_load` plus a hand-rolled check that no `description:` / `default:` value contains a literal `${{`. Runs as a normal `pull_request` test job.
- Adopting an existing CI tool (reviewdog/actionlint, ghalint) configured to fail on the same patterns.

Pick whichever fits cleanest with the existing CI scaffolding.

## Out of scope

- Backporting tests for the bug fixed in the follow-up PR — that fix is one-line and the regression is now permanent-evidence in the git log.
- Anything that changes pr-lint.yml's `pull_request_target` trigger.

## Recovered context

[...]

PR #506, opened 2026-05-02
improvement: pr-lint.yml uses released pr-lint-validator wheel

aidanns:
==COMMIT_MSG==
#489 extracted the PR-title and ==COMMIT_MSG== validators into a
released `pr-lint-validator` workspace package, but pr-lint.yml's
jobs were still invoking the in-tree `scripts/validate-*.py`
duplicates. The two surfaces drift over time — a fix in one half
silently lapses past the other (this is exactly what #500 had to
backport into the wheel after rule 8 landed in tree first).

Cut the five jobs that consume validator scripts (`pr-title-style`,
`pr-body-commit-msg`, `validator-self-test`, `pr-title-self-test`,
`pr-body-warning-self-test`) over to the v0.17.1 wheel published on
the GitHub Release. A new local composite action at
.github/actions/install-pr-lint-validator/ encapsulates the
fetch + sha256-verify + venv-install boilerplate so each consuming
job collapses to two steps. The composite is base-ref pinned in
every consumer (`pull_request_target` shape) so a fork PR cannot
weaken the install path in the same PR; the wheel itself is immune
to fork tampering by construction.

scripts/validate-pr-title.py is deleted — no other in-tree caller
imports from it. scripts/validate-commit-msg-block.py stays because
the merge-bot's extract-commit-msg-block.py and the changelog-bot
both `importlib`-load it as a parser surface; cutting those bots
over to the wheel is its own piece of work and is captured in the
plan's "Decision" section.

The release-impact-comment / predict-release-impact-self-test
jobs are unchanged: the wheel reserves a `release-impact`
subcommand but #489 explicitly defers packaging
predict_release_impact.py. A follow-up issue tracks the swap once
the subcommand lands.

Closes #477
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

### Test plan

- [ ] CI: every existing required check on this PR passes against the v0.17.1 wheel — proves the live CLI invocation path end-to-end.
- [ ] CI: validator-self-test exits 0 against every fixture under .github/workflows/tests/pr-lint-fixtures/*.md, including invalid-leading-subject-line.md (rule 8 backport via #500 reaches the released wheel surface).
- [ ] CI: pr-title-self-test exits 0 against the wheel's title --self-test (29 cases).
- [ ] CI: pr-body-warning-self-test exits 0 against the wheel's commit-msg --self-test (10 cases).
[...]

PR #512, opened 2026-05-02
fix(ci): drop literal ${{ }} from install-pr-lint-validator description

aidanns:
==COMMIT_MSG==
GitHub Actions evaluates `${{ ... }}` expressions inside action.yml
metadata fields, but only the `inputs` context is in scope at
metadata-load time. The literal `${{ github.token }}` reference in
the `github-token` input description (added by #506) made the
runner reject the manifest with `Unrecognized named-value:
'github'`, breaking every PR-Lint job that consumes the composite
on every PR opened against main since #506 landed (#510 was the
first observed casualty).

#506's own CI did not catch it: pr-lint.yml runs on
`pull_request_target` and explicitly checks out the base ref's
copy of the workflow + composite, so the new file never ran on
the PR that introduced it. The structural gap is tracked in #511.

Drop the `${{ ... }}` braces from the prose so the description is
no longer parsed as an expression. Add a short YAML comment above
the field so future editors do not reintroduce the syntax.

==COMMIT_MSG==

## Review notes

PR Lint failure on main:

- Originating PR: #506 (merged 2026-05-02).
- First observed casualty: #510 — every consumer of the composite (`pr-title-style`, `pr-body-commit-msg`, `validator-self-test`, `pr-title-self-test`, `pr-body-warning-self-test`) failed at the "Install pr-lint-validator from release" step with `Unrecognized named-value: 'github'`. Failed run: https://github.com/aidanns/agent-auth/actions/runs/25242007784.

### Self-validation gap

This PR's own CI **cannot** validate the fix. `.github/workflows/pr-lint.yml` runs on `pull_request_target` and pins `actions/checkout@…` to `${{ github.event.pull_request.base.sha }}`, so the composite resolves against `main` (still broken) rather than the head ref of this PR. Every PR-Lint job here is expected to fail with the same manifest-load error. After this PR merges, main holds the fixed action.yml and subsequent PRs should pass.

The structural gap is tracked separately in #511 — a follow-up will add some form of head-ref-aware lint (actionlint on a `pull_request` trigger, or a unit test that loads each `action.yml` and asserts no `${{ }}` appears in metadata fields) so a future regression of the same shape fails CI in the PR that introduces it.

### Test plan

- [x] CI on this PR: every non-PR-Lint required check still passes (PR-Lint failures are expected — see "Self-validation gap" above). Run: https://github.com/aidanns/agent-auth/pull/512/checks — confirmed 30+ checks green, 5 expected PR-Lint failures, no unexpected failures.
- [ ] After merge: open or rebase any subsequent PR and confirm every PR-Lint job runs to completion (i.e. reaches the actual validator, not the manifest-load error). Acceptable outcomes: validator-pass on a clean PR, or validator-fail on a deliberately bad PR — what matters is that the "Install pr-lint-validator from release" step no longer aborts at manifest load.

### DIFF
--- a/.github/actions/install-pr-lint-validator/action.yml
+++ b/.github/actions/install-pr-lint-validator/action.yml
@@ -34,9 +34,12 @@ inputs:
     required: false
     default: "0.17.1"
   github-token:
+    # Don't write `${{ ... }}` here — action.yml metadata only has
+    # the `inputs` context in scope, so `${{ github.token }}` fails
+    # to load with "Unrecognized named-value: 'github'".
     description: >
       GitHub token used to download the wheel asset. Callers
-      typically pass the workflow's `${{ github.token }}`, which
+      typically pass the workflow's `github.token`, which
       has `contents: read` and is sufficient for downloading
       release assets from the same private repository.
     required: true


aidanns:
Claude Reviewer: LGTM. Independently verified the fix and the surrounding claims:

- **Root cause**: confirmed. Action metadata fields are evaluated at manifest-load with only the `inputs` context in scope, so a literal `${{ github.token }}` in a `description:` field is correctly rejected with `Unrecognized named-value: 'github'`. Stripping the braces and falling back to backticked prose (`\`github.token\``) is the right fix and the description still reads cleanly.
- **YAML comment**: well-placed directly above the affected field, names the exact error string a future editor would hit, and is short enough not to bury the `description:` it's annotating. Nice.
- **Self-validation gap claim**: spot-checked `.github/workflows/pr-lint.yml` — every job runs on `pull_request_target` and pins `actions/checkout` to `${{ github.event.pull_request.base.sha }}`, so the description's explanation for why this PR's own PR-Lint runs are expected to fail is accurate.
- **#511 scoping**: agreed this is the right granularity. The composite-load regression and the meta-bug that lets such a regression slip past CI on its introducing PR are genuinely separate concerns; #511 is structural follow-up, not a punted-fixable concern.

No blocking comments, no nits. Ship it once required non-PR-Lint checks are green.

PR #521, opened 2026-05-02
chore(ci): split error-taxonomy contract per package

### LINKED ISSUE #519: test-unit shards: things_client_fake import fails for agent-auth (per-package shard architecture)
## Observed

After Phase 5 (#516) restores CI startup, `test-unit / unit-tests (agent-auth)` fails at collection:

\`\`\`
ImportError while importing test module '.../packages/agent-auth/tests/test_error_taxonomy.py'.
packages/agent-auth/tests/test_error_taxonomy.py:51: in <module>
    from things_client_fake.store import FakeThingsClient, FakeThingsStore
E   ModuleNotFoundError: No module named 'things_client_fake'
\`\`\`

`things_client_fake` is a test-only fake shipped with the things-bridge package's tests/ tree (see `packages/things-bridge/tests/things_client_fake/`). The agent-auth test imports it cross-package.

## Why this is latent today

Per #491 the per-package shards run \`pytest packages/<package>/tests/ --no-cov\` directly (not \`task test --unit\`) to bypass the workspace-wide coverage floor enforced by \`scripts/check-package-coverage.sh\`. This isolates each shard to that package's own dev extras — but the agent-auth test depends on things-bridge's test-only fake, which is NOT a dev extra of agent-auth.

The bug has been latent since #491 merged (2026-05-02 03:35 UTC) because the CI workflow has been failing at \`startup_failure\` since #508 landed before it (2026-05-02 02:37 UTC) — issue surfaced once #516 fixed the permissions chain.

## Fix options

1. **Move the cross-package test out of `agent-auth/tests/`.** The test exercises things-bridge's fake against agent-auth's error taxonomy — it could live under \`tests/integration/\` or under \`packages/things-bridge/tests/\`.
2. **Install all dev extras across the workspace in the shard.** Defeats the per-package isolation that #491 introduced for parallelism.
3. **Add a tests-only \`things_client_fake\` install path that the agent-auth shard can pull in.** Keeps isolation but requires a pyproject change.
4. **Revert the per-package shard split.** Goes back to the single \`unit\` job that runs \`task test --unit\` workspace-wide.

Suggested: option 1 is the smallest change with the cleanest semantics.

## References

- Originating PR: #491 (the test.yml split implementation note flags the workspace-coverage-floor bypass).
- Discovered while landing the Phase 5 cutover #516.
- Related: #518 (test-* not in aggregator \`needs:\`, which is why this hadn't surfaced as a blocker on main).

aidanns:
==COMMIT_MSG==
The per-package unit-test shards introduced in #491 each run only
against ``packages/<package>/tests/`` with that package's own
``pyproject.toml`` rootdir, so the workspace-wide pythonpath that used
to surface ``things_client_fake`` to every shard no longer applies.
``packages/agent-auth/tests/test_error_taxonomy.py`` therefore failed
the agent-auth shard at collection time once Phase 5 of the test.yml
split (#516) re-enabled CI: the file imported ``things_client_fake``
to drive things-bridge's half of the taxonomy, and that module ships
under ``packages/things-bridge/tests/``.

Split the contract along the same per-package boundary that the
shards already enforce. The agent-auth file keeps the agent-auth half
and loses every cross-package import; a sibling file under
``packages/things-bridge/tests/`` carries the things-bridge half and
keeps the ``things_client_fake`` dependency confined to the package
that already ships it. Each docstring cross-references the other so
future readers see the full taxonomy at a glance, and
``scripts/verify-standards.sh`` now grep-checks both files together
so the workspace-wide guarantee (every documented error code is
exercised) stays intact.

Closes #519
==COMMIT_MSG==

==NO_CHANGELOG==

### Why split the file (option 1) instead of moving it whole

Both halves of the contract are about each package's own server, so
the natural per-package boundary is per-test-class — not per-file.
Splitting:

- Preserves the per-package test-shard isolation #491 introduced
  (each shard imports only its package's own test fixtures).
- Lets each package's coverage gate measure its own taxonomy
  enforcement (things-bridge keeps its 83% floor; agent-auth keeps
  its 78% floor).
- Avoids forcing one package's shard to depend on the other's
  ``tests/`` tree as a python path — which the issue explicitly
  flagged as a smell that #491's per-package isolation is meant to
  prevent.

The alternative (move the whole file under ``packages/things-bridge/tests/``
or to a workspace-level ``tests/integration/``) would either
re-create the cross-package dependency in the opposite direction
or introduce a new test-tree that nothing else uses.

### Test plan

- [ ] `task check-standards` (verify-standards confirms both contract
      files exist and together reference every documented error code).
- [ ] `uv run --no-sync pytest --no-cov --ignore=packages/agent-auth/tests/integration packages/agent-auth/tests/`
      (replicates the `test-unit / unit-tests (agent-auth)` shard;
      collection no longer fails on the missing
      ``things_client_fake`` import).
- [ ] `uv run --no-sync pytest --no-cov --ignore=packages/things-bridge/tests/integration packages/things-bridge/tests/`
      (things-bridge shard still passes with the new file added).
- [ ] `uv run --no-sync mypy` (no duplicate-module collision between
      the two shards' ``test_*_taxonomy`` modules).

### DIFF
--- a/packages/agent-auth/tests/test_error_taxonomy.py
+++ b/packages/agent-auth/tests/test_error_taxonomy.py
@@ -2,15 +2,25 @@
 #
 # SPDX-License-Identifier: MIT
 
-"""Contract tests for the error-code taxonomy.
+"""Contract tests for the agent-auth half of the error-code taxonomy.
 
-Every error code documented in ``design/error-codes.md`` is exercised here.
-Each test triggers the precise condition that should produce a given code and
-asserts the response body. Changes to error strings or HTTP statuses will
-fail these tests, which is the intent: the error taxonomy is public API.
+The error taxonomy is split across two packages — agent-auth's response
+codes are exercised here, and things-bridge's response codes are
+exercised in
+``packages/things-bridge/tests/test_things_bridge_error_taxonomy.py``.
+The split keeps each package's unit-test shard self-contained (the
+bridge tests need ``things_client_fake``, which is shipped under
+``packages/things-bridge/tests/`` as a test-only fake and is not on the
+agent-auth shard's pythonpath).
 
-Documented codes
-----------------
+Every error code documented in ``design/error-codes.md`` is exercised by
+exactly one of the two files. Each test triggers the precise condition
+that should produce a given code and asserts the response body. Changes
+to error strings or HTTP statuses will fail these tests, which is the
+intent: the error taxonomy is public API.
+
+Documented codes covered here
+-----------------------------
 agent-auth /v1/validate:
   malformed_request (400), invalid_token (401), token_expired (401),
   token_revoked (401), scope_denied (403).
@@ -28,18 +38,6 @@ agent-auth /metrics (unversioned):
   scope_denied (403).
 agent-auth server-wide:
   not_found (404), rate_limited (429).
-things-bridge /v1/* data endpoints:
-  unauthorized (401), token_expired (401), scope_denied (403),
-  authz_unavailable (502), not_found (404),
-  things_permission_denied (503), things_unavailable (502).
-things-bridge /health (unversioned):
-  unauthorized (401), token_expired (401), scope_denied (403),
-  authz_unavailable (502).
-things-bridge /metrics (unversioned):
-  unauthorized (401), token_expired (401), scope_denied (403),
-  authz_unavailable (502).
-things-bridge server-wide:
-  not_found (404), method_not_allowed (405), rate_limited (429).
 """
 
 import os
@@ -48,7 +46,6 @@ import threading
 from datetime import UTC, datetime, timedelta
 
 import pytest
-from things_client_fake.store import FakeThingsClient, FakeThingsStore
 
 from agent_auth.approval import ApprovalManager
 from agent_auth.approval_client import ApprovalClient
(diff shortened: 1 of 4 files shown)


aidanns:
Claude Reviewer: LGTM. Verified: PR title (`chore(test):` prefix, 51 chars); PR body has `==COMMIT_MSG==` with `Closes #519` immediately followed by `Signed-off-by:` (no blank line); commit is DCO-signed; `no changelog` label applied. Scope is minimal (test split + 1-line mypy module list addition + verify-standards.sh update). Cross-package imports (`things_client_fake`, `agent_auth_client`, `things_bridge.*`) are fully removed from the agent-auth shard, preserving #491 per-package isolation. The new `packages/things-bridge/tests/test_things_bridge_error_taxonomy.py` is byte-for-byte identical to the bridge half of the original (no silent regression in `_FakeAuthz`, `_InjectableThings`, `bridge_server`, or any of the 16 bridge tests). things-bridge shard already has `agent_auth_client` and `tests_support` on its pythonpath via the agent-auth-common dep, so discovery works. `verify-standards.sh` now grep-checks both files together so the workspace-wide taxonomy guarantee is intact (and `rate_limited` continues to be exercised in both halves).

Issue #523, opened 2026-05-02
Update stale test path references in codeql-config and ADR 0040 after #520 relocation


aidanns:
## Context

#520 relocated several gpg-bridge test files from
``packages/gpg-bridge/tests/`` to ``packages/gpg-bridge/tests/integration/``
to fix per-package unit-shard failures. Two reference comments still
point at the pre-move paths:

- ``.github/codeql/codeql-config.yml`` line 65 references
  ``packages/gpg-bridge/tests/test_gpg_subprocess_client_passphrase.py::test_passphrase_absent_in_outputs``.
  After #520 the file lives at
  ``packages/gpg-bridge/tests/integration/test_gpg_subprocess_client_passphrase.py``.
  Comment-only; doesn't affect codeql analysis. Touched separately
  because ``.github/`` paths are gated for agent-driven changes.

- ``design/decisions/0040-gpg-bridge-docker-integration-tests.md``
  line 151 references
  ``test_gpg_bridge_server.py``. The file moved but the test
  name is unchanged, so the ADR text still reads correctly; only
  the implicit path under ``packages/gpg-bridge/tests/`` is now
  stale. Trivial cleanup.

## Why it was deferred

``.github/`` updates are out of scope for the agent pipeline that
implemented #520. The ADR cleanup is a touch-up that makes more
sense as a single sweep alongside the codeql-config edit.

## References

- Originating PR: #520

aidanns:
Claude: Closed by /workflow:implement session wfi-2026-05-02-7ebf — superseded by #522.

Per the resolution recorded on #522, that PR will relocate the relocated gpg-bridge tests back from `packages/gpg-bridge/tests/integration/` to `packages/gpg-bridge/tests/`. Once that lands, the path references in `.github/codeql/codeql-config.yml` line 65 and `design/decisions/0040-gpg-bridge-docker-integration-tests.md` line 151 are valid again — no standalone cleanup needed.

Issue #522, opened 2026-05-02
Fix PYTHONPATH propagation for test-unit shards (eliminate test-layout workaround)


aidanns:
## Decisions (resolved 2026-05-02 in /workflow:implement session wfi-2026-05-02-7ebf)

- **Fix approach**: option (b) — move the `PYTHONPATH` export into `.github/actions/setup-toolchain` as the single source of truth across all workflows. This removes the dependency on `scripts/_bootstrap_venv.sh` for shard environments while keeping the bootstrap script unchanged for local use. Other workflows that already use `setup-toolchain` (verify-standards, codeql, etc.) inherit the export automatically.
- **Scope**: include the #520 revert in this PR. Specifically: relocate the relocated tests back from `packages/gpg-bridge/tests/integration/` to `packages/gpg-bridge/tests/`, and restore `--cov-fail-under=62` in `packages/gpg-bridge/pyproject.toml` (was lowered to `55` as the workaround in #520). End-to-end fix in one PR.
- **Supersedes #523**: the codeql-config and ADR 0040 path references that #523 was tracking become valid again once the tests are relocated back. #523 will be closed as superseded.

## Context

#520 worked around a per-package `test-unit.yml` shard failure for
gpg-bridge by relocating the gpg-subprocess + HTTP wiring tests
under `packages/gpg-bridge/tests/integration/`. The actual root
cause is that the per-package shard's bare ``pytest`` invocation
does not source ``scripts/_bootstrap_venv.sh``, which is the only
place the workspace exports the ``packages/<package>/tests`` paths
on ``PYTHONPATH`` — so subprocesses spawned by tests (e.g. the
``gpg_backend_fake`` and ``things_client_fake`` modules launched
via ``python -m <fake>``) can't resolve their module from the
shard's environment. ``pytest``'s ``pythonpath`` config only
affects the in-process ``sys.path``, not subprocess
``PYTHONPATH``.

## Why it was deferred

#520 explicitly chose the test-layout workaround
(option 1 in the issue body). Touching ``.github/workflows/`` is
out of scope for the agent-driven pipeline (CI/security-config
paths require user review). This issue tracks the proper fix.

## Suggested fix

Either:

1. Source ``scripts/_bootstrap_venv.sh`` in
   ``.github/workflows/test-unit.yml``'s ``Run unit tests`` step
   (or factor out just the ``PYTHONPATH`` export so the bootstrap
   step doesn't run twice — sync is already separate).
[...]

PR #524, opened 2026-05-02
chore(ci): move gpg-bridge subprocess tests to integration shard

### LINKED ISSUE #520: test-unit shards: gpg-bridge unit tests fail (gpg subprocess + 502 errors in per-package shard)
## Observed

After Phase 5 (#516) restores CI startup, \`test-unit / unit-tests (gpg-bridge)\` fails 19 of 95 tests with a mix of:

- HTTP 502 / 503 errors in server tests:
  \`assert 502 == 200\`, \`urllib.error.HTTPError: HTTP Error 502: Bad Gateway\`
- gpg subprocess errors:
  \`gpg sign exited 1 without a recognised error\`

Failing test files include \`test_bridge_network_drop.py\`, \`test_gpg_bridge_perf_budget.py\`, \`test_gpg_bridge_server.py\`, \`test_gpg_subprocess_client.py\`, \`test_gpg_subprocess_client_passphrase.py\`.

## Why this is latent today

Per #491 the per-package shards run \`pytest packages/<package>/tests/ --no-cov\` directly. The gpg-bridge tests exercise a real gpg subprocess (or expect the bridge to spawn one) — the per-package shard's environment doesn't include the gpg setup steps that the previous \`unit\` job in test.yml had (apt-get install gnupg, key import, etc.).

Hidden since #491 (2026-05-02 03:35 UTC) because CI startup failure from #508 prevented the matrix from running at all on main; surfaced when #516 restored startup.

## Fix options

1. **Move the gpg subprocess tests into \`tests/integration/\`** so the integration shard (which already sets up the gpg toolchain) exercises them, and the unit shard skips them.
2. **Add a \`Setup gpg\` step to test-unit.yml's matrix when \`<package> == 'gpg-bridge'\`.** Conditional on package name; bloats the per-package shard.
3. **Revert the unit/integration split for gpg-bridge** so all gpg-bridge tests run under one shard with the full toolchain.

Suggested: option 1 — test_gpg_subprocess_client.py and test_gpg_bridge_server.py exercise real subprocess + HTTP wiring; that's integration-test territory.

## References

- Originating PR: #491.
- Discovered while landing the Phase 5 cutover #516.
- Related: #518 (test-* not in aggregator \`needs:\`).

aidanns:
==COMMIT_MSG==
The per-package `test-unit / unit-tests (gpg-bridge)` shard
introduced in #491 fails 19 of 95 tests after #516 restored CI
startup. The shard runs raw `pytest packages/<pkg>/tests/ --no-cov`
without sourcing `scripts/_bootstrap_venv.sh`, which is the only
place the workspace exports `packages/<pkg>/tests` on `PYTHONPATH`
so subprocesses launched by tests (e.g. `python -m
gpg_backend_fake`) can resolve the test-only fakes. Pytest's own
`pythonpath` config only affects the in-process `sys.path`, not
subprocess `PYTHONPATH`, so the failing tests greet the bridge with
`gpg sign exited 1 without a recognised error`,
`urllib.error.HTTPError: 502 Bad Gateway`, etc.

Per #520's preferred fix (option 1), relocate the gpg subprocess +
HTTP wiring tests under `packages/gpg-bridge/tests/integration/` so
the integration shard (`test-integration.yml`, which goes through
`task test --integration` and therefore sources the bootstrap)
exercises them with the right environment. The unit shard's
`--ignore=packages/<pkg>/tests/integration` already excludes the
relocated tests from its sweep.

Files moved:

- test_gpg_bridge_server.py
- test_gpg_bridge_perf_budget.py
- test_gpg_subprocess_client.py
- test_gpg_subprocess_client_passphrase.py
- tests/fault/test_bridge_network_drop.py
- the gpg_backend_fake-driven wedge from
  tests/fault/test_backend_subprocess_hang.py (extracted into
  tests/integration/test_wedged_gpg_subprocess.py so the rest of
  that fault file -- script-on-disk substitutes that don't need
  PYTHONPATH propagation -- keeps running in the unit shard)

The relocated tests bind to ephemeral `127.0.0.1` ports rather
than Docker-routed names, which
`scripts/verify-integration-isolation.sh` treats as a regression by
default. Carve out a per-file marker
(`# integration-isolation: in-process-server`) so genuinely
in-process tests opt out while the deny-list stays the default for
new Docker-backed tests.

The gpg-bridge `--cov-fail-under` floor drops from 62 to 55: the
workspace coverage gate
(`scripts/check-package-coverage.sh`) only counts unit-shard
coverage, so moving 33 tests out of unit takes the observed
coverage from above 62 to 58. The relocated tests still execute
(under `test-integration.yml`); only their contribution to the
unit-shard coverage gate disappears. The deeper fix -- sourcing
the bootstrap from the per-package `test-unit.yml` shard -- is
`.github/` territory and is tracked as follow-up #522 alongside the
codeql/ADR comment cleanup #523.

Closes #520
==COMMIT_MSG==

==NO_CHANGELOG==

### Why option 1 (and not options 2/3)

The issue body proposed three fix options. Option 1 (relocate the
tests under `tests/integration/`) was the chosen approach because:

- Option 2 (add a `Setup gpg` step conditional on `<package> ==
  'gpg-bridge'`) would need a `.github/workflows/test-unit.yml`
  edit, which is gated for agent-driven changes; it also bloats the
  per-package shard.
- Option 3 (revert the unit/integration split for gpg-bridge) is
  more surgery than the failure mode warrants.

The real root cause is PYTHONPATH propagation, not gpg toolchain --
none of the relocated tests need a real `gpg` binary; they all
spawn the in-tree `gpg_backend_fake` Python module. Follow-up
#522 tracks fixing the propagation in `test-unit.yml` so future
similar tests do not need this workaround.

### Why the wedge test was split out of test_backend_subprocess_hang.py

The fault file holds five tests; four use a script-on-disk `gpg`
substitute (no PYTHONPATH propagation needed -- the unit shard runs
them happily), and one drives the `gpg_backend_fake` Python module
through the full HTTP path. Moving the whole file would have taken
4 healthy unit-shard tests offline; extracting just the
PYTHONPATH-dependent test into
`tests/integration/test_wedged_gpg_subprocess.py` keeps the unit
shard's fault coverage intact while the wedge test runs under the
integration shard alongside its cousins.

### Test plan

- [ ] `task test -- --unit` -- the `gpg-bridge` per-package slice
      passes (62 tests).
- [ ] `scripts/check-package-coverage.sh` -- the gpg-bridge floor
      check returns `gpg-bridge 58% (floor 55%)`.
- [ ] Local `pytest packages/gpg-bridge/tests/integration/` with
      bootstrap PYTHONPATH set -- 33 relocated tests pass alongside
      the existing `test_bridge.py` Docker tests (which need Docker
      to run; they collect cleanly without it).
- [ ] `bash scripts/verify-integration-isolation.sh` -- passes.
- [ ] `bash scripts/verify-standards.sh` -- passes.
- [ ] `bash scripts/verify-function-tests.sh` -- passes.
- [ ] `bash scripts/verify-design.sh` -- passes.
- [ ] `treefmt --no-cache --fail-on-change` -- 0 changed.
- [ ] `uv run --no-sync ruff check .` -- passes.
- [ ] `uv run --no-sync mypy` -- 0 errors.


aidanns:
Claude Reviewer: PR title `chore(test): relocate gpg-bridge subprocess+http tests to integration` is rejected by `pr-lint` (the `PR title (subject style)` check is FAILURE on this PR). The repo allowlist `commit_taxonomy.AREA_SCOPES` is `(ci, claude, deps, deps-dev, design, docs, release, security)` — `test` is not in it, so `chore(test):` is not accepted. Same trap the parallel #521 PR hit. Acceptable rewrites for this change: `chore(ci): relocate gpg-bridge subprocess+http tests to integration` (the motivating constraint really is the per-package CI shard) or a bare `chore: relocate gpg-bridge subprocess+http tests to integration`. Update both the PR title and the `==COMMIT_MSG==` subject line in the PR body to keep them in sync.

pyproject.toml
```diff
@@ -442,9 +440,9 @@ tests_dir = [
     "packages/agent-auth/tests/test_scopes.py",
     "packages/agent-auth/tests/test_store.py",
     "packages/agent-auth/tests/test_tokens.py",
+    "packages/gpg-bridge/tests/integration/test_gpg_bridge_server.py",
```
aidanns: Claude Reviewer: This entry adds `packages/gpg-bridge/tests/integration/test_gpg_bridge_server.py` to mutmut's `tests_dir`, but the prose comment a few lines above (lines ~428-434) explicitly says "Integration tests under packages/<svc>/tests/integration/ ... are too slow for the per-mutant loop, and the remaining unit-test files don't add signal on the modules under paths_to_mutate." After this PR `test_gpg_bridge_server.py` is itself an integration test (you moved it under `tests/integration/` and added the `# integration-isolation: in-process-server` marker), so the prose now contradicts the inclusion. Either drop the entry from `tests_dir` (matching the documented rationale that integration tests don't belong in the mutmut loop), or update the comment to clarify why this specific in-process integration test is the deliberate exception (it's in-process and fast, unlike the Docker-backed `test_bridge.py`). Worth confirming the per-mutant runtime hasn't blown out either way.

aidanns: Claude: Done — dropped the entry. The accompanying prose comment already says integration tests under packages/<svc>/tests/integration/ are too slow for the per-mutant loop; carrying the in-process server test as an exception contradicted that without enough mutmut signal lift to justify the contradiction. The remaining gpg-bridge mutmut entries (test_gpg_backend_fake.py, test_gpg_bridge_config.py) keep unit-test coverage on the bridge module.

aidanns:
Claude: Done — title is now `chore(ci): relocate gpg-bridge subprocess+http tests to integration` (`ci` is in the AREA_SCOPES allowlist). The body's `==COMMIT_MSG==` block didn't need editing — that block carries only the commit *body*; the squash-merge subject is taken from the PR title directly.

PR #529, opened 2026-05-02
chore(ci): merge-bot workflow_run includes new CI workflow

aidanns:
==COMMIT_MSG==
The `workflow_run` allowlist still names `Test`, `Check`, `Typecheck`,
and `CodeQL` — workflows retired in #491, #509, and #516 (Phase 5
cutover). None of them fire on `main` anymore, so the post-label
`workflow_run.completed` retry never matches the new umbrella `CI`
workflow that replaced them.

Reproduced on PR #524 (2026-05-02): bot fired once on `automerge`
label while CI was still failing, exited with `Refuse to merge`,
then never re-fired once CI turned green. The PR sat with green
checks and `automerge` applied until a maintainer used the
`gh workflow run merge-bot.yml -f pr_number=<N>` break-glass.

Drop the four retired names and add `CI` so the parent workflow's
single `workflow_run.completed` event triggers the retry. CI is
the `workflow_call` aggregator for every gating check shard
(`check-*`, `test-*`, codeql), so its completion already signals
"every PR-gating shard has settled" — no need to enumerate the
shards individually.

Closes #525
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

This is a one-line-area edit to `.github/workflows/merge-bot.yml`
that swaps four retired workflow names (`Test`, `Check`,
`Typecheck`, `CodeQL`) for the new umbrella `CI` workflow in the
`on.workflow_run.workflows` allowlist. No behaviour changes for
the bot itself — only the set of completion events that re-fire
it.

The bot fixing its own auto-retry trigger is exactly the
"applied-label-early-on-failing-CI" recovery path described in
the issue body, so the orchestrator may need the
`gh workflow run merge-bot.yml -f pr_number=<N>` break-glass on
this very PR. That's expected.

### Test plan

- [ ] `git diff origin/main` shows exactly 4 deletions + 1 insertion in `.github/workflows/merge-bot.yml`
- [ ] `pr-lint` check passes (title scope `ci` is in the allowlist; `==COMMIT_MSG==` block well-formed)
- [ ] `no changelog` label present so `check-changelog/exists` passes
- [ ] All CI checks turn green after the empty-commit re-trigger fires `ci.yml` with the label visible

Issue #347, opened 2026-04-26
fix(merge-bot): dedupe statusCheckRollup by name+latest before evaluating required checks


aidanns:
## Problem

`Inspect required checks` (`.github/workflows/merge-bot.yml:224-274`) queries `gh pr view --json statusCheckRollup` and runs a jq filter that selects every check entry with a failed conclusion:

```jq
map(select((.workflowName // "") != "Merge Bot"))
| map(select(
    ((.conclusion // "") | ascii_upcase | IN("FAILURE","TIMED_OUT","CANCELLED","STARTUP_FAILURE","ACTION_REQUIRED"))
    or ((.state // "") | ascii_upcase | IN("FAILURE","ERROR"))
))
```

`statusCheckRollup` returns one entry per check *run*, not one per check *name*. Any check that has been retried — workflow rerun, post-`reopened` retrigger, fix-then-push — appears multiple times. The filter selects every historical FAILURE without first reducing to the latest run per name, so once a check has *ever* failed on the head SHA, merge-bot refuses to merge forever.

Observed on PR #346: `changelog-lint` failed on the original `opened` event (label-applied-after-create timing), the run was retriggered via close+reopen and passed, but merge-bot continued to refuse the merge with `FAILED_CHECKS: changelog-lint` until I manually deleted the stale failed run via `gh run delete`. The rollup at refusal time:

```
{"conclusion":"FAILURE","name":"changelog-lint","startedAt":"2026-04-26T04:04:06Z"}
{"conclusion":"SUCCESS","name":"changelog-lint","startedAt":"2026-04-26T04:16:24Z"}
```

The bug surfaces any time a required check fails-then-passes on the same head SHA — i.e. exactly the workflow GitHub's "rerun failed jobs" UI is built for.

## Proposed change

In `.github/workflows/merge-bot.yml:250-266`, group rollup entries by `name` (falling back to `context` for legacy commit statuses) and keep only the latest entry per group by `startedAt` before applying the failure / pending filters.

Sketch:

```jq
map(select((.workflowName // "") != "Merge Bot"))
| group_by(.name // .context // "<unnamed>")
| map(max_by(.startedAt // ""))
| map(select( ... existing failure predicate ... ))
```

The same dedupe applies to the `pending` selector (`merge-bot.yml:258-266`) so a stale `IN_PROGRESS` entry from a cancelled run can't keep merge-bot in the wait-clean-exit path indefinitely.

## Alternatives considered

[...]

PR #532, opened 2026-05-02
chore(ci): merge-bot reads required checks from ruleset

### LINKED ISSUE #528: merge-bot: refuse-on-failure evaluates entire rollup, blocks on unrelated/environmental failures
## Observed

`.github/workflows/merge-bot.yml`'s `Inspect required checks` step (lines 391–460 on main) reads the FULL `gh pr view --json statusCheckRollup`, dedupes-by-latest-run, excludes its own `Merge Bot` workflow runs, and refuses to merge on ANY check whose conclusion is in `[FAILURE, TIMED_OUT, CANCELLED, STARTUP_FAILURE, ACTION_REQUIRED]`.

This is stricter than what the repo's branch-protection ruleset itself requires. The active ruleset on `main` (rule id 15002184) lists `Required checks passed` as the only required context; any other failing check would let `gh pr merge --squash` succeed, but the merge-bot's exit-1 refusal blocks the automated path regardless.

## Impact

Reproduced on PR #521 (issue #519, 2026-05-02): the PR had `test-unit / unit-tests (gpg-bridge)` FAILURE because of issue #520 (a separate, parallel PR). The gpg-bridge failure was unrelated to #519's diff (which moved a different test file). `Required checks passed` was SUCCESS (test-* aren't in its `needs:` list — that's what #518 fixed). Branch protection allowed merge. The merge-bot did not — it refused with `Claude: Cannot merge — required check(s) failed: test-unit / unit-tests (gpg-bridge)`. The user manually broke the deadlock with `gh pr merge --squash --admin`.

The same pattern hits any PR where:
- A flaky test fires once and the rollup carries the FAILURE entry until manual rerun.
- A parallel PR introduces a CI regression that surfaces on every other in-flight PR (the "circular dep" pattern from this 2026-05-02 run).
- An environmental check fails (rate-limit, infra outage) — the bot won't merge until the check is rerun.

This is at odds with the orchestrator's automerge gate in `workflow:implement` (Phase 5 step 5), which classifies CI as "acceptable" when "every failing check is environmental / flaky / unrelated to the diff." The orchestrator's classification is correct for the user's intent, but the bot won't honor it — leading to the "set automerge → bot refuses → user has to break-glass or merge by hand" loop the orchestrator now runs into routinely.

## Design principle

The merge-bot should not encode policy about what is required to merge. The repo's branch-protection configuration is the single authoritative source of "what must be SUCCESS for a merge to land." The bot's job is to read that policy, evaluate the rollup against it, and execute the merge once the requirements are met (plus handle BEHIND auto-update). Adding / removing required checks should be a branch-protection edit, not a bot-workflow edit.

Today's bot violates this — it has its own static notion of "any check failure blocks merge," which doesn't agree with branch protection. The fix is to make the bot derive its gating set from repo config, not from a hard-coded heuristic.

Override is intentionally NOT a bot-controllable surface. When a required check is failing for a known-environmental reason and the user wants to merge anyway, the right path is `gh pr merge --squash --admin` invoked manually by the user — *not* a label or workflow input that the orchestrator (or any agent) could apply. The merge gate is the durable safety contract; the bot honours it without exception, and any escape hatch lives at the human-in-the-loop level.

## Recommended fix — derive required checks from branch protection

Replace the current `Inspect required checks` step with two steps:

### Step A — read the required-check contexts from the active ruleset(s)

```yaml
- name: Read required-check contexts from branch protection
  id: required
  env:
    GH_TOKEN: ${{ steps.app-token.outputs.token }}
    BASE_REF: ${{ steps.pr.outputs.base_ref }}   # PR's target branch, usually 'main'
  run: |
    set -euo pipefail
    # Prefer the modern rulesets API (target == 'branch'). Filter to rulesets
    # whose conditions include the PR's base ref, then union their
    # `required_status_checks` contexts. Falls back to the legacy
    # branches/{branch}/protection API for repos that haven't migrated.
    contexts=$(
      gh api "repos/$GITHUB_REPOSITORY/rulesets" --paginate --jq \
        '.[] | select(.target == "branch" and .enforcement == "active") | .id' \
      | while read -r id; do
          gh api "repos/$GITHUB_REPOSITORY/rulesets/$id" \
            --jq '.. | objects | select(.type == "required_status_checks")
                  | .parameters.required_status_checks[]?.context'
        done \
      | sort -u
    )

    if [[ -z "$contexts" ]]; then
      # Legacy branch-protection fallback for repos that pre-date rulesets.
      contexts=$(
        gh api "repos/$GITHUB_REPOSITORY/branches/$BASE_REF/protection" \
          --jq '.required_status_checks.contexts[]?' 2>/dev/null \
        | sort -u
      )
    fi

    {
      echo "contexts<<MERGE_BOT_EOF"
      printf '%s\n' "$contexts"
      echo "MERGE_BOT_EOF"
    } >> "$GITHUB_OUTPUT"
```

### Step B — evaluate only those contexts against the rollup

```yaml
- name: Inspect required checks
  id: checks
  env:
    GH_TOKEN: ${{ steps.app-token.outputs.token }}
    PR_NUMBER: ${{ steps.pr.outputs.pr }}
    REQUIRED_CONTEXTS: ${{ steps.required.outputs.contexts }}
  run: |
    set -euo pipefail
    rollup=$(gh pr view "$PR_NUMBER" --repo "$GITHUB_REPOSITORY" \
             --json statusCheckRollup --jq '.statusCheckRollup')

    fail=()
    pending=()
    while IFS= read -r ctx; do
      [[ -z "$ctx" ]] && continue
      # Per-context: take the latest run by startedAt; consult `conclusion`
      # for Actions check-runs and `state` for legacy commit statuses.
      latest=$(jq -r --arg n "$ctx" '
        [.[] | select((.name // .context) == $n)]
        | sort_by(.startedAt // "")
        | .[-1]
        | (.conclusion // .state // "missing")
      ' <<<"$rollup" | tr '[:upper:]' '[:lower:]')

      case "$latest" in
        success|neutral|skipped) ;;                     # required and satisfied
        ""|missing|in_progress|queued|waiting|pending|requested)
          pending+=("$ctx=$latest") ;;
        *) fail+=("$ctx=$latest") ;;                    # required and failing
      esac
    done <<<"$REQUIRED_CONTEXTS"

    {
      echo "failed<<MERGE_BOT_EOF"; printf '%s\n' "${fail[@]:-}"; echo "MERGE_BOT_EOF"
      echo "pending<<MERGE_BOT_EOF"; printf '%s\n' "${pending[@]:-}"; echo "MERGE_BOT_EOF"
    } >> "$GITHUB_OUTPUT"
```

The downstream `Refuse to merge — required check failed` and `Wait for pending checks (clean exit)` steps stay as-is; they consume `steps.checks.outputs.failed` / `pending` and don't care how those were computed.

## What the bot now does (and doesn't)

Stays the same:
- Squash-merging via `PUT /pulls/{n}/merge` with the `==COMMIT_MSG==` block as the body.
- `BEHIND` → `update-branch` auto-resolution.
- DCO trailer enforcement (operates on the squash-merge body, independent of which checks gate).
- Comment posting on refusal so the user knows what the bot is waiting on.
- `Filter out the merge-bot's own checks` rule (still relevant — the bot's own check-runs shouldn't gate; though under the new design they wouldn't anyway, since `Merge ==COMMIT_MSG==` isn't a required context).
- Dedupe-by-latest-startedAt (#347 fix — still needed because the rollup carries every historical run).

Gone:
- The hard-coded "any FAILURE in the rollup blocks merge" rule.
- Implicit policy decisions about which checks count.

## Properties of the derived-from-config design

- **Policy lives in one place.** Adding a check to branch protection makes the bot wait on it. Removing a check from branch protection lets the bot merge despite its failure. There is no second policy surface to keep in sync with the first.
- **Naturally handles the "unrelated failure" case from #519.** `test-unit / unit-tests (gpg-bridge)` isn't in the ruleset's required contexts (only `Required checks passed` is). The bot ignores its FAILURE and merges. The orchestrator's "ENVIRONMENTAL" classification finally agrees with the bot's evaluation.
- **Composes with #525 / #527.** The bot still re-fires on `workflow_run.completed` for the `CI` parent (#525, just landed). Once #527 lands and the diff-dependent jobs skip on label-only events, the bot sees `skipped` for them — already classified as success above, no further changes. Both fixes preserve the "branch protection is the source of truth" property.
- **Aggregator correctness becomes load-bearing.** If the user requires `Required checks passed` as the only context, the aggregator's `needs:` list MUST include every check that should gate (which is what #518 was fixing). A misconfigured aggregator would let things through. This is correctly the user's responsibility — branch protection points at one context, that context's correctness is the user's contract with the bot.
- **Permissions.** The rulesets / branch-protection APIs accept the App token's existing `administration: read` (or `metadata: read` on the legacy endpoint). The bot's GitHub App needs to be installed with the appropriate scope; today's bot already has read access to rollup data, so the additional API may need a permissions-block update on the App's installation. Verify before merging the change.

### Alt A — static allowlist of ignorable check patterns in `merge-bot.yml`

Add a `ignore_failures:` config block listing check patterns the bot ignores:

```yaml
ignore_failures:
  - test-unit / unit-tests (*)
  - test-integration / integration-tests (*)
```

Closest in shape to the current code — the diff is a filter step at the start of the failing-checks aggregation. Downside: introduces a *second* policy surface alongside branch protection, requiring sync. The recommended design eliminates this surface entirely; this alternative makes the problem worse.

### Alt B — status quo + better orchestrator handling

Leave the bot strict; have the orchestrator (workflow:implement and similar) gate `automerge` ONLY when the FULL rollup is green or the failing checks are explicitly classified ENVIRONMENTAL by the user. The user-prompt cost: every PR with an unrelated FAILURE needs human eyes-on classification. This is what we're doing today after the 2026-05-02 lessons.

This is the no-change-to-bot path. It puts policy enforcement on the orchestrator side, which is the wrong place for it — orchestrators come and go (workflow:implement, Dependabot, manual `gh pr merge`); branch protection is durable.

## Implementation sequence

1. Land Step A in isolation — add the `Read required-check contexts` step but don't change the existing `Inspect required checks` step. Log its output for observability. Verify the same set of contexts appears as the active ruleset listed in `gh api repos/.../rulesets/<id>`.
2. Land Step B by replacing the existing evaluation. The downstream refuse / wait / proceed steps should not need to change.
3. Drop the static failure-classification rule and its commentary in `merge-bot.yml`.
4. Verify the App's installation has `administration: read` (or appropriate scope to read rulesets).
5. Run a few test PRs through the bot — including one with an intentional non-required failure (e.g. cancel a check) to confirm the bot proceeds.

## References

- Concrete example: PR #521 (issue #519) on 2026-05-02 — bot refused on `test-unit / unit-tests (gpg-bridge)` failure that #520 was actively fixing in parallel. User broke deadlock with `gh pr merge --squash --admin`.
- Bot logic: `.github/workflows/merge-bot.yml` lines 391–460 (`Inspect required checks` + `Refuse to merge — required check failed` steps).
- Branch protection ruleset: `gh api repos/aidanns/agent-auth/rulesets/15002184` — only `Required checks passed` is required.
- Workaround memo: `~/.claude/projects/.../memory/project_merge_bot_evaluates_full_rollup.md`.
- Related: #525 (now closed — workflow_run.workflows config out of date; same merge-bot, different bug). #518 (now closed — aggregator's `needs:` list missing test-* shards; the recommended design depends on the aggregator being correct, since the user has chosen to require `Required checks passed` as the gating context). #527 (ci.yml's pull_request trigger missing `labeled`/`unlabeled` types — interacts cleanly with the recommended design via skipped-as-success).

aidanns:
==COMMIT_MSG==
The merge-bot used to refuse on ANY check failure in the rollup,
which is stricter than what branch protection actually requires —
unrelated / flaky / parallel-PR-regression failures blocked the
automated merge path even when `gh pr merge --squash` would have
been allowed by branch protection. The bot now reads the active
branch-protection ruleset(s) on the PR's base branch (with a
fallback to the legacy `branches/{branch}/protection` API) and
gates only on the contexts those rules declare required. Branch
protection becomes the single authoritative source of "what must
be SUCCESS for a merge to land."

The change lands in three commits, in the order the issue's
"Implementation sequence" specifies:

  1. Add a new `Read required-check contexts from branch
     protection` step that derives the required-context set from
     repo config. Sets a `source` output (`rulesets` /
     `branch_protection` / `empty`).
  2. Replace the `Inspect required checks` step's hard-coded
     classification with a per-context loop that consults
     `conclusion` (Actions check-runs) or `state` (legacy commit
     statuses) on the latest run by `startedAt`. The downstream
     `Refuse to merge` and `Wait for pending checks` steps consume
     the same step-output shape and don't change.
  3. Refresh comments and docs (merge-bot.yml header, ADR 0038
     amendment, merge-bot-setup.md) to describe the new policy.
     Add `Administration: read` to the App's required permissions.

Fail-closed when both APIs return empty: an unreachable branch-
protection config refuses the merge with a Claude comment naming
the App-permission hypothesis, rather than silently merging with
no required-check verification (the wrong fail mode).

Override is intentionally NOT a bot-controllable surface; manual
`gh pr merge --squash --admin` remains the only escape hatch.

Verified locally against PR #529:
  - Step A produces `Required checks passed` from the active
    ruleset (id 15002184).
  - Step B classifies it as `success` and emits empty
    `failed`/`pending`.

**Maintainer pre-merge action item:** the App installation
currently lacks `Administration: read`. Update the App's
permissions per `docs/release/merge-bot-setup.md` Step 1 and
re-authorize the install before merging — without this grant,
the rulesets API returns 403 / 404, the legacy fallback also
fails, and the bot will fail-closed (refuse to merge) on every
PR. Documented as a failure mode in the setup guide so future
operators recognise the symptom.

Closes #528
==COMMIT_MSG==

### Test plan

- [ ] CI passes (or any failure is unrelated to this change).
- [ ] Confirm `Read required-check contexts from branch protection` step's log output names the rulesets source and lists `Required checks passed` as the only context.
- [ ] Confirm the `Inspect required checks` step iterates exactly that one context and classifies it as success on a green PR.
- [ ] Maintainer: grant the merge-bot App `Administration: read` and re-authorize the install before applying `automerge`.

### Verified locally

```text
$ gh api repos/aidanns/agent-auth/rulesets --paginate --jq '.[] | select(.target=="branch" and .enforcement=="active") | .id'
15002184

$ gh api repos/aidanns/agent-auth/rulesets/15002184 --jq '.. | objects | select(.type? == "required_status_checks") | .parameters.required_status_checks[]?.context'
Required checks passed
```

End-to-end simulation against PR #529 (no required-context failures, no pending):
```text
Step A: source=rulesets, contexts:
  Required checks passed
  PASS: Required checks passed=success
Step B outputs: failed=[], pending=[]
```

### DIFF
--- a/.github/workflows/merge-bot.yml
+++ b/.github/workflows/merge-bot.yml
@@ -79,12 +79,26 @@ name: Merge Bot
 # other repos. Issue closure failures are warning-only; the merge
 # itself has already landed by that point.
 #
+# Required-check policy: the bot derives its gating set from the
+# active branch-protection ruleset(s) on the PR's base branch (with
+# a fallback to the legacy `branches/{branch}/protection` API for
+# repos that pre-date rulesets). Branch protection is the single
+# authoritative source — adding / removing required checks is a
+# branch-protection edit, not a bot-workflow edit. A check that is
+# failing in the rollup but is NOT in the required-contexts set
+# does not block the merge. Override is intentionally NOT a bot-
+# controllable surface; manual `gh pr merge --squash --admin` is
+# the only escape hatch. See issue #528.
+#
 # Failure modes (ALL surface as a `Claude: Cannot merge — <reason>`
 # comment so the PR author sees the block):
-#   - Required check is FAILURE / TIMED_OUT / CANCELLED.
+#   - A required-context check is FAILURE / TIMED_OUT / CANCELLED.
 #   - `==COMMIT_MSG==` extraction failed (missing block, malformed).
 #   - Block has no `Signed-off-by:` trailer.
 #   - PR is not mergeable (conflict, missing reviews, etc.).
+#   - Branch protection is unreachable (rulesets + legacy APIs both
+#     returned empty) — fail-closed, since merging with no required-
+#     check verification is the wrong fail mode.
 #
 # The label stays sticky on lint/check failure so the
 # workflow_run.completed retrigger picks up the recovered run
--- a/.github/workflows/merge-bot.yml
+++ b/.github/workflows/merge-bot.yml
@@ -496,7 +496,15 @@ jobs:
 
           echo "merge-bot: required-check contexts source = ${source}"
           echo "merge-bot: required-check contexts ="
-          printf '  %s\n' ${contexts:-(none)}
+          if [[ -z "${contexts}" ]]; then
+            echo "  (none)"
+          else
+            # `${contexts}` is a newline-separated list; indent each
+            # line for log readability. Avoids word-splitting on
+            # whitespace because context names can contain spaces
+            # (e.g. "Required checks passed").
+            printf '%s\n' "${contexts}" | awk 'NF { print "  " $0 }'
+          fi
 
           {
             echo "source=${source}"
--- a/.github/workflows/merge-bot.yml
+++ b/.github/workflows/merge-bot.yml
@@ -376,6 +376,121 @@ jobs:
           echo "merge-bot: PR #${PR_NUMBER} state=${PR_STATE} label=${HAS_AUTOMERGE_LABEL}; nothing to do."
           exit 0
 
+      # Read the required-check contexts from the active branch-
+      # protection ruleset(s) for the PR's base branch. Branch
+      # protection is the single authoritative source of "what must
+      # be SUCCESS for a merge to land" — the bot derives its gating
+      # set from this rather than encoding a static "any failure
+      # blocks" rule, so adding / removing required checks is a
+      # branch-protection edit rather than a bot-workflow edit. See
+      # issue #528.
+      #
+      # Two API surfaces, queried in priority order:
+      #   1. The modern rulesets API — `GET /repos/.../rulesets`
+      #      lists rulesets, each detail call returns the
+      #      `required_status_checks` rule's contexts. We filter to
+      #      `target == "branch"` and `enforcement == "active"` so
+      #      disabled / evaluation-mode rulesets and tag rulesets
+      #      don't pollute the contexts list.
+      #   2. The legacy `GET /repos/.../branches/{branch}/protection`
+      #      endpoint — fallback for repos that pre-date rulesets.
+      #
+      # If the rulesets API returns rulesets but extraction yields
+      # zero contexts, we still try the legacy fallback before giving
+      # up — a ruleset with `required_status_checks: false` should
+      # not silently zero out the gating set.
+      #
+      # The `source` output records which path was used so the next
+      # step can fail-closed when neither path produced contexts —
+      # an empty contexts list under the new design would otherwise
+      # let the bot merge with NO required-check verification at all.
+      # That is the wrong fail mode (fail-open); branch protection
+      # being unreachable should refuse the merge with a clear
+      # comment so the maintainer investigates the App permission /
+      # API outage.
+      #
+      # Permissions: the rulesets endpoint requires the calling App
+      # installation to have `Administration: read`; the legacy
+      # endpoint requires the same scope. See
+      # docs/release/merge-bot-setup.md Step 1.
+      - name: Read required-check contexts from branch protection
+        id: required
+        if: steps.secrets-check.outputs.configured == 'true' && steps.pr.outputs.pr != '' && steps.meta.outputs.state == 'OPEN' && steps.meta.outputs.has_automerge_label == 'true'
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          # `gh pr view --json baseRefName` returns the PR's target
+          # branch (typically `main`). Refetch here rather than wiring
+          # it through the earlier `meta` step so the dependency on
+          # the base branch is co-located with the step that uses it.
+          PR_NUMBER: ${{ steps.pr.outputs.pr }}
+        run: |
+          set -euo pipefail
+          base_ref="$(gh pr view "${PR_NUMBER}" \
+            --repo "${GITHUB_REPOSITORY}" \
+            --json baseRefName \
+            --jq '.baseRefName')"
+          echo "merge-bot: PR base branch = ${base_ref}"
+
+          # Step 1 — modern rulesets API. List active branch-targeted
+          # rulesets, then per-ruleset detail call extracts the
+          # required-status-checks contexts. The `--jq` filter on the
+          # detail call walks the full rule tree (`.. | objects`)
+          # because the schema may evolve; selecting on
+          # `type == "required_status_checks"` is shape-stable.
+          contexts=""
+          ruleset_ids="$(gh api \
+            "repos/${GITHUB_REPOSITORY}/rulesets" \
+            --paginate \
+            --jq '.[] | select(.target == "branch" and .enforcement == "active") | .id' \
+            2>/dev/null || true)"
+          if [[ -n "${ruleset_ids}" ]]; then
+            while IFS= read -r id; do
+              [[ -z "${id}" ]] && continue
+              detail="$(gh api "repos/${GITHUB_REPOSITORY}/rulesets/${id}" 2>/dev/null || true)"
+              if [[ -z "${detail}" ]]; then
+                echo "::warning::Could not fetch ruleset #${id}; skipping."
+                continue
+              fi
+              ruleset_ctx="$(jq -r '
+                .. | objects
+                | select(.type? == "required_status_checks")
+                | .parameters.required_status_checks[]?.context
+              ' <<<"${detail}" || true)"
+              if [[ -n "${ruleset_ctx}" ]]; then
+                contexts+="${ruleset_ctx}"$'\n'
+              fi
+            done <<<"${ruleset_ids}"
+          fi
+          # Dedupe and strip blank lines.
+          contexts="$(printf '%s' "${contexts}" | awk 'NF' | sort -u)"
+
+          source="rulesets"
+          if [[ -z "${contexts}" ]]; then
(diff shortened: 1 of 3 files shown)


aidanns:
Claude Reviewer: LGTM. Step A correctly filters rulesets to target=="branch" AND enforcement=="active", paginates, walks `.. | objects | select(.type? == "required_status_checks")` per ruleset detail, falls back to `branches/{branch}/protection` only on empty, and emits the contexts via `<<MERGE_BOT_EOF` heredoc. Step B iterates only those contexts, dedupes per-context by latest `startedAt`, classifies via `conclusion // .state` into `success|neutral|skipped` (satisfied), pending bucket (`""|null|missing|in_progress|queued|waiting|pending|requested`), or fail bucket (everything else), and preserves both the merge-bot self-filter and the dedupe-by-latest fix from #347. Static "any FAILURE in the rollup blocks merge" loop and its commentary are fully removed; `failed`/`pending` outputs keep the newline-separated shape so `Refuse to merge` and `Wait for pending checks` are unchanged. Fail-closed when both APIs return empty refuses with a Claude comment naming the App-permission hypothesis (the right fail mode). No `merge-anyway` label or workflow input. Docs (ADR 0038 amendment, merge-bot-setup.md Step 1 + failure-modes section) are updated with `Administration: read` and the new failure mode. PR title 53 chars under `chore(ci):`, `==COMMIT_MSG==` block has `Closes #528` immediately followed by `Signed-off-by:` with no blank line, all five commits carry DCO sign-off, `no changelog` label is applied, empty re-trigger commit is present.

PR #526, opened 2026-05-02
chore(ci): aggregator gates on test-* shards

aidanns:
==COMMIT_MSG==
The required-checks-passed job in ci.yml aggregated only the
non-test workflow_call children (plan, check-*, build); the four
test-* siblings landed in #491 (test-unit, test-integration,
test-smoke, test-system) were never added to its needs: list. Per
the #441 contract ("fails on any failed/skipped/timed-out
result"), missing children silently bypass the gate, so a red
test shard would still leave Required checks passed green and
unblock merge under the post-#5.2 branch-protection ruleset.

Extend the list to cover all four shards so the aggregator
faithfully reflects every PR-time CI signal.

Closes #518
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

==NO_CHANGELOG==

### Test plan

- [ ] CI on this PR shows the four `test-*` children listed under the `Required checks passed` aggregator's `needs:` (visible by inspecting the job graph).
- [ ] `Required checks passed` succeeds only when every test shard is `success` (verified implicitly by this PR's own green run).

Issue #441, opened 2026-04-29
ci: create ci.yml orchestrator with plan job and required-checks-passed aggregator


aidanns:
Part of #440.

## Scope
Create the new `ci.yml` parent orchestrator that will eventually invoke every PR-time `check-*` and `test-*` child workflow via `workflow_call`. Initially calls zero children — this issue establishes the shell only.

## Plan
- Create `.github/workflows/ci.yml` with:
  - `on: pull_request, push (main), merge_group, workflow_dispatch`
  - `plan` job that detects labels and changed-files matrix and emits gating booleans as outputs
  - `required-checks-passed` aggregator job that `needs:` every child workflow (none yet) and fails on any failed/skipped/timed-out result
  - Concurrency group keyed on `workflow + ref + PR or SHA`
- No existing workflows are modified or removed
- Mark `required-checks-passed` as a status check that runs on every PR (not yet required in branch protection — see #5.2)

## Dependencies
Independent — no other migration issues need to land first.

aidanns:
Closed by merge of PR #473 (squash commit on main). Auto-close didn't fire — closing manually per merge-bot quirk.

PR #533, opened 2026-05-02
chore(ci): verified partial re-run on label events

### LINKED ISSUE #527: ci.yml: pull_request trigger missing 'labeled' / 'unlabeled' types — label-only state changes don't re-fire CI
## Observed

`.github/workflows/ci.yml` declares its `pull_request` trigger as:

```yaml
on:
  pull_request:
    branches: [main]
    types: [opened, synchronize, reopened, ready_for_review]
```

`labeled` and `unlabeled` are NOT in the list. As a result, applying or removing a label after `gh pr create` (the most common path) does NOT re-fire ci.yml. Specifically, `check-changelog/exists` and `check-changelog/format` (called as `workflow_call` children of ci.yml) read `github.event.pull_request.labels` from the event payload that fired ci.yml — when ci.yml fired on `opened`, that payload had `labels: []`, even if the `no changelog` label is on the PR by the time the human / orchestrator looks at the rollup.

## Impact

Reproduced four times in the 2026-05-02 `/workflow:implement` runs (PRs #521, #524, #526, #529). On every PR:

1. Agent runs `gh pr create`.
2. Agent runs `gh pr edit <pr#> --add-label "no changelog"`.
3. ci.yml fired on step 1 with `labels=[]`. `check-changelog/exists` evaluates `[[ ",${PR_LABELS}," == *",no changelog,"* ]]` → false → emits `::error::PR #<N>: no changelog entry added under changelog/@unreleased/pr-<N>-<slug>.yml`.
4. The label is now applied, but ci.yml has no trigger to re-evaluate. `Required checks passed` (the aggregator) fails.

The current workaround is to push an empty commit immediately after applying the label (`git commit --allow-empty -s -m 'chore: re-trigger ci.yml after no-changelog label applied'` + `git push`) — that fires `synchronize` with the new label list visible. This is recorded in `~/.claude/projects/.../memory/feedback_ci_yml_no_labeled_trigger.md` and is now embedded in every dispatched-agent prompt by the orchestrator, but it costs an extra CI cycle on every PR (~5–10 min).

The legacy `changelog-lint.yml` workflow does include `labeled, unlabeled` in its types — confirming this gap is most likely a copy-paste oversight in the ci.yml split (#491 / #509 / #516).

### Option 1 — add `labeled, unlabeled` to ci.yml's `types:`

```yaml
on:
  pull_request:
    branches: [main]
    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
```

Smallest change. Every label change now triggers a full CI run (~5–10 min). For agent-auth's low label volume (`automerge`, `no changelog`, `blocked`, `in-progress`, `paused`, `scheduled` — typically 2-3 toggles per PR during an orchestrator's lifecycle) the wasted CI minutes are small. Cheapest by surface area; doesn't address the underlying "running everything for a label-only signal" wastefulness.

### Option 2 — relax the aggregator's "skipped = failure" rule on label-only events

Add `labeled, unlabeled` to ci.yml's `types:` as in Option 1, but also gate heavy children with `if: github.event.action != 'labeled' && github.event.action != 'unlabeled'` so they skip on label-only re-runs. The aggregator (`required-checks-passed`) is then taught to ignore `skipped` results on label-only events for the gated children, narrowing #441's contract specifically for this case.

Lower wasted runtime than Option 1, but the relaxation is unconditional — a label change against a SHA whose CI never finished would mark the heavy children skipped without any prior-success verification. The aggregator carve-out is a deliberate weakening that documents itself but doesn't *prove* the skip is safe. See Option 4 for the verified version.

### Option 3 — move `check-changelog` out of ci.yml into a label-aware sibling workflow

Drop `check-changelog` from ci.yml's `needs:` chain. Create a new top-level workflow `check-changelog-pr.yml` that triggers on `[opened, synchronize, reopened, ready_for_review, labeled, unlabeled]` and calls the existing `check-changelog.yml` workflow_call. Branch protection then requires `check-changelog / exists` and `check-changelog / format` directly, not through ci.yml's aggregator.

Surgical to check-changelog, but adds a new workflow file, two new branch-protection contexts, and the aggregator no longer hands you a single-context gating story. Deferred features (other label-aware checks added later) would each need this same split.

### Option 4 (recommended) — verified partial re-run, gated on prior-success on the same head SHA

Add `labeled, unlabeled` to ci.yml's `types:` (as Option 1/2), but *verify* that the diff-dependent jobs already passed on the current head SHA before allowing them to skip. The verification lives in the `plan` job and uses the GitHub Checks API. The aggregator's "skipped = failure" rule (#441) is relaxed only for the verified-prior-success case.

#### Why "diff-dependent"

A `labeled` event's `github.event.pull_request.head.sha` is the *same* SHA the previous `synchronize` ran against, so any check-run with `conclusion: success` for that SHA is still authoritative — the diff hasn't changed, only PR metadata (labels) has. The list we want to skip on label-only re-runs is therefore the set of jobs whose result depends *only on the code diff* — not on PR metadata. Naming this set `diff_dependent_jobs` (rather than the imprecise "heavy jobs") makes the safety property visible at the call site:

> "If the diff is unchanged AND every diff-dependent job has SUCCESS for this head SHA, skip them."

`check-changelog`, `check-pull-request` and similar are *not* diff-dependent — their evaluation depends on labels, title, body, etc. — so they re-run on every event regardless.

#### Sketch — verification step in `plan`

Add an output to the existing `plan` job (which already emits `label_no_changelog`, `label_automerge`, and changed-files booleans):

```yaml
plan:
  outputs:
    label_no_changelog: ...
    label_automerge: ...
    diff_dependent_jobs_already_passed: ${{ steps.prior.outputs.diff_dependent_jobs_already_passed }}
  steps:
    # ... existing labels + filter steps ...

    - name: Probe prior run state on this head SHA
      id: prior
      env:
        GH_TOKEN: ${{ github.token }}
        HEAD_SHA: ${{ github.event.pull_request.head.sha }}
        EVENT_ACTION: ${{ github.event.action }}
      run: |
        set -euo pipefail
        # Only relevant on label-change re-runs; on opened / synchronize
        # the diff itself changed, so prior check-runs are stale by definition.
        if [[ "$EVENT_ACTION" != "labeled" && "$EVENT_ACTION" != "unlabeled" ]]; then
          echo "diff_dependent_jobs_already_passed=false" >> "$GITHUB_OUTPUT"
          exit 0
        fi

        # Diff-dependent children — outcome determined entirely by the code in
        # the head SHA. If a prior run on this SHA produced SUCCESS for each,
        # the same SUCCESS still holds on a label-only re-run.
        # NOTE: keep this list in sync with `required-checks-passed.needs:`
        # MINUS the metadata-dependent set (check-changelog, check-pull-request,
        # etc.). See "Keeping the list in sync" below for an automation idea.
        diff_dependent=$(cat <<'EOF'
        Check Fmt / Required checks passed
        check-lint / Required checks passed
        check-security / Required checks passed
        check-docs / check-docs
        check-publish / check-publish
        check-release / check-release
        check-standards / verify-standards
        build / build
        test-unit / unit-tests (agent-auth)
        test-unit / unit-tests (agent-auth-common)
        test-unit / unit-tests (gpg-bridge)
        test-unit / unit-tests (gpg-cli)
        test-unit / unit-tests (things-bridge)
        test-unit / unit-tests (things-cli)
        test-unit / unit-tests (things-client-cli-applescript)
        test-integration / integration-tests (agent-auth)
        test-integration / integration-tests (gpg-bridge)
        test-integration / integration-tests (things-bridge)
        test-integration / integration-tests (things-cli)
        test-integration / integration-tests (things-client-applescript)
        test-smoke / install-from-wheels (gpg-bridge, gpg-bridge)
        test-smoke / install-from-wheels (things-cli, things-cli)
        test-system / macos-applescript
        EOF
        )

        rollup=$(gh api "repos/$GITHUB_REPOSITORY/commits/$HEAD_SHA/check-runs" \
                 --paginate --jq '.check_runs')

        all=true
        while IFS= read -r name; do
          [[ -z "$name" ]] && continue
          latest=$(jq -r --arg n "$name" \
            '[.[] | select(.name == $n)] | sort_by(.started_at) | .[-1].conclusion // "missing"' \
            <<<"$rollup")
          # `success` / `neutral` / `skipped` from a prior label-aware run all pass.
          # `failure` / `cancelled` / `timed_out` / `missing` invalidate the skip.
          case "$latest" in
            success|neutral|skipped) ;;
            *)
              echo "::notice::diff-dependent job '$name' conclusion=$latest on head $HEAD_SHA; will re-run."
              all=false
              break
              ;;
          esac
        done <<<"$diff_dependent"

        echo "diff_dependent_jobs_already_passed=$all" >> "$GITHUB_OUTPUT"
```

Each diff-dependent child gates on the flag:

```yaml
test-unit:
  needs: [plan]
  if: needs.plan.outputs.diff_dependent_jobs_already_passed != 'true'
  uses: ./.github/workflows/test-unit.yml
```

The metadata-dependent children (`check-changelog`, `check-pull-request`) keep their unconditional `needs: [plan]` and re-run on every event.

The aggregator gets a matching carve-out: when `needs.plan.outputs.diff_dependent_jobs_already_passed == 'true'`, treat `result: skipped` on the diff-dependent children as success rather than failure. The carve-out is auditable — it only fires when *both* the action is `labeled`/`unlabeled` AND a prior SUCCESS for every diff-dependent child was observed at this SHA.

#### Properties of the verified design

- **Safe default.** If any diff-dependent child's prior conclusion is `failure` / `cancelled` / `timed_out` / `missing`, the flag is `false` and the children run as today. No regression vs. Option 1.
- **Skip is verified, not assumed.** The aggregator's relaxation only applies when `plan` saw the prior SUCCESS. A label change against a SHA whose CI never finished re-runs everything.
- **Concurrency-safe.** If a label change fires while the previous synchronize's ci.yml is still running, the new run cancels the old via `cancel-in-progress: true`. The Checks API at that moment shows the old run's children as `cancelled` → flag is `false` → new run executes everything. Wasteful but safe; the label-change re-run in this window can't claim the skip.
- **Forced-rerun-aware.** If the user clicks "Re-run failed jobs" in the UI on a synchronize-triggered ci.yml, the prior diff-dependent children become SUCCESS; a subsequent label change skips correctly.
- **Permissions.** The Checks API endpoint accepts `contents: read`, which the `plan` job already has — no permission additions needed.
- **Plays well with merge-bot's stricter gate (#528).** The bot reads `gh pr view --json statusCheckRollup`. On a label-change re-run with skip enabled, each diff-dependent child's check-run ends up `conclusion: skipped`. The bot's `Inspect required checks` step already classifies `skipped` as not-a-failure (per the existing logic — it's only `FAILURE / TIMED_OUT / CANCELLED / STARTUP_FAILURE / ACTION_REQUIRED` that block), so the bot also tolerates the skip without further changes.

#### Keeping the list in sync

The `diff_dependent` list in `plan` has to mirror ci.yml's actual children — drift either causes silent false skips (a new child added without being in the list still gets gated by the flag → skips when the flag is true → the aggregator carve-out doesn't cover it → aggregator fails) or false re-runs (a child added to the list but no longer in `needs:` does nothing).

One option: derive the list from ci.yml's own `required-checks-passed.needs:` minus a hard-coded metadata-dependent set (`check-changelog`, `check-pull-request`, anything else that reads PR metadata). That way adding a new child to `needs:` auto-extends the diff-dependent list. The derivation could live in a helper script (e.g. `scripts/ci/diff-dependent-jobs.sh`) called from `plan`'s shell step, with the metadata-dependent allowlist a small constant alongside.

Alternative: enumerate explicitly as in the sketch above and lean on the test-suite to catch drift. Less elegant but easier to audit.

## Recommendation

Option 4 is the architecturally right path. The reason to prefer it over Option 1: the same skipped-on-label-only-event gap will bite future label-aware checks — anything new that wants to gate on `automerge` or `no changelog` will run into the same trigger-types omission. Fix the aggregator once (with verified-prior-success as the safety property) and the trigger-types list can grow safely without paying full CI cost on every label flip.

The implementation sequence is:

1. Land the `plan`-job verification step + the `diff_dependent_jobs_already_passed` output. Verify it produces sensible values on a few label-flip events (no change to downstream gating yet — it's an observable-only first pass).
2. Land the `if:` gates on diff-dependent children + the aggregator carve-out in one PR. This is the breaking change that needs eyes-on review.
3. Land the `labeled, unlabeled` types addition. Now label flips re-fire ci.yml; the prior steps ensure the partial-re-run path is already wired and verified.

## References

- Originating session: 2026-05-02 `/workflow:implement` runs (PRs #521, #524, #526, #529), each of which paid the empty-commit-retrigger cost.
- Workaround memo: `~/.claude/projects/.../memory/feedback_ci_yml_no_labeled_trigger.md`.
- Legacy precedent: `.github/workflows/changelog-lint.yml` already lists `labeled, unlabeled` in its types.
- Related: #525 (now closed — merge-bot's `workflow_run` config; same family of "trigger list out of sync with current behaviour" bugs). #528 (merge-bot evaluates the full rollup — interacts with Option 4's skip semantics; this issue's resolution should compose cleanly with that one's).
- Aggregator semantics: #441 (the "skipped = failure" contract that Option 2/4 narrow).

aidanns:
==COMMIT_MSG==
Applying or removing a label after `gh pr create` did not re-fire
ci.yml — `pull_request: types:` omitted `labeled, unlabeled` — so
`check-changelog/exists` evaluated against the original `labels=[]`
event payload, emitted a stale-failure status, and forced an
empty-commit retrigger to clear it. Reproduced four times in the
2026-05-02 /workflow:implement runs (PRs #521, #524, #526, #529);
each PR paid the full ~5-10 min CI cost on the workaround.

Implements Option 4 — verified partial re-run, gated on prior-
success on the same head SHA — in three coupled phases landed
together so the trigger widening never ships without its safety
property:

1. `plan` job's `Probe prior run state on this head SHA` step
   queries the Checks API for prior conclusions on the head SHA
   on `labeled`/`unlabeled` events; emits
   `diff_dependent_jobs_already_passed=true` only when every
   diff-dependent child has a non-failing conclusion on the same
   SHA. The diff-dependent leaf list comes from a new helper
   `scripts/ci/diff-dependent-jobs.sh` that derives from
   `required-checks-passed.needs:` minus a hard-coded
   metadata-dependent allowlist (`check-changelog`,
   `check-pull-request`), so adding a child to `needs:` extends
   the probe automatically.
2. Every diff-dependent ci.yml `<job>:` block carries
   `if: needs.plan.outputs.diff_dependent_jobs_already_passed !=
   'true'` so it skips on verified-prior-success label re-runs.
   `check-changelog` and `check-pull-request` keep their
   unconditional `needs: [plan]` and re-run on every event.
3. `required-checks-passed` aggregator gets a tight carve-out:
   when the flag is `true` AND the entry is in the diff-dependent
   set, `skipped` maps to SUCCESS. Otherwise the strict #441
   skipped-equals-failure contract still applies — a label change
   against a SHA whose CI never finished re-runs everything by
   design.
4. `pull_request: types:` adds `labeled, unlabeled`.

Tests in `tests/test_ci_diff_dependent_probe.py` pin the four
properties as static-yaml regressions: trigger types include
`labeled, unlabeled`, every diff-dependent child carries the gate,
metadata-dependent children carry no gate, the aggregator carve-out
gates on both flag and per-key set membership, and the helper
script's emitted leaves match what GitHub's Checks API publishes.

Closes #527
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

The diff is ordered to match the issue's "Implementation sequence":
plan-job probe + helper + new output first; then `if:` gates on each
diff-dependent child; then aggregator carve-out; then the
`labeled, unlabeled` types addition. Reading top-to-bottom in
ci.yml mirrors that sequence as well — the probe step and output
appear before any `if:` gate that consumes them, the gate appears
on each child before the aggregator that handles its skipped result,
and the trigger types live in `on:` at the top.

Drift checks the test file pins:
- helper output equals expected leaf set (one row per matrix variant);
- helper is byte-deterministic across multiple invocations;
- helper never emits a leaf rooted under `check-changelog` /
  `check-pull-request`;
- helper's covered ci.yml job IDs equal `required-checks-passed.needs:`
  minus `plan` minus the metadata-dependent allowlist (drift here
  catches a new child added to `needs:` without extending the helper);
- aggregator's `DIFF_DEPENDENT_JOB_IDS` env value equals the same
  set (drift here catches a gate added without a carve-out hook
  or vice versa);
- aggregator's jq filter references both the flag and per-key set
  membership (drift here catches a future widening of the relaxation).

This PR's own base branch lacks the `labeled` / `unlabeled` trigger
(meta-circular case from the issue body) so the customary
`gh pr edit --add-label "no changelog"` will not re-fire CI on its
own — the implementor will push an empty commit after labelling, as
recorded in CLAUDE.md / auto-memory. Once this PR merges to main the
workaround is no longer needed.

### Test plan

- [ ] `task lint` passes (lefthook ran on commit; treefmt + ruff
      auto-formatted the test file before the second commit attempt)
- [ ] `task test` passes (covers the 11 new tests in
      `tests/test_ci_diff_dependent_probe.py`)
- [ ] `bash scripts/ci/diff-dependent-jobs.sh` prints the 23
      expected leaf check-run names (one per ci.yml diff-dependent
      job + matrix expansion)
- [ ] `bash scripts/verify-standards.sh` is green
- [ ] CI on this PR runs the diff-dependent children at least once
      (first run is on `opened` → flag is `false` → children execute)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,14 +18,26 @@ name: CI
 
 on:
   pull_request:
-    # Default activity types — `opened`, `synchronize`, `reopened`,
-    # `ready_for_review`. The aggregator surfaces as a status check
-    # on every PR, including draft → ready transitions. Restricted
-    # to PRs targeting `main` to match the convention every other
-    # PR-time workflow in this repo (`test.yml`, `check.yml`,
+    # Activity types — `opened`, `synchronize`, `reopened`,
+    # `ready_for_review` cover the diff-changing transitions; the
+    # aggregator surfaces as a status check on every PR, including
+    # draft → ready transitions. `labeled` and `unlabeled` are
+    # included so PR-time workflows that gate on labels
+    # (`check-changelog/exists` reading `no changelog`,
+    # dependabot-adapter reading `automerge`) re-fire when the label
+    # set changes — without them, applying `no changelog` after
+    # `gh pr create` left ci.yml stuck on the original `labels=[]`
+    # payload and `Required checks passed` reported a stale failure
+    # (issue #527). The diff-dependent children are skipped on
+    # label-only re-runs via the verified-prior-success path
+    # below, so the wasted-runtime cost of widening the trigger
+    # surface is bounded to the metadata-dependent children
+    # (`check-changelog`, `check-pull-request`). Restricted to PRs
+    # targeting `main` to match the convention every other PR-time
+    # workflow in this repo (`test.yml`, `check.yml`,
     # `typecheck.yml`, ...) uses.
     branches: [main]
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
   push:
     branches: [main]
   merge_group:
@@ -67,6 +79,16 @@ jobs:
       docs_changed: ${{ steps.filter.outputs.docs }}
       design_changed: ${{ steps.filter.outputs.design }}
       workflows_changed: ${{ steps.filter.outputs.workflows }}
+      # Verified-prior-success flag (issue #527 / Option 4). `true`
+      # only on `labeled` / `unlabeled` events whose head SHA already
+      # has a non-failing conclusion for every diff-dependent child
+      # check-run; downstream `<job>:` `if:` gates and the aggregator
+      # carve-out both consume this. `false` on every other event
+      # (the diff itself changed → prior check-runs are stale by
+      # definition) and whenever the prior conclusion of any
+      # diff-dependent child is `failure` / `cancelled` /
+      # `timed_out` / `missing` (the safe-default branch).
+      diff_dependent_jobs_already_passed: ${{ steps.prior.outputs.diff_dependent_jobs_already_passed }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
@@ -128,6 +150,78 @@ jobs:
               - '.github/workflows/**'
               - '.github/actions/**'
 
+      - name: Probe prior run state on this head SHA
+        id: prior
+        # Verified partial-rerun gate for label-only events
+        # (issue #527 / Option 4). On `labeled` / `unlabeled` the
+        # head SHA is identical to the previous `synchronize`'s
+        # head SHA — so any check-run on that SHA whose conclusion
+        # is non-failing is still authoritative. We probe each
+        # diff-dependent child via the Checks API; if every one
+        # is non-failing on this SHA, downstream `<job>:` blocks
+        # skip via the `if:` gate below and the aggregator's
+        # `skipped = failure` rule (#441) is relaxed for that
+        # specific case. If ANY child's prior conclusion is
+        # `failure` / `cancelled` / `timed_out` / `missing`, the
+        # flag is `false` and downstream children run as today —
+        # safe default, no regression.
+        #
+        # The Checks API endpoint
+        # (`repos/{owner}/{repo}/commits/{sha}/check-runs`) accepts
+        # the workflow-level `contents: read` permission, which the
+        # `plan` job inherits — no additional permissions needed.
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          EVENT_ACTION: ${{ github.event.action }}
+        run: |
+          set -euo pipefail
+          # On `opened` / `synchronize` / `reopened` /
+          # `ready_for_review` the diff itself changed (or this is
+          # the first run on the SHA), so prior check-runs are
+          # stale by definition.
+          if [[ "${EVENT_ACTION}" != "labeled" && "${EVENT_ACTION}" != "unlabeled" ]]; then
+            echo "diff_dependent_jobs_already_passed=false" >> "${GITHUB_OUTPUT}"
+            exit 0
+          fi
+
+          # Diff-dependent check-run names — derived from
+          # `required-checks-passed.needs:` minus the metadata-
+          # dependent allowlist. Driving this from a helper rather
+          # than enumerating inline keeps the list in lockstep with
+          # ci.yml itself: when a new child is added to `needs:`,
+          # the helper auto-extends and the carve-out covers it
+          # without churning this file.
+          diff_dependent=$(scripts/ci/diff-dependent-jobs.sh)
+
+          rollup=$(gh api \
+            "repos/${GITHUB_REPOSITORY}/commits/${HEAD_SHA}/check-runs" \
+            --paginate --jq '.check_runs')
+
+          all=true
+          while IFS= read -r name; do
+            [[ -z "${name}" ]] && continue
+            latest=$(jq -r --arg n "${name}" \
+              '[.[] | select(.name == $n)] | sort_by(.started_at) | .[-1].conclusion // "missing"' \
+              <<<"${rollup}")
+            # `success` / `neutral` / `skipped` from a prior label-
+            # aware run all preserve the SUCCESS contract for the
+            # aggregator. Anything else (`failure`, `cancelled`,
+            # `timed_out`, `missing`, `action_required`,
+            # `startup_failure`, `null`) invalidates the skip and
+            # the children re-run.
+            case "${latest}" in
+              success | neutral | skipped) ;;
+              *)
+                echo "::notice::diff-dependent job '${name}' conclusion=${latest} on head ${HEAD_SHA}; will re-run."
+                all=false
+                break
+                ;;
+            esac
+          done <<<"${diff_dependent}"
+
+          echo "diff_dependent_jobs_already_passed=${all}" >> "${GITHUB_OUTPUT}"
+
   check-fmt:
     # Intermediate orchestrator carrying formatting / lint children
     # (issue #452 lands `spdx-license-headers`; #459 will add treefmt
@@ -136,6 +230,14 @@ jobs:
     # opt into changed-files gating.
     name: Check Fmt
     needs: [plan]
+    # Diff-dependent: outcome is determined by code in the head SHA.
+    # Skip on `labeled`/`unlabeled` re-runs whose head SHA already
+    # passed (issue #527 / Option 4). The aggregator carve-out at
+    # the bottom of this file converts the resulting `skipped`
+    # result back into a SUCCESS contract, but only when the same
+    # flag is true — so the relaxation can't fire unless the probe
+    # in `plan` verified the prior pass.
+    if: needs.plan.outputs.diff_dependent_jobs_already_passed != 'true'
     uses: ./.github/workflows/check-fmt.yml
   check-security:
     # Security-cadence child orchestrator (issue #447). Carries CodeQL
@@ -147,6 +249,8 @@ jobs:
     # reference plan outputs without churning this file.
     name: Check Security
     needs: [plan]
+    # Diff-dependent — see `check-fmt` above for the gate contract.
+    if: needs.plan.outputs.diff_dependent_jobs_already_passed != 'true'
     uses: ./.github/workflows/check-security.yml
     permissions:
       # Granted to the `check-security.yml` chain. workflow_call
@@ -172,6 +276,8 @@ jobs:
     # continues to run alongside this chain until the Phase 5 sweep.
     name: check-lint
     needs: [plan]
+    # Diff-dependent — see `check-fmt` above for the gate contract.
+    if: needs.plan.outputs.diff_dependent_jobs_already_passed != 'true'
     uses: ./.github/workflows/check-lint.yml
     secrets: inherit
   check-docs:
@@ -181,6 +287,8 @@ jobs:
     # `.github/workflows/check-docs.yml` for the TODO marker.
     name: check-docs
     needs: [plan]
+    # Diff-dependent — see `check-fmt` above for the gate contract.
+    if: needs.plan.outputs.diff_dependent_jobs_already_passed != 'true'
     uses: ./.github/workflows/check-docs.yml
 
   test-unit:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -18,14 +18,26 @@ name: CI
...
+    # included so PR-time workflows that gate on labels
+    # (`check-changelog/exists` reading `no changelog`,
+    # dependabot-adapter reading `automerge`) re-fire when the label
+    # set changes — without them, applying `no changelog` after
+    # `gh pr create` left ci.yml stuck on the original `labels=[]`
+    # payload and `Required checks passed` reported a stale failure
+    # (issue #527). The diff-dependent children are skipped on
+    # label-only re-runs via the verified-prior-success path
+    # below, so the wasted-runtime cost of widening the trigger
+    # surface is bounded to the metadata-dependent children
```
aidanns, line 34: Claude Reviewer: Issue #527's "Implementation sequence" lays out three commits — (1) plan-job probe + helper + new output, (2) `if:` gates + aggregator carve-out, (3) `labeled, unlabeled` types addition — and the PR brief explicitly asked for the branch to land them as three separate commits in that order so each phase reviews on its own. This PR squashes all three into the single commit `6e563d3` ("verified partial re-run on label events"). Functionally correct (and the squash-merge collapses it anyway), but the multi-commit shape is what the issue asked for so a reviewer can confirm the safety property in isolation: phase 1 alone (probe + output, no consumers) is observable-only and provably side-effect-free; phase 2 alone (gates + carve-out) is the breaking change that needs eyes-on; phase 3 alone (trigger widening) is the one-liner that would otherwise ship without the safety wiring. Splitting them in a follow-up isn't worth churning the PR; flagging for next time so the commit-ordered review path the issue calls out actually lands.

aidanns: Claude: Squashing into one commit was deliberate at push time — splitting after review (now) requires a force-push to a branch under review, which CLAUDE.md and the dispatch template explicitly disallow. The diff is reviewable as a single unit and the per-phase boundaries are still visible in the YAML structure (probe → gates → trigger-types). Will keep this in mind for future PRs of similar shape — split BEFORE first push.

PR #541, opened 2026-05-02
chore(ci): export PYTHONPATH from setup-toolchain action

### LINKED ISSUE #522: Fix PYTHONPATH propagation for test-unit shards (eliminate test-layout workaround)

aidanns:
==COMMIT_MSG==
The per-package ``test-unit / unit-tests (gpg-bridge)`` shard
introduced in #491 fails 19 of 95 tests because the shard runs
raw ``pytest`` without sourcing ``scripts/_bootstrap_venv.sh``,
which is the only place the workspace exports
``packages/<pkg>/tests`` on ``PYTHONPATH`` so subprocesses
launched by tests (``python -m gpg_backend_fake``,
``python -m things_client_fake``) can resolve the test-only
fakes. Pytest's ``pythonpath`` config only affects the in-process
``sys.path``, not subprocess ``PYTHONPATH``.

Move the export into ``.github/actions/setup-toolchain`` so every
workflow that uses the action (per-package shards, verify-standards,
codeql, ...) inherits the right environment from $GITHUB_ENV
without sourcing the bootstrap script. Mirrors the bootstrap export
verbatim so local + CI behaviour stays in lockstep; the bootstrap
script is unchanged so local developers keep their existing flow.

With the action-side export in place, the #520 / #524 workaround
that relocated the gpg-subprocess + HTTP wiring tests under
``packages/gpg-bridge/tests/integration/`` is no longer needed.
Revert it end-to-end:

- ``git mv`` the 5 relocated tests back to
  ``packages/gpg-bridge/tests/`` (test_gpg_bridge_server,
  test_gpg_bridge_perf_budget,
  test_gpg_subprocess_client[_passphrase] back to ``tests/``;
  test_bridge_network_drop back to ``tests/fault/``).
- Re-fold the gpg_backend_fake-driven wedge case from
  ``tests/integration/test_wedged_gpg_subprocess.py`` back into
  ``tests/fault/test_backend_subprocess_hang.py`` and delete the
  extracted file.
- Restore ``--cov-fail-under=62`` in
  ``packages/gpg-bridge/pyproject.toml`` (was lowered to 55 as
  the workaround).
- Restore the ``test_gpg_bridge_server`` /
  ``test_gpg_subprocess_client`` entries in the workspace
  ``[mypy.overrides] module`` and the ``test_gpg_bridge_server.py``
  entry in ``[mutmut] tests_dir``, both dropped when the files
  moved out of ``tests/``.
- Drop the ``# integration-isolation: in-process-server`` carve-out
  from ``scripts/verify-integration-isolation.sh``; with no
  in-process tests left under ``tests/integration/`` the marker
  no longer has any callers and the simpler grep loop is the
  pre-#524 default.

Verified locally: ``pytest packages/gpg-bridge/tests/
--ignore=packages/gpg-bridge/tests/integration --no-cov`` runs
95 passing tests with the action-style PYTHONPATH set, and 19
failing without it (same shape as the original CI failure mode).

Supersedes #523 -- the codeql-config.yml and ADR 0040 path
references it tracked become valid again once the tests are
relocated back; #523 has been closed.

Closes: #522
==COMMIT_MSG==

==NO_CHANGELOG==

### Test plan

- [ ] CI ``test-unit (gpg-bridge)`` shard passes (was the original failure
      mode in #520; should now go green with PYTHONPATH coming from the
      setup-toolchain action).
- [ ] CI ``test-unit (things-bridge)`` shard passes (sibling case for
      ``things_client_fake``, tracked as #519 -- same root cause, fixed
      by the same export).
- [ ] CI ``test-integration (gpg-bridge)`` shard still passes (no
      regression on the relocated-then-restored tests; they stay green
      whether the unit shard sees them or the integration shard does).
- [ ] CI ``check-standards`` passes (the simplified
      ``verify-integration-isolation.sh`` grep loop still rejects
      stray 127.0.0.1 references in ``tests/integration/`` directories).
- [ ] No regression on the workspace ``unit`` job in the legacy
      ``test.yml`` (gpg-bridge cov gate is back at 62, mutmut sees
      ``test_gpg_bridge_server`` again).


aidanns:
Claude Reviewer: LGTM. Verified: (1) PYTHONPATH export step uses `>> "$GITHUB_ENV"` with the exact `things-bridge/tests:gpg-bridge/tests${PYTHONPATH:+:$PYTHONPATH}` ordering from `_bootstrap_venv.sh`; (2) `test_backend_subprocess_hang.py` post-revert is byte-identical to the pre-#524 version (zero-diff `git show 6823b2c^:... vs pr-541:...`); (3) `verify-integration-isolation.sh`, root `pyproject.toml`, and `packages/gpg-bridge/pyproject.toml` all byte-identical to pre-#524 — no stranded comment fragments around the restored `--cov-fail-under=62`; (4) no remaining `# integration-isolation: in-process-server` markers anywhere in `packages/`, `scripts/`, or `.github/` (the only files left in `tests/integration/` are `conftest.py` and the Docker-backed `test_bridge.py`, neither of which carried the marker); (5) `git diff origin/main...pr-541 --name-status` shows clean R099/R100 rename markers for the 4 file moves (blame preserved), with one D for `test_wedged_gpg_subprocess.py` and one M for `test_backend_subprocess_hang.py` (the re-merge target); (6) PR title uses valid Palantir-style `chore(ci):` prefix and the ==COMMIT_MSG== block has contiguous `Closes: #522` + `Signed-off-by:` trailers.

PR #535, opened 2026-05-02
chore(ci): cap merge-bot sweep fan-out + needs fix gate

aidanns:
## Review notes

Caps the merge-bot `push:main` sweep at `N = vars.MERGE_BOT_SWEEP_MAX` (default 3) eligible PRs per main-advance event, eliminating the thundering-herd of CI re-runs that saturated the runner queue during the 2026-05-01 #440 milestone push.

### What changed

The sweep job's single bash step is rewritten to:

1. List open `automerge` PRs with `number,headRefName,title,labels` in one `gh pr list` call (frozen snapshot for downstream filters).
2. For each PR: skip if it carries the `needs fix` label (the PR-level park gate that `label-needs-fix` (#503/#515) maintains in response to `workflow_run.completed`); otherwise issue a single `gh api graphql` query that walks `pullRequest.commits(last: 100).nodes.commit.statusCheckRollup.state` and treats the PR as eligible iff any commit's rollup has ever been `SUCCESS`.
3. Partition the eligible set into release PRs (head ref `^release/X.Y.Z$` AND title `^chore(release): X.Y.Z$`, both required) and other PRs; sort each ascending by PR number; concatenate release-first; head at `${{ vars.MERGE_BOT_SWEEP_MAX || '3' }}`.
4. Log `total / eligible / needs-fix-excluded / cap / dispatched / numbers` so the sweep decision is observable.
5. Dispatch `gh workflow run` for the head-N PRs (existing behaviour, just narrower).

### Eligibility decision principle

The eligibility filter reads only GitHub-native data — never the bot's own prior outputs. Comment-text / commit-author heuristics are explicitly rejected because they create circular dependencies with the bot's implementation drift. The GraphQL `statusCheckRollup` walk is a structurally-independent signal, and the `needs fix` label is maintained by an orthogonal job (#503/#515) on `workflow_run.completed` — not by this sweep job.

### What this does NOT change

- The per-PR `merge-bot-<n>` concurrency group on the `merge` job (the source of truth for *per-PR* serialisation) is untouched — this cap is purely about *across-PR* fan-out.
- No trigger surface change: `pull_request: labeled`, `workflow_run: completed`, `pull_request_review`, and `workflow_dispatch` all behave identically. The cap applies only to the `push:main` sweep.
- The `label-needs-fix` job's logic and trigger surface are untouched.

### Implementation choices

[...]

PR #542, opened 2026-05-02
chore(ci): narrow needs-fix label gate to required failures

### LINKED ISSUE #536: merge-bot: `label-needs-fix` over-applies on superseded/infra failures (CANCELLED, STARTUP_FAILURE, etc.)
## Decisions (resolved 2026-05-02 in /workflow:implement session wfi-2026-05-02-7ebf)

- **TIMED_OUT classification**: NO-OP. Move it alongside `CANCELLED` / `STARTUP_FAILURE` / `ACTION_REQUIRED` in the no-label-change branch. Only `FAILURE` (and `state == FAILURE` / `state == ERROR`) trigger `needs fix`.
- **Contexts-extraction refactor**: composite action under `.github/actions/read-required-contexts/`. Both the existing `Inspect required checks` step in `merge-bot.yml` (landed in #532 / PR for #528) and the new `label-needs-fix` consumer share this action. The existing inline implementation in `merge-bot.yml` is moved into the composite action; both call sites consume its output.

## Observed

Reproduced on PR #535 (2026-05-02 ~08:30–08:38 UTC):

\`\`\`
08:30:54Z  CI run #1 starts
08:30:58Z  CI run #2 starts (cancels #1)
08:31:05Z  CI run #3 starts (cancels #2)
08:33:25Z  CI run #4 starts (cancels #3)
08:33:55Z  github-actions[bot] applies "needs fix"
08:37:45Z  CI run #4 completes SUCCESS
08:37:53Z  github-actions[bot] removes "needs fix"
\`\`\`

Three rapid pushes superseded their predecessor CI runs. When a \`workflow_run.completed\` event fired for one of the cancelled runs, \`label-needs-fix\` evaluated the rollup, saw a \`CANCELLED\` check-run, classified it as \`FAILURE\`, and applied \`needs fix\`. Four minutes later the final SUCCESS run cleared it. There was never a real problem with the PR — every CI run on the final head was green from the start.

## Cause

\`.github/workflows/merge-bot.yml\` line 1062–1078 (the \`label-needs-fix\` job's rollup classifier):

\`\`\`yaml
if any(.[];
  ((.conclusion // "") | ascii_upcase | IN("FAILURE","TIMED_OUT","CANCELLED","STARTUP_FAILURE","ACTION_REQUIRED"))
  or ((.state // "") | ascii_upcase | IN("FAILURE","ERROR"))
) then "FAILURE"
\`\`\`

Two over-broad cuts in this classifier:

1. **Scope**: it evaluates the FULL \`statusCheckRollup\`, not just the required-check contexts from branch protection. So an unrelated check-run (e.g. a parallel-PR failure surfacing on this PR's rollup, or an environmental check the repo doesn't actually require) flips the label even when branch protection would let the PR merge. This is the same shape as #528, but on the label gate rather than the merge gate.
2. **Conclusion class**: \`CANCELLED\`, \`STARTUP_FAILURE\`, and \`ACTION_REQUIRED\` are all classified as \`FAILURE\`, but none of them correspond to a problem with the PR author's commit:
   - **\`CANCELLED\`** — almost always means a newer push or merge-queue event superseded the run. Author already moved past whatever the cancelled run was checking.
   - **\`STARTUP_FAILURE\`** — runner provisioning or workflow-yaml-parse failure. Infra problem, not the PR's diff.
   - **\`ACTION_REQUIRED\`** — workflow paused awaiting manual approval (first-time-contributor gate, environment protection, etc.). \"Reviewer needs to approve\" is a different signal from \"PR author needs to fix.\"
   - **\`TIMED_OUT\`** — ambiguous: could be a real performance regression in the PR's diff, or a flaky job exceeding its timeout for unrelated reasons. Reasonable to either keep or split into a separate signal; I'd lean toward keeping it as needs-fix since timeouts in the author's own code ARE author-actionable.
   - **\`FAILURE\`** / \`state == FAILURE\` / \`state == ERROR\` — genuine check failure. Author-actionable.

## Impact

\`needs fix\` is the sweep-eligibility signal (line 999 + line 1198–1219): when applied, the merge-bot skips the PR in its sweep over open automerge PRs until the label clears. False-positive applies don't cause permanent harm because the next workflow_run.completed event clears them on SUCCESS — but during the window between the cancelled-run completion and the final success completion, the PR is parked from the sweep, and a noisy \"PR was failing then was fixed\" signal lands in the timeline / notifications even when the author did nothing wrong. On a busy push burst (as in #535), the label flaps multiple times.

The label is also surfaced visually in the PR list, so \"needs fix\" appearing on a PR that's actually green is misleading state for a human reviewer skimming the queue.

## Design principle

\`needs fix\` should mean exactly: **\"the PR head, evaluated against branch protection's required checks, has at least one author-actionable failure that the PR author needs to address before this PR can merge.\"**

That implies two narrowings:

1. **Scope**: only required-check contexts from branch protection count. Mirror the approach #528 prescribes for the merge gate — derive the contexts from rulesets / legacy branch-protection API, not from the full rollup. (Bonus: keeps the label and merge gates in sync about what \"required\" means, with branch protection as the single authoritative source.)
2. **Classification**: only \`FAILURE\` (and arguably \`TIMED_OUT\`, debatable per above) maps to \`needs fix\`. \`CANCELLED\`, \`STARTUP_FAILURE\`, and \`ACTION_REQUIRED\` should NOT trigger the label.
   - \`CANCELLED\` on a non-latest run is supersession; the latest run will resolve to its own conclusion shortly.
   - \`STARTUP_FAILURE\` is infra; the user retries via re-run, not via a code change.
   - \`ACTION_REQUIRED\` is reviewer-actionable, not PR-author-actionable; if it warrants its own label, that's a separate signal (e.g. \`needs review approval\`).

## Suggested fix

After #528 lands the branch-protection-derived contexts list for the merge gate, refactor that contexts-extraction step into a reusable form (composite action, shared script, or sibling job that emits an output the label-needs-fix job consumes). Then:

- \`label-needs-fix\` iterates only those contexts (same per-context latest-by-startedAt dedup the merge job uses).
- For each, classify \`success | neutral | skipped\` → satisfied; \`null | empty | missing | in_progress | queued | waiting | pending | requested\` → pending (no label change); \`FAILURE\` (and \`TIMED_OUT\` if we keep it) → \`needs fix\`; \`CANCELLED | STARTUP_FAILURE | ACTION_REQUIRED\` → no label change (treat as transient / non-author-actionable).
- If overall classification flips from a previous \`needs fix\` apply, remove it; if it newly fails, apply it; otherwise no-op.

Worth deciding in the PR for this issue whether \`TIMED_OUT\` stays in the apply set or moves to no-op alongside the other transients — current default of \"keep\" is a safe choice for now since author-authored timeouts are real, but reasonable people could split the other way.

## Verification

- Reproduce the #535 scenario: open an automerge PR, push three commits in rapid succession, observe that no \`needs fix\` apply event fires for any of the cancelled CI runs (only the final completion of run-on-latest-head triggers a label decision).
- Synthesize a workflow that emits a \`STARTUP_FAILURE\` (or check the next time a real one occurs) and confirm \`needs fix\` is not applied.
- Existing positive case: a real test failure on the PR head still applies \`needs fix\` and is removed once the author pushes a fix.

## Relationship to other issues

- **#528** — same "narrow to branch-protection contexts" fix on the merge gate. This issue extends the design principle to the label gate; ideally both share the contexts-extraction implementation.
- **#525 / #527** — workflow_run / labeled trigger plumbing problems that are about WHEN the bot fires, not about WHAT it does once it fires. Disjoint from this issue.

aidanns:
==COMMIT_MSG==
Reproduced on PR #535 (2026-05-02 ~08:30-08:38 UTC): three rapid
pushes superseded their predecessor CI runs, the last run came
back green from the start, but `label-needs-fix` flipped 'needs
fix' on for 4 minutes because a `workflow_run.completed` event
for a CANCELLED predecessor run was classified as FAILURE. The
PR was never broken.

Two over-broad cuts in the previous classifier:

1. Scope: it evaluated the FULL `statusCheckRollup`, so an
   unrelated check-run could flip the label even when branch
   protection would let the PR merge. Same shape as #528, but on
   the label gate.
2. Conclusion class: CANCELLED, STARTUP_FAILURE, ACTION_REQUIRED,
   TIMED_OUT were classified as FAILURE, but none correspond to a
   problem with the PR author's commit.

Two narrowings:

- Required-context contexts only. Mirror #528's source-of-truth:
  the active branch-protection ruleset(s). The contexts-extraction
  logic that landed inline in #528 / PR #532 is lifted into a new
  composite action, `.github/actions/read-required-contexts`, so
  the merge job and the label-needs-fix job share one
  implementation. The merge job's downstream classifier is
  unchanged (the action's output shape is preserved: `contexts`
  newline-delimited + `source` indicator).
- FAILURE classifier narrowed to `FAILURE` / `state == FAILURE` /
  `state == ERROR`. CANCELLED / STARTUP_FAILURE / ACTION_REQUIRED
  / TIMED_OUT now map to no-label-change. Pending states still
  map to no-label-change (unchanged). An unknown future status
  also no-ops with a log line so a GitHub schema addition cannot
  silently flap the label.

The label-needs-fix job now mints the merge-bot App token (the
composite action's rulesets call requires `Administration: read`,
which the default GITHUB_TOKEN does not carry), checks out `main`
(needed to load the local composite action), and applies the
required-only classifier. Per-PR concurrency, the head-SHA-based
PR resolution, and the workflow-run-only trigger are unchanged.

Closes: #536
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

CI-only change with no user-visible effect — `==NO_CHANGELOG==` marker applied.

The diff is in two parts:

1. New composite action `.github/actions/read-required-contexts/action.yml` — extracted verbatim from the inline contexts-extraction step that landed in PR #532 (issue #528). Same rulesets-API-then-legacy-API fallback, same `(source, contexts)` outputs. Parameterised on `pr-number` + `github-token`.
2. `.github/workflows/merge-bot.yml`:
   - `merge` job: replace the inline 100-line bash step with `uses: ./.github/actions/read-required-contexts`. Downstream `Inspect required checks` step is unchanged because the action preserves the output shape (`steps.required.outputs.{source,contexts}`).
   - `label-needs-fix` job: rewrite. Now mints the App token (needed for `Administration: read`), checks out `main` (needed for `uses: ./...`), calls the same composite action, then iterates ONLY the required-context list with a per-context classifier:
     - `success | neutral | skipped` -> satisfied
     - pending states (empty / null / missing / in_progress / queued / waiting / pending / requested) -> no label change
     - `cancelled | startup_failure | action_required | timed_out` -> no label change (NEW — was FAILURE in the old classifier)
     - `failure | error` -> failing
     - unknown future status -> log + no-op (defensive against schema additions)

### Test plan

- [ ] Inspect the diff: `gh pr diff <pr#>` — confirm the merge job's downstream steps still consume `steps.required.outputs.{source,contexts}` and the label-needs-fix rewrite uses the same composite action.
- [ ] Confirm `yq '.' .github/workflows/merge-bot.yml` and `yq '.' .github/actions/read-required-contexts/action.yml` parse cleanly (verified locally before push).
- [ ] After merge, observe the next `workflow_run.completed` on a CANCELLED predecessor CI run does NOT flip `needs fix` on. The #535 reproduction was three rapid pushes producing CANCELLED runs; any future rapid-push burst should be a clean repro.
- [ ] After merge, confirm a real test failure on a PR head still applies `needs fix` and the label is removed once the author pushes a fix. (Same observable behaviour as today for the FAILURE / state == FAILURE / state == ERROR cases — the narrowing only affects the no-op cases.)
- [ ] Spot-check: the merge job's hard-fail behaviour on a CANCELLED required check is unchanged (the merge job's classifier was not narrowed — only the label gate was). Confirm by reading the `Inspect required checks` step.


aidanns:
Claude Reviewer: LGTM. Verified: (1) no scope creep — diff is exactly the composite-action extraction + label-needs-fix narrowing called out in #536, no #523 path-refresh content; (2) composite action `Permissions:` block correctly documents `Administration: read`, `docs/release/merge-bot-setup.md` Step 1 still lists it as `Administration: Read-only`; (3) label-needs-fix secrets-check + `actions/create-github-app-token@1b10c78c…` mint mirror the `merge` job shape; (4) `actions/checkout@de0fac2e…` pinned with `ref: main` matches the merge job's defence-in-depth pin and the SHA-pin convention in `.claude/instructions/tooling-and-ci.md`; (5) unknown-status `*` branch is sound — explicit cases cover `success/neutral/skipped`, pending, infra-cancellations, and `failure/error`, leaving only genuinely-novel statuses to log-and-no-op (cannot mask known failures); (6) the merge job's own `Inspect required checks` classifier is unchanged — TIMED_OUT/CANCELLED/STARTUP_FAILURE/ACTION_REQUIRED still hit the catch-all → failing branch there, intentional and out of scope; (7) action output shape `(source, contexts)` preserved so the merge job's downstream `Inspect required checks` step needs no changes; (8) trailers contiguous, `chore(ci):` prefix matches ADR 0037; (9) `==NO_CHANGELOG==` correctly applied for a CI-only change; (10) Decisions-block call (TIMED_OUT → no-op) implemented as decided despite the issue's Discussion bullet musing the other way.

PR #543, opened 2026-05-02
fix(ci): switch dependabot-adaptor trigger to pull_request_target

### LINKED ISSUE #537: dependabot-adaptor-bot: `pull_request` trigger uses Dependabot-scoped secrets; bot App secrets unreachable
## Symptom

PR #531 (the first Dependabot PR opened since `.github/workflows/dependabot-adaptor-bot.yml` was introduced on 2026-04-26) did not have its `==COMMIT_MSG==` block injected and was not labelled `automerge` / `no changelog`. The adaptor workflow ran (6 attempts via `synchronize`), all reporting `success`, but every run short-circuited at the secrets-check step:

```
Secret source: Dependabot
##[notice]CHANGELOG_BOT_APP_ID secret is not set; skipping adaptor run.
```

## Root cause

GitHub runs workflows triggered by `dependabot[bot]`-authored PRs in the **Dependabot secrets context**, not the Actions secrets context (`Secret source: Dependabot` in the run log is the giveaway). The `CHANGELOG_BOT_APP_ID` / `CHANGELOG_BOT_PRIVATE_KEY` App credentials are provisioned in the Actions store, so `${{ secrets.CHANGELOG_BOT_APP_ID }}` resolves to empty inside the run, the secrets-check guard at `dependabot-adaptor-bot.yml:84` flips `configured=false`, and the body-edit + label-add steps are skipped.

This is a latent bug that has existed since the workflow shipped (#473 / #474 era) but went unnoticed because PR #531 is the first Dependabot PR to actually exercise the live branch — every prior run was on a human-authored PR and was filtered out by `if: ... == 'dependabot[bot]'`.

## Fix (recommended): switch trigger to `pull_request_target`

`pull_request_target` runs in the base-repo context, so the existing Actions secrets resolve normally. No duplicated secret store to keep in sync on App key rollovers.

The workflow's current header comment rejects `pull_request_target` on fork-safety grounds. That concern doesn't apply here — the standard `pull_request_target` risk is running malicious PR-supplied code with elevated secrets, and **this workflow never checks out the PR**. Every step is either a bash block with hard-coded inputs (the `==COMMIT_MSG==` template) or a `gh pr edit` call by PR number. The body content fetched via `gh pr view ... --jq .body` is only `grep`'d for a literal marker. There is no PR-bytes → code-execution path.

The existing `if: github.event.pull_request.user.login == 'dependabot[bot]'` gate remains spoof-proof on `pull_request_target` (GitHub sets `user.login` from the authenticated author; contributors cannot impersonate the Dependabot App), and Dependabot PRs are always pushed to branches in the base repo, never from forks.

Required workflow changes:

```yaml
on:
  pull_request_target:
    types: [opened, reopened, synchronize]

permissions:
  contents: read
  pull-requests: read
```

Plus an updated header comment explaining the inverted security argument so a future reader doesn't revert on instinct, and ideally a guard / comment that introducing `actions/checkout` of the PR ref into this workflow requires re-evaluating the trust model.

## Alternative: duplicate the App secrets to the Dependabot store

Adding `CHANGELOG_BOT_APP_ID` / `CHANGELOG_BOT_PRIVATE_KEY` under Settings → Secrets and variables → Dependabot also unblocks the workflow without code changes. Tradeoff: two stores to rotate on App key rollover; easy to forget. Prefer the trigger change unless there's a reason not to.

## Recovery for PR #531

After the fix lands, push an empty commit to PR #531 (or close/reopen) so the `synchronize` event re-fires the now-functional adaptor. Idempotent.

## Docs

`docs/release/changelog-bot-setup.md` should call out the Dependabot-scoped-secrets gotcha and explain why the workflow uses `pull_request_target`.

aidanns:
==COMMIT_MSG==
PR #531 (the first Dependabot PR opened since the adaptor shipped)
was not adapted: every workflow run reported success but
short-circuited at the secrets-check guard with `CHANGELOG_BOT_APP_ID
secret is not set`. Workflows triggered by `dependabot[bot]`-authored
PRs run in the Dependabot secrets context (not the Actions secrets
context), so the App credentials provisioned in the Actions store
resolved to empty inside the run. Switch the trigger to
`pull_request_target`, which runs in the base-repo context and sees
the Actions secrets, instead of duplicating the App credentials into
a second store that has to be kept in sync on key rollovers.

The standard `pull_request_target` risk (a fork PR's code running
with elevated secrets) does not apply: this workflow never checks
out the PR ref, every `run:` block reads only `env:`-bound values
and `gh pr edit / view` arguments by PR number, and the only
PR-author-controlled byte stream that enters a `run:` block is the
PR body, which is `grep -Fq`'d for a literal marker. The
`if: user.login == 'dependabot[bot]'` gate remains spoof-proof
under `pull_request_target`. The header comment now walks through
the inverted security argument and adds a guard sentence flagging
that any future `actions/checkout` of the PR ref invalidates the
trust model. The changelog-bot setup doc grows a section calling
out the Dependabot-scoped-secrets gotcha so the next maintainer
hits the explanation before the workflow.

Closes: #537
==COMMIT_MSG==

### Test plan

- [ ] Confirm the YAML parses (`uv run python -c "import yaml; yaml.safe_load(open('.github/workflows/dependabot-adaptor-bot.yml'))"`).
- [ ] Confirm the `==COMMIT_MSG==` block is well-formed (the `pr-lint` check on this PR exercises the validator).
- [ ] Audit the workflow for inline `${{ }}` interpolation in `run:` bodies under the new `pull_request_target` trigger — every value flows through `env:` + `${VAR}` shell expansion (verified by `grep -n '\\$\\{\\{' .github/workflows/dependabot-adaptor-bot.yml`; only `env:` / `with:` mappings hit, no `run:` bodies).
- [ ] Recovery for the originally-affected PR: after this lands, push an empty commit to PR #531 (or close/reopen) so the now-functional adaptor's `synchronize` event re-fires.


changelog/@unreleased/pr-543-fix-dependabot-adaptor-trigger.yml
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2026 Aidan Nagorcka-Smith
+#
+# SPDX-License-Identifier: MIT
+
+type: fix
+fix:
+  description: |
+    `dependabot-adaptor-bot.yml` now triggers on `pull_request_target`
```
aidanns: Claude Reviewer: `changelog-lint` rule 5 (`changelog/scripts/lint.py`, `MAX_DESCRIPTION_WORDS = 25`, single-sentence cap) fails this entry: `description` is 65 words across three sentences, and the `check-changelog / format` job already errored on it (job 74039964203). The block reads as the body/changelog hybrid that the lint was added to prevent — drop it to a single ≤25-word sentence (e.g. "Trigger `dependabot-adaptor-bot.yml` on `pull_request_target` so the App credentials resolve under the Dependabot secrets context (#537).") and move the trigger / secrets-context rationale into the `==COMMIT_MSG==` block in the PR body, which already covers the same ground. Without this, `automerge` will not pass the required-checks gate.

aidanns: Claude: Done!

PR #552, opened 2026-05-02
fix(ci): inject ==NO_CHANGELOG== marker from dependabot-adaptor

### LINKED ISSUE #551: fix(ci): dependabot-adaptor and changelog-bot fight over the `no changelog` label
## Symptom

Dependabot PRs land in CI with the \`no changelog\` label flipping on and
off. The PR ends up red on \`check-changelog / exists\` and \`/ format\`,
forcing manual intervention even though the dependabot-adaptor is
explicitly designed to keep these PRs green.

Concrete example — PR #531 (\`dorny/paths-filter\` 4.0.1):
\`\`\`
10:47:10 changelog-bot[bot] LABELED   no changelog   ← adaptor (run 25250203202)
10:47:28 changelog-bot[bot] UNLABELED no changelog   ← changelog-bot Arm 2
11:39:19 changelog-bot[bot] LABELED   no changelog   ← adaptor again
11:40:34 changelog-bot[bot] UNLABELED no changelog   ← changelog-bot Arm 2 again
\`\`\`

## Root cause

Both \`.github/workflows/dependabot-adaptor-bot.yml\` and
\`.github/workflows/changelog-bot.yml\` mint installation tokens from the
same \`agent-auth-changelog-bot\` GitHub App, so the GitHub-recorded
actor for either bot's \`labeled\` event is the same login
(\`agent-auth-changelog-bot[bot]\`).

The changelog-bot decision tree (\`scripts/changelog/bot.py\`) has two
relevant arms:
- **Arm 1** (\`bot.py:594\`): if \`==NO_CHANGELOG==\` is in the PR body,
  add (or keep) the \`no changelog\` label.
- **Arm 2** (\`bot.py:612\`): if the marker is absent **and** the label
  is present **and** \`label_was_applied_by_bot\` returns true, **remove**
  the label.

\`label_was_applied_by_bot\` (\`bot.py:421\`) only checks whether the
most recent \`labeled\` event for \`no changelog\` was authored by the
configured \`bot_identity.login\`. Because the dependabot-adaptor uses
the same App, that check returns true even though the adaptor — not
the changelog-bot itself — applied the label. The adaptor injects a
\`==COMMIT_MSG==\` block into the body but **not** a \`==NO_CHANGELOG==\`
marker, so Arm 1 doesn't fire and Arm 2 strips the label.

## Proposed fix

Have the dependabot-adaptor inject a \`==NO_CHANGELOG==\` marker into
the PR body alongside the existing \`==COMMIT_MSG==\` block (single-line
edit in \`dependabot-adaptor-bot.yml\` around the \`printf\` that builds
the new body). With the marker present in the body, the changelog-bot
takes Arm 1 on its next wake-up and keeps the label sticky. The
adaptor stops back-channel-labeling and starts using the system's
existing intent-declaration mechanism.

Idempotency is preserved: the adaptor's existing
\`grep -Fq '==COMMIT_MSG=='\` short-circuit guards against double-prepending.

## Alternatives considered

- **Separate App for the adaptor.** Cleanest trust separation but a
  setup chore (App provisioning, two more secrets to keep in sync on
  rollover); the trust argument in the adaptor workflow header would
  also need to be re-evaluated.
- **Exception in \`bot.py\` Arm 2 for Dependabot PRs.** Couples the
  changelog-bot to knowledge of who else holds its App credentials.

## Acceptance

- A fresh Dependabot PR's labels stabilise to \`no changelog\` +
  \`automerge\` (plus Dependabot's own \`dependencies\` /
  \`<ecosystem>\`) and stay that way without flapping.
- \`check-changelog / exists\` and \`/ format\` are green on the PR.
- merge-bot picks up the PR once required checks are green.

## Out of scope

- Whether to migrate the adaptor to a dedicated App (option 2) or
  add a Dependabot-author exception to \`bot.py\` Arm 2 (option 3).
  Both are tracked here as alternatives but not in scope for this
  fix.

aidanns:
==COMMIT_MSG==
The dependabot-adaptor and changelog-bot share the
agent-auth-changelog-bot GitHub App identity. The adaptor applied the
`no changelog` label directly; the changelog-bot's Arm 2 then saw a
PR body without a `==NO_CHANGELOG==` marker and a label its own
identity had applied, and stripped the label. Dependabot PRs landed
red on the changelog gate every time despite the adaptor having just
fixed them.

Have the adaptor inject `==NO_CHANGELOG==` into the PR body alongside
the existing `==COMMIT_MSG==` block. Arm 1 fires on the next bot
wake-up and keeps the label sticky. The marker sits outside the
COMMIT_MSG block so it stays out of the squash commit body.

Heal PRs adapted by an earlier version of the workflow in-place: a
per-marker check splits the run into three arms (skip when both
markers are already present, prepend only `==NO_CHANGELOG==` when
the block is already in place, full inject otherwise). The legacy
arm leaves the existing `==COMMIT_MSG==` block alone so the squash
commit body does not change.

Closes: #551
==COMMIT_MSG==

### What changed

- `.github/workflows/dependabot-adaptor-bot.yml`:
  - Per-marker check (`grep -Fq` for `==COMMIT_MSG==`, `grep -Fxq`
    for `==NO_CHANGELOG==`) replaces the single short-circuit.
  - Three-arm dispatch: skip / legacy-heal / fresh.
  - Legacy-heal arm prepends only the missing marker; the existing
    `==COMMIT_MSG==` block is untouched so `extract_block` keeps
    returning the same content for the squash commit body.
  - Header comment block updated to enumerate the new step and to
    record the trust argument for the marker (it sits outside the
    `==COMMIT_MSG==` block so `extract_block` does not pick it up
    into the squash commit body).
- `.github/workflows/tests/pr-lint-fixtures/valid-dependabot-adapter.md`:
  matches the fresh-PR `printf` output verbatim, per the fixture's
  own "keep this in sync" header.
- `.github/workflows/tests/pr-lint-fixtures/valid-dependabot-adapter-legacy-healed.md`
  (new): documents the legacy-heal arm's output shape so a
  regression in the new `printf` is caught alongside the fresh-PR
  case.

### Validation

- `python3 scripts/validate-commit-msg-block.py <fixture>` — both
  fixtures pass (`==COMMIT_MSG== block OK`).
- `python3 scripts/extract-commit-msg-block.py <fixture>` — both
  fixtures extract the same block content; the `==NO_CHANGELOG==`
  line stays out.
- `bot.has_no_changelog_marker(<fixture>)` returns `True` and
  `bot.extract_changelog_msg(<fixture>)` returns `None` on both
  fixtures, confirming Arm 1 fires and no spurious YAML auto-author.
- Manual bash simulation across all four input combos (both / only
  COMMIT_MSG / only NO_CHANGELOG / neither) — each routes to the
  expected arm and produces a body that the validator + extractor
  + changelog-bot accept.
- `uv run pytest tests/test_extract_commit_msg_block.py scripts/changelog/tests/` — 263 passed.

### Test plan

- [ ] Wait for CI to go green (the validator self-test should still pass).
- [ ] After merge, watch the next Dependabot PR — confirm `no changelog` label stops flapping and the changelog gate stays green without manual intervention.
- [ ] Confirm an in-flight Dependabot PR (e.g. PR #531) gets healed in-place on its next `synchronize` event (legacy-heal arm).
- [ ] Confirm the merge-bot picks up the next adaptor-handled Dependabot PR through the `automerge` label without manual `no changelog` reapplication.

### Out of scope

The two follow-up options recorded in #551 (separate App for the
adaptor; per-author exception in `bot.py` Arm 2) are not addressed
here; this PR is the cheapest fix. Re-evaluate if the marker-based
intent declaration ever needs to be revoked mid-flight.


aidanns (commented):
Claude Reviewer: LGTM with a few non-blocking observations (cannot formally approve own PR).

**What I checked**
- 3-branch correctness across the (commit_msg, no_changelog) Cartesian product. Confirmed `if "${has_commit_msg}" && "${has_no_changelog}"` invokes `true`/`false` as commands and dispatches each of the four combos correctly (verified locally).
- `grep -Fq` / `grep -Fxq` semantics. Asymmetry is intentional and benign: the `==COMMIT_MSG==` check is loose because `extract_block` re-validates strictness downstream; the `==NO_CHANGELOG==` check matches `bot.py:has_no_changelog_marker`'s `line.strip() == marker` discipline.
- Trust-model guard at lines 79-88. Still holds. `existing_body` is still only `grep -F`'d as fixed strings and written via `printf '%s'` to a tmp file consumed by `--body-file`. No new PR-bytes -> code-execution surface.
- Fixture/workflow sync. `valid-dependabot-adapter.md` matches the fresh-PR printf exactly.
- `set -euo pipefail` discipline + env-binding pattern unchanged.

**Non-blocking observations**

1. **Idempotency edge case.** A body with `==NO_CHANGELOG==` but no `==COMMIT_MSG==` (anomalous — Dependabot wouldn't write that) hits the `else` branch and produces a second `==NO_CHANGELOG==` line. Converges to a fixed point on the next run, and `has_no_changelog_marker`'s `any(...)` makes the duplicate harmless because it's outside the `==COMMIT_MSG==` block. Not worth handling in code, but worth knowing if it ever shows up in logs.

2. **Legacy-heal printf is uncovered by fixtures.** `valid-dependabot-adapter.md` covers the fresh-PR printf only. A regression in the legacy-heal printf (e.g. a dropped trailing blank line) wouldn't be caught by the fixture self-test. A `valid-dependabot-adapter-legacy-healed.md` fixture (existing block + prepended marker) would close the gap if you want symmetric coverage.

3. **Boolean-as-command idiom.** `\"\${has_commit_msg}\" && \"\${has_no_changelog}\"` works, but a one-line comment noting that `\$var` resolves to the `true`/`false` builtin would help readers who haven't seen the idiom before. Tiny nit.

4. **`# Labels are idempotent — gh pr edit --add-label is a no-op` comment** at L222-223 leans WHAT-not-WHY. Fine as-is given how short it is; mention only because CLAUDE.md asks for WHY-focused comments.

Verdict: approve (commented since GitHub blocks self-approval).

aidanns (commented):
Claude Reviewer: Second-commit follow-up review (HEAD `80868d6`).

**What I re-checked**

- Prior-review nits both addressed. The legacy-heal fixture (`valid-dependabot-adapter-legacy-healed.md`) is line-for-line consistent with the legacy-heal `printf` (NO_CHANGELOG, blank, then the existing legacy block + dependabot release notes). The boolean-as-command idiom now has a clarifying comment (lines 191-193).
- Three-arm idempotency. Each branch reaches a fixed point on re-run: skip arm is a literal no-op; legacy-heal arm transitions the body to "both markers present" so the next event hits skip; fresh arm same. Verified locally via bash simulation across the four `(commit_msg, no_changelog)` combos.
- Validator + bot agreement. Both fixtures pass `python3 scripts/validate-commit-msg-block.py`, both return `has_no_changelog_marker=True` and `extract_changelog_msg=None` under `scripts/changelog/bot.py`.
- Trust-model header at L79-88 still holds — no new `${{ github.event.pull_request.* }}` flowing into a `run:` block as code; `existing_body` is still only `grep -F`'d as a fixed string and written via `printf '%s' >> tmp` for `--body-file`.
- `grep -Fq` vs `grep -Fxq` asymmetry (loose for COMMIT_MSG, line-anchored for NO_CHANGELOG) remains intentional — `extract_block` re-validates strictness downstream, and `has_no_changelog_marker` does `line.strip() == marker` so the workflow's NO_CHANGELOG check matches it semantically.

**One non-blocking observation**

- Step name at L154 (`Inject ==COMMIT_MSG== block and apply labels`) is now stale — the step also injects `==NO_CHANGELOG==` and on the legacy-heal arm it injects *only* that marker. Updating to e.g. `Inject ==COMMIT_MSG== / ==NO_CHANGELOG== markers and apply labels` would make the Actions UI accurately reflect the work being done. Minor; not a blocker.

Verdict: LGTM (commented since GitHub blocks self-approval).

aidanns:
Claude: Done! Step name updated in de87fc8 to reflect both markers.

Issue #565, opened 2026-05-03
claude-instructions: tell agents to include Co-Authored-By + canonical Closes: trailers


aidanns:
## What to build

Two recent on-main commits (e06245a, 46c6fe4) landed without a
`Co-Authored-By: Claude` trailer, and one of them used the bare
`Closes #561` form instead of the canonical `Closes: #561`. The
merge-bot pasted the PR `==COMMIT_MSG==` blocks verbatim — both
omissions originated in the agent that authored the PRs.

Root cause is a documentation gap. The Bash tool's
"Committing changes with git" guidance auto-appends
`Co-Authored-By: Claude` to direct `git commit` invocations, but
the PR `==COMMIT_MSG==` block is hand-authored from a template
on a separate path. `CONTRIBUTING.md` § "Writing release-worthy
commits" describes the trailer format if used and notes
`Closes:` is canonical, but neither the new
`.claude/instructions/commit-messages.md` (#561) nor the PR
template states either as a hard requirement for agent-authored
PRs. Result: no session-startup-loaded rule, agents drift.

Add a "Trailers" section to
`.claude/instructions/commit-messages.md` codifying both rules
together. The instruction file is loaded into every Claude
session, so it's the most reliable home — strictly upstream of
the validator and the PR template.

Proposed wording (final phrasing at the author's discretion):

> ## Trailers
>
> When Claude authors the PR, the `==COMMIT_MSG==` block must
> include, in this order, immediately above `Signed-off-by:`:
>
> - `Closes: #N` linking the issue (canonical form, with the
>   colon — never the bare `Closes #N`)
> - `Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>`

Cross-link the new section from `CONTRIBUTING.md` § "Writing
release-worthy commits" so a human reviewer pushing back on a
draft has a single anchor to cite.

This is the documentation half of the fix; the validator-
tightening half (rejecting the bare `Closes #N` form so the
canonical form is mechanically enforced) is tracked separately.

## Acceptance criteria

- [ ] `.claude/instructions/commit-messages.md` gains a
      "Trailers" section that names `Co-Authored-By:` (with the
      exact `Claude Opus 4.7 (1M context) <noreply@anthropic.com>`
      identity string) and `Closes:` (canonical colon form) as
      required for agent-authored PRs
- [ ] Cross-link from `CONTRIBUTING.md` § "Writing release-worthy
      commits" to the new section
[...]

PR #564, opened 2026-05-03
chore(docs): expand commit-message rule rationale

aidanns:
## Review notes

Documentation-only change to CONTRIBUTING.md "Writing release-worthy commits" — extends it with the rationale layer for the new commit-message rules being codified for Claude in #561.

### Test plan

- [ ] CONTRIBUTING.md renders cleanly on GitHub (the table reflows after mdformat re-aligned column widths to fit the new convention-row).
- [ ] All five existing ADR 0037 references still resolve to design/decisions/0037-palantir-commit-prefixes-and-commit-msg-block.md.
- [ ] The 7ab4c6a counter-example is preserved verbatim (not replaced by the new 8859338 one).
- [ ] The new cross-link points at `.claude/instructions/commit-messages.md` (created by the parallel #561; brief broken-link window is acceptable per the issue).
- [ ] `scripts/format.sh --check` passes.
- [ ] `scripts/lint.sh` passes.

### Notes for the reviewer

- New subsections sit under the existing Body rules between "Don't restate the diff" (which keeps its 7ab4c6a counter-example) and "When a long body *is* warranted" (which keeps its three-cases list). Reading order is: lead-with-why -> don't-restate-the-diff -> causal-narrative-ordering -> brevity -> bullets -> arrows -> backticks-and-shouty-markers -> 8859338 counter-example -> when-a-long-body-is-warranted.
- The 8859338 example is its own subsection (not folded into one of the new rule bullets) because it demonstrates a different failure mode from 7ab4c6a — bundling and bullet-as-recap, where 7ab4c6a was straight diff-recap. Same structural shape (linked SHA, quoted issue, the keep-worthy form spelled out) so reviewers read both at the same level of abstraction.
- The CI-vs-convention table gets one new row for the five rules collectively; mdformat re-aligned the existing rows to match the new widest row.

==COMMIT_MSG==
The terse, machine-applied rules in
`.claude/instructions/commit-messages.md` (issue #561) tell Claude
what to do; reviewers pushing back on a draft need the WHY and an
anti-pattern they can point at. Extend CONTRIBUTING.md with five
rationale subsections (causal narrative ordering, brevity KEEP /
CUT, bullets-only-for-2+-parallel-sets, arrows-vs-prose, shouty-
marker rendering) and a second worked counter-example built on
commit 8859338, which fails on bundling rather than diff-recap.

The 8859338 example sits alongside the existing 7ab4c6a one —
together they cover the two distinct failure modes the new rules
[...]

PR #563, opened 2026-05-03
chore(claude): commit-message authoring rules for sessions

aidanns:
## Review notes

Implements #561 (parent #560). New file:
.claude/instructions/commit-messages.md, with one-line pointers from
CLAUDE.md's "Detailed instructions" list and from the new file back
to CONTRIBUTING.md's "Writing release-worthy commits" section. The
reverse pointer (CONTRIBUTING.md -> instruction file) is scoped to
the parallel #562 PR, per the issue body — brief broken-link window
is accepted.

The new file applies its own rules to itself: bare self-identifying
tokens (CONTRIBUTING.md, ==COMMIT_MSG==, NO_CHANGELOG), backticks
only for ambiguous-bare-word demonstrations and the arrows-vs-prose
worked example, brevity-first phrasing, terse cluster bullets in
the same shape as coding-standards.md and service-design.md.

### Test plan

- [ ] `task format -- --check` passes (ran locally pre-push)
- [ ] `task reuse-lint` passes (ran locally pre-push)
- [ ] CI: pr-title, pr-title-style, pr-body-commit-msg, dco
- [ ] Demoability: a fresh Claude session in this repo applies the
      rules on its first commit-message draft (verified by the
      orchestrator launching this very session — it loaded the file
      via the CLAUDE.md "Detailed instructions" pointer)

==COMMIT_MSG==
A fresh Claude session has no automatic exposure to the prose
conventions in CONTRIBUTING.md, so its first commit-message draft
in a new repo defaults to diff-recap bullets, scattered backticks,
and narrative-ordering that buries the observable behaviour. Adding
the rules to .claude/instructions/ pulls them into the
session-startup context so the draft lands within convention on the
first attempt instead of after review.

Closes #561
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

PR #588, opened 2026-05-03
improvement: reject bare Closes #N in COMMIT_MSG block

### LINKED ISSUE #566: pr-lint-validator: reject bare `Closes #N` (no colon) in COMMIT_MSG block
## What to build

The `==COMMIT_MSG==` validator at
`packages/pr-lint-validator/src/pr_lint_validator/commit_msg.py:53-55`
accepts both the canonical `Closes: #N` (with colon) and the bare
`Closes #N` (no colon) via a dedicated `GITHUB_KEYWORD_RE`:

    GITHUB_KEYWORD_RE = re.compile(
        r\"^(Closes|Fixes|Resolves)\s+(#\d+|...)\.?\$\"
    )

The PR template (`.github/PULL_REQUEST_TEMPLATE.md:42-44`) and
`CONTRIBUTING.md:979` both state the bare form is "accepted for
compatibility with the existing CHANGELOG style, but new PRs
should prefer `Closes: #N`". The recent on-main commit e06245a
(PR #563) used the bare form while its parallel sibling 46c6fe4
(PR #564) used the canonical one — the validator's two-form
acceptance let the inconsistency through, and `Closes:` is
already the form every other trailer uses.

Drop the bare form from the validator so the canonical
`Closes: #N` is mechanically enforced. The trailer-block parser
already validates colon-form trailers via the generic
`TRAILER_RE` at line 51 — once `GITHUB_KEYWORD_RE` is removed
from `_trailer_block_start_index` and `parse_trailer_block`, a
bare `Closes #N` will fall out of the trailer set, the body
will end at that line, and `check_blank_line_before_trailers`
or `check_signoff_present` will fail with a clear message.

Required side work in the same slice:

- Update `CONTRIBUTING.md:979` to drop the "with or without the
  colon" language and state the colon form is required.
- Update `.github/PULL_REQUEST_TEMPLATE.md:42-44` to remove the
  bare-`Closes #N` acceptance note.
- Confirm `scripts/changelog/` tooling (release-notes generator,
  predict-release-impact) doesn't grep for bare `Closes #N`
  anywhere — the changelog system reads the `Closes:` trailer
  via `git interpret-trailers`, so the canonical form already
  flows through, but verify with a quick grep before deleting
  the regex.
- Add a fixture under
  `.github/workflows/tests/pr-lint-fixtures/` (or wherever the
  validator's pytest cases live) covering the bare form being
  rejected, paired with the canonical form being accepted.

This is the enforcement half of a paired fix; the documentation
half — telling agents to use `Closes:` and `Co-Authored-By:` —
is tracked separately as #565.

## Acceptance criteria

- [ ] `GITHUB_KEYWORD_RE` removed from `commit_msg.py`'s
      trailer-detection paths (or kept only for the docstring's
      historical note); the bare `Closes #N` form fails the
      validator
- [ ] `CONTRIBUTING.md:979` and `.github/PULL_REQUEST_TEMPLATE.md`
      updated to require the canonical colon form
- [ ] New pytest fixture exercises the rejection path with a
      clear error message pointing at the offending line
- [ ] `scripts/changelog/` tooling verified to not depend on the
      bare form (single grep + a note in the PR body)
- [ ] Validator's existing pytest suite still passes

## Blocked by

Not strictly blocked, but worth landing #565 first so the
agent-facing documentation matches the validator's new behaviour
on the same release.

Depends on #565

aidanns:
==COMMIT_MSG==
Drop GITHUB_KEYWORD_RE from the trailer-detection paths in
pr_lint_validator.commit_msg and the legacy
scripts/validate-commit-msg-block.py so the canonical
`Closes: #N` (colon form) is the only shape the validator
accepts. The bare `Closes #N` now falls out of the trailer set,
the body extends down to that line, and the layout / signoff
checks fire with a clear message.

PRs #563 and #564 landed in the same week using the two
different forms — clear evidence the historical two-form
acceptance was just an inconsistency tax. The merge-bot's own
auto-close parser (scripts/parse-close-keywords.py) keeps
reading the bare form from already-merged commit bodies because
GitHub's UI auto-closer accepts it (distinct contract, distinct
file, scripts/changelog/ does not depend on it).

CONTRIBUTING.md, .github/PULL_REQUEST_TEMPLATE.md, and
.claude/instructions/commit-messages.md updated to state the
colon form is required. Pytest fixture
.github/workflows/tests/pr-lint-fixtures/invalid-bare-closes-form.md
(renamed from valid-bare-closes-form.md) plus two new pytest
cases pin the contract.

Closes: #566
==COMMIT_MSG==

## Review notes

Companion to #565 (which landed the agent-facing prose telling
Claude to use `Closes:` and `Co-Authored-By:` trailers); this PR
is the enforcement half — the validator now mechanically rejects
the bare form so the inconsistency caught between #563 (bare) and
#564 (canonical) cannot recur.

Blocker comment that gated the diff: https://github.com/aidanns/agent-auth/issues/566#issuecomment-4365856241
User approval to continue: https://github.com/aidanns/agent-auth/issues/566#issuecomment-4365945899

### Test plan

- [ ] ` task pr-lint-validator -- check-commit-msg --fixture-dir packages/pr-lint-validator/tests/fixtures` (or the equivalent pytest invocation under `packages/pr-lint-validator/tests/`)
- [ ] `uv run pytest packages/pr-lint-validator/tests/ -v` passes including the new bare-`Closes #N` rejection cases
- [ ] `uv run pytest tests/test_parse_close_keywords.py -v` passes (legacy auto-close parser still reads the bare form)
- [ ] Manual: confirm the new `.github/workflows/tests/pr-lint-fixtures/invalid-bare-closes-form.md` fixture fires the rejection error message
- [ ] CI: `pr-lint` workflow green on this PR (the canonical `Closes: #566` in the body above must satisfy the new validator)


packages/pr-lint-validator/tests/test_commit_msg.py
```diff
@@ -262,6 +262,52 @@ def test_validate_rejects_empty_block_after_comments() -> None:
...
+        "Body paragraph.\n\n"
+        "Closes #1\n"
+        "Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>\n"
+        "==COMMIT_MSG==\n"
+    )
+    with pytest.raises(commit_msg.ValidationError) as exc_info:
+        commit_msg.validate(body)
+    # The error message must point at the offending line so the
+    # author can find the bare ``Closes #N`` quickly.
+    assert "blank line" in str(exc_info.value)
```
aidanns: Claude Reviewer: The acceptance criterion in #566 says the rejection error should "point at the offending line." The current behaviour relies on `check_blank_line_before_trailers` firing once the bare `Closes #N` falls out of the trailer set — the error message says `... has no blank line between the body and the trailer block (line N is the first trailer; the line above it is non-blank)`, which is technically correct but does not name the actual cause (the bare `Closes #N` form). A contributor hitting this for the first time will likely be confused that their `Closes #N` line has been re-classified as body.

Consider adding a targeted detection in `commit_msg.py` that, when a bare `Closes|Fixes|Resolves <ref>` line appears between the body and the trailer block, raises something like: bare `Closes #N` (no colon) form is rejected since #566; use canonical `Closes: #N` instead.

The current test only asserts `"blank line" in str(exc_info.value)`, which masks that the rejection path is generic rather than diagnostic — if a future change to the layout check rephrases the message, this test still passes even though the bare-form rejection path no longer fires for the documented reason.

aidanns: Claude: Done! Added `check_no_bare_close_keyword` to both the `pr_lint_validator.commit_msg` library and the legacy `scripts/validate-commit-msg-block.py`, raising a targeted diagnostic that names the bare form as the cause and points at the specific reference number plus the canonical `Closes: #N` (with colon) shape. The check runs before the layout / blank-line-before checks so the bare-form message wins, and is anchored start-to-end so prose mentions like "this closes #123 because foo" don't false-positive. The existing bare-form rejection test now asserts on the targeted prose (`bare \`Closes #1\``, `no colon`, `Closes: #1 (with colon)`); a new test covers the `Fixes` / `Resolves` keyword variants; another new test pins the prose-mention non-match contract. Commit fbb3dfb.

Issue #568, opened 2026-05-03
improvement(ci): pre-baked macOS Tart image with Things 3 + TCC seed (Option A)


aidanns:
> *This was generated by AI during triage.*

## Background

Issue #485 chose **Option B** (install Things 3 + seed TCC at runtime per VM boot) over **Option A** (pre-baked custom Tart image hosted on GHCR). Option B's trade-off: faster initial delivery and no image-bake pipeline to maintain, at the cost of more moving parts per-PR (Homebrew install, cask download, TCC seed) that each represent a potential flake source.

This issue is a deferred breadcrumb to revisit Option A if Option B reliability proves insufficient. **It is intentionally filed in a deferred state — do NOT pick this up under normal triage cadence.**

## Reactivation criteria

Move this issue to `needs-triage` (or open-and-pick-up directly if you're maintainer) when **either** of the following is true:

- The `test-system-macos-tart` workflow produces ≥3 flakes in 30 days that are traceable to install-time issues — Homebrew install failure, Things 3 cask install regression, TCC seed schema mismatch on a macOS minor bump that landed faster than the seed script could be updated. **OR**
- The per-PR runtime of `test-system-macos-tart` (the macOS lane, not the no-op decide leg) consistently exceeds 20 minutes over 30 days, attributable to install-time costs that a pre-baked image would amortise.

Until one of those criteria fires, the cost-of-maintenance for an image-bake pipeline (image rebuild cadence, GHCR storage, Dependabot bumping the base-image digest, separate workflow + secrets to push, etc.) outweighs the per-run runtime savings.

## Scope (when revived)

Build a custom Tart image on top of `cirruslabs/macos-sequoia-base@<digest>`:

- Install Homebrew, `brew install --cask things`, install `uv`.
- Pre-seed `/Library/Application Support/com.apple.TCC/TCC.db` so the image boots with the Automation grant in place. (This bakes the per-Sequoia-minor TCC schema into the image; image rebuilds on minor bumps become mandatory.)
- Bake the image build into a sibling workflow that pushes to a project-owned GHCR repository on a schedule (weekly?) and on PRs that touch the image-bake recipe.
- Switch `test-system-macos-tart.yml` to pull the project image instead of the upstream base image; boot-prep collapses to "seed a to-do, run pytest" (~30s instead of ~7 min).
[...]

Issue #567, opened 2026-05-03
feature(ci): full agent-auth → things-bridge → things-client-cli-applescript → real Things 3 e2e in macOS Tart VM


aidanns:
> *This was generated by AI during triage.*

## Background

Issue #485 ships the macOS-Tart-VM CI plumbing — VM boot, Things 3 install, TCC seed — and a smoke test that proves `things-client-cli-applescript` can read a to-do back from a real Things 3 instance.

This issue tracks the next layer: a full end-to-end test where `agent-auth` and `things-bridge` are also running inside the VM, exercising the production HTTP and subprocess contracts against a fixture-seeded Things 3 database.

## Topology

Same shape as #485 (Topology A — everything inside the VM):

```
Tart VM:
  ├── agent-auth (FastAPI)
  ├── things-bridge (FastAPI) → subprocess → things-client-cli-applescript → osascript → Things 3
  ├── pytest driver
  └── Things 3.app (with fixture-seeded library)
```

This matches production deployment (all three services on the same macOS host) and avoids the SSH/HTTP cross-host complexity of running services on the GH runner with the VM as a "macOS slave".

## Things to design (open questions for triage / grilling)

- **Workflow shape:** extend the existing `test-system-macos-tart.yml` (one workflow, two test variants) or a sibling `test-system-macos-tart-e2e.yml`? Trade-off: shared boot-prep cost amortised vs. workflow file complexity.
- **Fixture database:** where does the seeded `.things` library live in the repo? Binary blob? Script that produces a known-state library by replaying a sequence of AppleScript `make new to do with properties` calls? How is it versioned across Things 3 minor updates that change the library schema?
- **Path-filter expansion:** add `agent_auth_updated` and `things_bridge_updated` filters to `ci.yml`'s `plan` job, feed them into the e2e workflow alongside the three from #485.
- **Test selector:** likely a sibling marker `requires_things3_fixture_db` alongside `requires_real_things3` from #485, with its own env-var gate.
- **Service lifecycle inside the VM:** how do `agent-auth` and `things-bridge` get started? Background processes via the boot-prep script? A process supervisor? `tmux`? A systemd-like wrapper? Each service has its own config; how do tests inject test-only configs (token-store path, bridge URL pointing at localhost, scope tier overrides)?
- **Health-check fixtures:** pytest fixtures that wait for service health endpoints (`/<svc>/health`) before exercising contracts. Probably belongs in `agent-auth-common`'s `tests_support`.
[...]

Issue #504, opened 2026-05-02
merge-bot: respond to @-mention comment commands (first: update)


aidanns:
## Problem

Today the only way to nudge `agent-auth-merge-bot` is to apply or remove the `automerge` label. There is no way to ask it to perform a discrete operation on a PR — most concretely, to update a PR's branch against `main` without also asking it to merge.

The bot already knows how to do an `update-branch`: the `Detect BEHIND and auto-update branch` step in `.github/workflows/merge-bot.yml` calls `PUT /pulls/{n}/update-branch` when a labeled PR is `BEHIND`. But that path only runs as part of a merge attempt. A contributor who wants to verify CI on top of the latest `main` *before* committing to a merge has no way to trigger it.

## Proposed change

Teach the bot to respond to `@agent-auth-merge-bot <command>` mentions in PR comments. First command:

- **`update`** — call `PUT /pulls/{n}/update-branch` for the PR the comment was posted on, using the same logic as the existing BEHIND auto-update step (pin `expected_head_sha`, surface `update-branch` API errors). Idempotent: if the PR is already up to date with `main`, no-op cleanly.

Triggering and execution:

- **Trigger:** add `issue_comment` (type: `created`) to `merge-bot.yml`'s `on:` block. Filter: comment is on a PR (`event.issue.pull_request != null`) and body starts with `@agent-auth-merge-bot ` followed by a known verb.
- **Authorization:** only honor commands from users with at least `write` permission on the repo (resolve via `GET /repos/{owner}/{repo}/collaborators/{user}/permission`). Anything below reacts 👎 and exits, so drive-by commenters cannot drive the bot.
- **Acknowledgement:** 👀 reaction on the trigger comment when the bot starts, then 👍 on success or 👎 on rejection. A follow-up comment is added only on structured failures (mirroring the existing `Claude: Cannot merge — …` style, e.g. `update-branch` API errors).
- **Self-mention guard:** ignore comments authored by the App itself, so any future command output that mentions the bot cannot loop.

Design the command surface as a small dispatch table keyed on verb so follow-up commands (`merge`, `cancel`, `rebase`, etc.) can be added without revisiting the trigger plumbing.

## Alternatives considered

- **Stick with the `automerge` label as the only trigger.** Doesn't address the on-demand update use case without an "apply, wait for behind-detection, remove" dance that risks accidentally merging.
[...]

Issue #407, opened 2026-04-27
changelog: lint description: field to one terse sentence


aidanns:
## Problem

The \`description:\` field in \`changelog/@unreleased/*.yml\` and \`changelog/<version>/*.yml\` files routinely runs to several sentences of mechanical detail. That prose feeds two audiences:

1. \`CHANGELOG.md\` and the GitHub release notes (via \`render_release_notes\`).
2. The release-PR commit body's \`==COMMIT_MSG==\` block (via \`render_commit_msg_block\`).

Both are user-facing release-notes audiences. A user reading the GitHub release page wants to know *what changed that might affect them*, not the implementation details. The verbose form is commit-archaeology material; it belongs in the originating PR's hand-authored \`==COMMIT_MSG==\` block (which becomes the squash commit body and is reachable via \`git show\`), not in the changelog YAML.

## Example

[\`changelog/0.15.1/pr-383-merge-bot-auto-update-behind.yml\`](https://github.com/aidanns/agent-auth/blob/main/changelog/0.15.1/pr-383-merge-bot-auto-update-behind.yml) currently runs to ~120 words across 5 sentences (mergeStateStatus polling, loop guard mechanics, permissions requirements, etc.). The user-facing version should be a single sentence:

> \`merge-bot\` now auto-updates a PR whose head sits behind \`main\` instead of treating the merge API's 405 as a hard failure.

The dropped detail is not lost — PR #383's \`==COMMIT_MSG==\` block (and therefore the squash commit body on \`main\`) carries it, and the YAML's \`links:\` array points back to the PR for anyone who wants the full archaeology.

## Why this is the right shape

The audience separation already exists structurally for non-release PRs:

- The squash commit body comes from the PR's hand-authored \`==COMMIT_MSG==\` block — verbose-friendly, archaeology-focused.
- The changelog YAML's \`description:\` only feeds CHANGELOG.md / the GitHub release page — user-facing, terse-required.

The two audiences do not need to share a single prose blob. Authors are currently over-writing the YAML to satisfy the deepest audience, and the shallowest audience suffers.

For release PRs the same terse form propagates into the release commit body via \`render_commit_msg_block\`, which is fine — a release commit body is a roll-up of release-notes one-liners under section headings (#397 already specifies the bulleted-list shape).

## Proposed fix

Add a lint to \`scripts/changelog/\` (or as a new \`pr-lint.yml\` job that runs the linter against changelog YAMLs touched by the PR):

[...]

PR #589, opened 2026-05-03
improvement: macOS Tart VM smoke for Things 3 + TCC seed

### LINKED ISSUE #485: Add macOS integration test workflow using Tart VM on macos-15 runner
## Background

We need a way to run integration tests that require a macOS environment in CI. macOS-hosted GitHub runners are themselves VMs, but `macos-15` (Apple Silicon) supports nested virtualization, so we can boot a clean macOS guest per run via [Tart](https://github.com/cirruslabs/tart) and Apple's `Virtualization.framework`. This gives us throwaway, reproducible macOS environments — useful for tests that mutate global state, install software, or otherwise can't safely share a host.

If a test only requires *macOS* (no clean-state guarantee), it should run directly on the runner — Tart is for the subset that needs isolation.

## Proposal

Add `.github/workflows/macos-integration.yml` that:

1. Runs on `macos-15` (escalate to `macos-15-xlarge` if RAM/disk is tight).
2. Installs Tart, pulls a `cirruslabs/macos-sequoia-base` image, and clones it to a per-run VM.
3. Boots the VM headless, waits for SSH, and rsyncs the workspace in.
4. Runs the subset of integration tests tagged as requiring macOS (selector TBD per language/test framework — see Open Questions).
5. Pulls JUnit XML back out and publishes via `mikepenz/action-junit-report`.
6. Tears down the VM in an `if: always()` step.

A working sketch of the YAML lives in the originating conversation; this issue is to productionize it for `agent-auth`'s actual test layout.

## Cost / scope guardrails

- macOS runner minutes are ~10× Linux. The workflow MUST be gated so doc-only and unrelated PRs don't trigger it. Start with a `paths:` filter, escalate to a precursor `changed-files` job if the filter proves too coarse.
- Image pulls are multi-GB and slow. Add an `actions/cache` step keyed on the image digest, or pre-bake a project-specific image pushed to GHCR.
- Default timeout: 60 minutes. Revisit once we see real run times.

## Open questions

- **Test selector.** What's the canonical way to mark a test as \"requires macOS\" in this repo? (pytest marker? Go build tag? Cargo feature? Something else.) The workflow's run command depends on this — needs to be settled before implementation.
- **Image choice.** Base image vs. `-xcode` variant. Xcode images are ~70 GB and may force `macos-15-xlarge`; only worth it if tests actually need Xcode/SDK toolchain.
- **Trigger scope.** Run on every PR, or only when a label (e.g. `needs-macos-ci`) is applied? Cost-driven decision.

## Acceptance criteria

- [ ] `.github/workflows/macos-integration.yml` exists and runs green on a PR that touches a macOS-tagged test.
- [ ] At least one integration test is tagged with the chosen selector and exercised by the workflow.
- [ ] Workflow is gated so it does NOT run on doc-only PRs.
- [ ] VM is torn down even on failure (verified by inspecting a deliberately-failing run).
- [ ] JUnit results surface in the PR checks UI.
- [ ] README or `design/DESIGN.md` documents how to add a new macOS-tagged integration test.

aidanns:
==COMMIT_MSG==
Adds a per-PR macOS-Tart-VM smoke lane that boots a clean
`cirruslabs/macos-sequoia-base` guest under Apple's
Virtualization.framework, installs Things 3, seeds an
`kTCCServiceAppleEvents` Automation grant for `osascript` ->
`com.culturedcode.ThingsMac` directly into TCC.db, then proves
that `things-client-cli-applescript todos list --status open`
round-trips a seeded `ci-smoke-todo` through real osascript.

Why a third tier: the existing unit + Linux-Docker tiers cannot
exercise the AppleScript dictionary, the TCC grant, or the
runtime envelope contract against a real Things 3 instance --
all three are macOS-only surfaces that currently regress
silently on every Things release. The `macos-15` GitHub runner
supports nested virtualization, so a per-PR Tart guest is the
cheapest non-self-hosted way to land that coverage.

Load-bearing decisions:

- Decide-job composition lives in the child workflow (`decide`
  ORs the three change inputs on `ubuntu-latest`) so the gating
  logic stays close to the workflow that owns it; the parent
  `ci.yml` only forwards the booleans.
- TCC seed uses a NULL `csreq` and a path-based `client` so the
  row stays forward-compatible across Sequoia minor versions
  whose csreq blob format may drift.
- No `actions/cache` for the Tart base image: it is ~30 GB,
  exceeding GitHub's 10 GB per-repo cache cap. Pre-baked image
  is the longer-term fix and is tracked as #568.
- ADR 0047 captures the SIP-off cirruslabs-base + sqlite3
  direct-write decision and the rejected alternatives
  (pre-baked image, tccplus runtime hack, label-gated
  incubation). `design/DESIGN.md` documents the new third test
  tier; `docs/operations/macos-tart-ci.md` covers the
  failure-artefact bundle, common failure modes, local
  reproduction, and the digest-bump procedure.

Plumbing only -- full agent-auth -> things-bridge -> CLI ->
Things 3 e2e is tracked as #567. The implementing PR's own
first CI run exercises the macOS lane (touches
`macos_tart_plumbing_updated`).

The TART_VERSION / TART_TARBALL_SHA256 / base-image digest in
the new workflow are placeholders that will be tightened to
real values once the workflow's first CI run validates the
boot-prep end-to-end; the structural shape is what the blocker
post asks for confirmation on.

Closes: #485
==COMMIT_MSG==

==CHANGELOG_MSG==
Per-PR macOS-Tart-VM smoke lane that boots a clean Sequoia guest, installs Things 3, seeds a TCC Automation grant for `osascript` -> `com.culturedcode.ThingsMac`, and round-trips a `things-client-cli-applescript todos list` against the live app. Path-filter-gated so the `macos-15` runner cost only lands on PRs that touch `packages/things-client-cli-applescript/`, `packages/agent-auth-common/`, or `scripts/ci/macos-tart/`. Plumbing only -- the full agent-auth -> things-bridge -> CLI -> Things 3 e2e is tracked as #567.
==CHANGELOG_MSG==

## Review notes

Plan: https://github.com/aidanns/agent-auth/issues/485#issuecomment-4365823830
User-approved blocker comment: https://github.com/aidanns/agent-auth/issues/485#issuecomment-4365856577 (then "approved - continue" at https://github.com/aidanns/agent-auth/issues/485#issuecomment-4365949198)

### Reviewer flags

- **JSON envelope key**: the system test asserts on `todos[*].name` from the CLI's stdout JSON (the `Todo` dataclass's field is `name`, not `title` -- the earlier draft of this section said `title` and was wrong). The CLI's envelope shape is `{"todos": [...]}`, confirmed against `packages/agent-auth-common/src/things_client_common/cli.py` and the `Todo` dataclass in `packages/agent-auth-common/src/things_models/models.py`. An inline comment in `test_macos_tart_smoke.py` records the envelope-key mapping.
- **Pinned-version placeholders**: `TART_VERSION`, `TART_TARBALL_SHA256`, and the base-image tag in `.github/workflows/test-system-macos-tart.yml` carry `# TODO(aidanns): pin in #485 follow-up` markers. They will be tightened to real values once this PR's first CI run validates the boot-prep end-to-end. Until then `git grep '0000000000000000000000000000000000000000000000000000000000000000'` and `git grep ':latest'` against the workflow file land a future bisector on the deferred-pin trail. The macOS lane runs on this PR (touches `macos_tart_plumbing_updated`), so `required-checks-passed` will not go green until the placeholders are real -- they have to land in a follow-up commit on this branch before merge.
- **Title selection** (`improvement:` not `improvement(ci):`): `ci` is in `INTERNAL_ONLY_SCOPES` per #504's review and would be rejected by `pr-lint.yml`.
- **Throwaway-VM curl-pipe escape hatch** (`.claude/instructions/tooling-and-ci.md`): the bootstrap script's `curl ... | sh` for Homebrew + uv was flagged against the standard. Resolved by extending the standard with an explicit three-condition escape hatch (ephemeral VM, no long-lived secrets, wrapper-pin gives no payload guarantee) and referencing it from each invocation in `bootstrap.sh`.

### Test plan

- [ ] `task check` -- workspace lints + format
- [ ] `task test` -- unit + Linux-Docker tiers (the new system test is `skipif`-gated by `AGENT_AUTH_REAL_THINGS3=1`, so it stays out of these tiers)
- [ ] On this PR: `test-system-macos-tart / macos-tart-smoke` runs green on the `macos-15` runner -- if any of the placeholder pins (Tart tarball SHA, base-image digest) are wrong, that lane will fail and we tighten them in a follow-up commit before merge.
- [ ] `required-checks-passed` aggregator turns green (confirms the new lane is wired into both `needs:` and `DIFF_DEPENDENT_JOB_IDS`).
- [ ] `tests/test_ci_diff_dependent_probe.py` passes (confirms the new workflow + leaf-check-run names are catalogued).


.github/workflows/test-system-macos-tart.yml
```diff
@@ -0,0 +1,262 @@
...
+    steps:
+      - name: OR the three changed-files inputs
+        id: compute
+        # Composition lives here, not in `ci.yml`'s call site — keeps
+        # the per-workflow gating logic close to the workflow that
+        # owns it (see ci.yml's `plan` job convention-doc comment).
+        # Inputs route through env vars rather than inlined `${{ ... }}`
+        # expressions inside `run:` so a future change to the input
+        # values cannot inject shell syntax (matching the convention
+        # `ci.yml`'s `plan` job uses for label-driven gating).
```
aidanns, line 99: Claude Reviewer: The `==COMMIT_MSG==` body bullet list reads as a diff recap (`.github/workflows/test-system-macos-tart.yml`: two-job shape ...; `ci.yml` gains three new dorny/paths-filter outputs ...; `scripts/ci/diff-dependent-jobs.sh` auto-extends ...; `scripts/ci/macos-tart/bootstrap.sh` orchestrates ...; etc.). `.claude/instructions/commit-messages.md` -> "Bullets" explicitly says "Never bullet a diff recap. If the bullets read as `renamed X, adjusted Y, added Z`, the commit body is replaying git show — collapse to one sentence on *why*." The "why" here is captured in the lead paragraph and the closing "plumbing only / placeholders" paragraphs; the bullet list mostly restates the file-by-file shape. Recommend collapsing the file-list bullets into a short prose paragraph that names the load-bearing decisions (decide-job composition lives in the child workflow, NULL csreq for forward-compat, no actions/cache because the image exceeds the 10 GB cap), or dropping them entirely since the lead paragraph already covers the user-visible behaviour.

packages/things-client-cli-applescript/tests/system/test_macos_tart_smoke.py
```diff
@@ -0,0 +1,101 @@
...
+        "things-client-cli-applescript envelope missing 'todos' key — "
+        f"the CLI changed its contract (got keys: {sorted(body)})."
+    )
+    todos = body["todos"]
+    assert isinstance(
+        todos, list
+    ), f"'todos' must be a JSON list per the contract; got {type(todos).__name__}"
+    names = [t.get("name") for t in todos]
+    assert names == [_SEEDED_TODO_NAME], (
+        "expected exactly one seeded to-do "
```
aidanns, line 99: Claude Reviewer: The PR description's Reviewer flag says "the system test asserts on `todos[*].title` from the CLI's stdout JSON", but this test asserts on `todos[*].name` (`names = [t.get("name") for t in todos]`). The actual code is correct (`packages/agent-auth-common/src/things_models/models.py`'s `Todo` dataclass has a `name` field, and `to_json()` returns `asdict()`), so the test is right and the PR description is the thing to fix. Worth tightening the description so a future reader doesn't follow the description and end up confused about why the test asserts on a different key than the description claims.

.github/workflows/test-system-macos-tart.yml
```diff
@@ -0,0 +1,262 @@
...
+          tart clone "${IMAGE}" "${VM}"
+
+      - name: Boot VM Headless and Wait for SSH
+        env:
+          VM: ${{ env.TART_VM_NAME }}
+        # Run Tart in the background so the workflow can keep
+        # going. Loop on `tart ip` (which returns the guest's
+        # DHCP-assigned IP once the boot reaches the network stack)
+        # plus an SSH probe. Cap at 180s to keep failures quick.
+        run: |
```
aidanns, line 184: Claude Reviewer: The placeholder values shipped in the merged tree are an active footgun if anyone picks up the workflow on `main` before the bump commits land. `TART_TARBALL_SHA256: "0000...0000"` will hard-fail `shasum -a 256 -c -` on every run, and `TART_BASE_IMAGE: "ghcr.io/cirruslabs/macos-sequoia-base:latest"` is a floating tag (the workflow header comment promises a digest pin). The user-approved blocker comment explicitly accepts these as TBD, so this is not a blocker for landing the structural shape — but worth ensuring the placeholder-tightening commits land **before** the `automerge` label, since per the project's flow this PR is its own first runner of the macOS lane (touches `macos_tart_plumbing_updated`) and `required-checks-passed` will refuse to go green with these placeholders in place. Optionally land a `TODO` marker (`TART_TARBALL_SHA256: "0000..."  # TODO(aidanns): pin in #485 follow-up`) so a future bisector grep on `0000` lands on the trail.

scripts/ci/macos-tart/bootstrap.sh
```diff
@@ -0,0 +1,162 @@
...
+TCC_DB="/Library/Application Support/com.apple.TCC/TCC.db"
+
+# The Sequoia `access` table column set is documented in
+# `docs/operations/macos-tart-ci.md` -> "TCC.db schema". Auth values:
+#   service                            = `kTCCServiceAppleEvents` (Automation)
+#   client                             = `/usr/bin/osascript`
+#   client_type                        = 1 (path-based identifier)
+#   auth_value                         = 2 (allow)
+#   auth_reason                        = 3 (user-set; closest match for "operator-seeded")
+#   auth_version                       = 1
```
aidanns, line 88: Claude Reviewer: The Homebrew install (line 1078) and the uv install (line 1180) both `curl ... | sh` install scripts whose bodies are not verified against a sha256. `.claude/instructions/tooling-and-ci.md` -> "Pin sha256 for tool binary downloads" extends the rule to install scripts: "Replace the pipe with download -> sha256sum -c -> execute, with the hash pinned in the tool-versions manifest and bumped together with any upstream change. ... A ref-pin alone (e.g. a commit SHA in the URL) is not sufficient: the install.sh body at that ref is still editable if the upstream repo is a moving target, and the pipe form never sees the bytes it ran." The script's inline rationale ("throwaway VM that holds no secrets") is reasonable for the brew side but the rule has no escape hatch for that case in the instructions file. `verify-standards.sh` only enforces the rule on `setup-toolchain/action.yml` so this won't fail CI, but it is a deviation from the documented standard. Two options: (a) pin the install scripts' sha256 in the tool-versions manifest and add a sha256-verify step before `| sh` (matches the standard), or (b) extend the standard's escape-hatch wording to cover throwaway-VM contexts and reference that wording in the bootstrap comment. The latter is cheaper and aligns with the existing rationale comment, but it does mean editing `tooling-and-ci.md` in this PR or a follow-up.

.github/workflows/test-system-macos-tart.yml
```diff
@@ -0,0 +1,262 @@
...
+      - name: Collect Failure Artefacts from VM
+        if: always()
+        env:
+          VM: ${{ env.TART_VM_NAME }}
+        # Pull the in-VM artefact directory back to the runner so the
+        # upload step has a single tree to ship. `|| true` on rsync
+        # so a VM that died before SSH came up doesn't mask the
+        # original failure.
+        run: |
+          set -uo pipefail
```
aidanns, line 251: Claude Reviewer: The hard-coded path `tart-extracted/tart.app/Contents/MacOS/tart` is an unverified assumption about the layout of the official Tart release tarball. Looking at upstream cirruslabs/tart's release page, the tarball's actual top-level layout has historically just been a bare `tart` binary (not a `.app` bundle) — the `.app` directory layout is for the optional GUI variant which is shipped as a separate `.dmg`. If the tarball expands to `tart` directly, the `ln -sf` line will silently create a broken symlink and `tart --version` will fail. Worth replacing with a defensive `find tart-extracted -type f -name tart -executable -print -quit` or similar so the discovery survives an upstream layout tweak. (This will get caught on the first CI run anyway, but it's the kind of placeholder-adjacent assumption that's nicer to flag explicitly.)

.github/workflows/test-system-macos-tart.yml
aidanns: Claude: Done! Collapsed the file-by-file bullet list in the PR body's `==COMMIT_MSG==` block into a "Why a third tier" paragraph plus a four-bullet "Load-bearing decisions" list (decide-job composition, NULL csreq, no actions/cache, ADR/docs surface). The new bullets name decisions rather than recap files. Pushed in ec7bdb1 (PR body update via `gh pr edit`).

packages/things-client-cli-applescript/tests/system/test_macos_tart_smoke.py
aidanns: Claude: Done! Updated the PR description's reviewer flag to say `todos[*].name` (matching the actual code: the `Todo` dataclass's field is `name`, and the test asserts `names = [t.get("name") for t in todos]`). The test itself was already correct -- only the description's wording needed fixing. Pushed via `gh pr edit` (no code change).

.github/workflows/test-system-macos-tart.yml
aidanns: Claude: Done! Added `# TODO(aidanns): pin in #485 follow-up` markers next to both `TART_TARBALL_SHA256` and `TART_BASE_IMAGE`, plus an `env:`-block prelude comment naming the bisector grep recipes (`git grep '0000...'` and `git grep ':latest'`). The PR body's reviewer-flag section is updated to spell out that `required-checks-passed` won't go green until the placeholders are real -- so they have to land in a follow-up commit on this branch before merge, which matches the workflow you described. Pushed in ec7bdb1.

scripts/ci/macos-tart/bootstrap.sh
aidanns: Claude: Took option (b): extended `.claude/instructions/tooling-and-ci.md` -> "Pin sha256 for tool binary downloads" with an explicit three-condition throwaway-VM escape hatch (ephemeral guest, no long-lived secrets, wrapper-pin-gives-no-payload-guarantee). Each `curl ... | sh` invocation in `bootstrap.sh` now carries an inline comment that walks through all three conditions and names the long-term migration target (#568, baking the tool into the pre-warmed base image). Picked (b) over (a) because the wrapper-pin gives false comfort -- pinning Homebrew's installer would not pin the per-cask payload that `brew install --cask things` actually fetches -- so a sha256 verify on the wrapper would be theatre. Pushed in ec7bdb1.

.github/workflows/test-system-macos-tart.yml
aidanns: Claude: Done! Replaced the hard-coded `tart-extracted/tart.app/Contents/MacOS/tart` with a `find tart-extracted -type f -name tart -perm -u+x -print -quit` discovery + a defensive empty-result branch that prints `ls -lR tart-extracted` and exits 1. The discovery survives both the bare-binary and `.app`-bundle layouts you flagged, plus any future upstream tweak. Pushed in ec7bdb1.

aidanns:
Claude: pushed orchestrator-level CI-fix in 6ce32b7 — added missing SPDX header on bootstrap.sh + tightened changelog description to fit the 25-word / 1-sentence cap (per scripts/changelog/lint.py #407). Both flagged checks should re-run green.

aidanns:
Claude: Blocked — the new `test-system-macos-tart` workflow ships values that don't match what cirruslabs/tart actually publishes:

- `TART_VERSION: "2.36.0"` — the latest cirruslabs/tart release is **2.32.1** (no `2.36.0` exists).
- Asset naming `tart-${VERSION}-arm64.tar.gz` — actual cirruslabs/tart releases publish a single `tart.tar.gz` asset (with a sibling `tart_<ver>_checksums.txt`).
- `TART_TARBALL_SHA256` is `0000…0000` placeholder.
- `TART_BASE_IMAGE: ":latest"` floating tag.

Net effect: `macos_tart_plumbing_updated` triggers on this PR (because it adds the plumbing), the smoke job runs, and the curl-and-shasum step fails before Tart even starts. The aggregator (`required-checks-passed`) reports `failure` because `test-system-macos-tart` reusable-workflow result is `failure`.

Three options to unblock:
1. **Pin to real values now** — bump `TART_VERSION` to `2.32.1`, change the URL/extraction to match `tart.tar.gz` (the upstream layout discovery via `find -perm -u+x` should still work), pin the sha256 against the real tarball, and pin the base image digest.
2. **Ship plumbing without arming the gate** — mark `test-system-macos-tart` as advisory (drop it from `required-checks-passed`'s `needs:` and `DIFF_DEPENDENT_JOB_IDS`) until the pins land in a follow-up PR. This is what the in-tree TODOs already promise (`TODO(aidanns): pin in #485 follow-up`).
3. **Restructure so plumbing-only changes don't trigger the smoke job** — change the `decide` job logic so `macos_tart_plumbing_updated == true` alone doesn't cause `should_run=true` (only the package paths do).

Removing the `automerge` label and applying `blocked` to the issue. Awaiting decision.

aidanns:
Claude: pushed Tart pin fix in f297c73 — bumped TART_VERSION to 2.32.1, corrected tarball name to tart.tar.gz, pinned real sha256 (cross-checked against upstream tart_2.32.1_checksums.txt), and pinned TART_BASE_IMAGE to GHCR digest sha256:cae08898…63. Re-applied automerge.

aidanns:
Claude: Blocked — option 1 (real Tart pins) is implemented but cannot land as currently structured.

**Diagnosis:** \`.github/workflows/test-system-macos-tart.yml\` is a brand-new workflow file. GitHub Actions registers reusable workflows from the **default branch** only — until this file lands on \`main\`, \`uses: ./.github/workflows/test-system-macos-tart.yml\` from \`ci.yml\` returns \`failure\` with **zero leaf jobs**, which trips \`required-checks-passed\`'s \`needs:\` evaluation.

Confirmed:
- \`gh api repos/aidanns/agent-auth/actions/workflows\` lists only \`test-system.yml\`, not the new \`test-system-macos-tart.yml\`.
- No \`workflow_run\` with \`path: .github/workflows/test-system-macos-tart.yml\` exists for this PR's head SHA.
- The CI run's job list contains zero \`decide\` / \`macos-tart-smoke\` jobs (the workflow's leaf jobs).
- The aggregator's \`needs\` JSON shows \`test-system-macos-tart: { result: \"failure\" }\`.

The Tart pins (commit \`f297c73\`: \`TART_VERSION=2.32.1\`, real sha256 \`8554ab4f…29\`, GHCR digest \`cae08898…63\`, asset name \`tart.tar.gz\`) are correct on the file — but they can't be exercised until GitHub registers the workflow.

Two viable paths to unblock:

1. **Defer the gate (option 2 from earlier).** Drop \`test-system-macos-tart\` from both \`required-checks-passed\`'s \`needs:\` and \`DIFF_DEPENDENT_JOB_IDS\` in this PR. The workflow file lands as advisory plumbing. Once on main, GitHub registers it; a one-line follow-up PR re-adds it to the gate. This is a ~5 line change to ci.yml.
2. **Split the PR.** Open PR-A that lands JUST \`test-system-macos-tart.yml\` + \`scripts/ci/macos-tart/\` + the docs/ADR (no \`ci.yml\` wiring). Once merged, GitHub registers it. Then this PR keeps only the \`ci.yml\` wiring + the test file, and CI can reach the workflow.

Path 1 is faster but ships a strictly-advisory lane that won't gate merges until the follow-up. Path 2 keeps the gate strict from the moment it lands but doubles the PR count. Removing the \`automerge\` label and applying \`blocked\` to the issue. Awaiting decision.

PR #540, opened 2026-05-02
chore(ci): collapse merge-bot workflow_run listener to just CI

aidanns:
==COMMIT_MSG==
The listener used to enumerate every PR-gating workflow by name so
each completion would fire the bot. After #466 landed the
`Required checks passed` aggregator inside ci.yml, branch
protection is gated on that single check — so any non-CI workflow
completion is no longer a meaningful re-evaluate signal for the
bot. Several entries in the old list also referenced workflows
that have already been deleted in earlier migrations and were
silently ignored by GitHub Actions.

Scope the listener to `CI` only and refresh the surrounding
prose — the per-workflow exclusion enumeration is no longer
relevant, and `CI` does not need an anti-recursion exclusion
because Merge Bot authors no commits. The `pull_request: labeled`
primary trigger, the `pull_request_review` re-fire trigger, the
`push: main` sweep trigger, and the `workflow_dispatch`
break-glass entry point are all unchanged.

Closes: #467
Signed-off-by: Aidan Nagorcka-Smith <aidanns@gmail.com>
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

Pure CI workflow tuning — no user-visible change. The merge bot's
evaluation logic, merge logic, sweep job, and label-needs-fix job
are all unchanged; only the `on.workflow_run.workflows` listener
list and its accompanying prose move. The listener's collapse to
`["CI"]` will be exercised end-to-end when this PR is merged via
the merge bot itself.

### Test plan

- [ ] CI is green on this branch.
- [ ] `pr-lint.yml` passes (the `==COMMIT_MSG==` block carries
  contiguous `Closes:` + `Signed-off-by:` trailers and no leading
  subject line).
- [ ] Once `automerge` is applied, the merge bot fires correctly
  on this PR — that is the intended live verification of the
  collapsed listener.

PR #590, opened 2026-05-03
fix: repair merge-bot rollup-dedupe regression guard

### LINKED ISSUE #580: fix(test): test_merge_bot_rollup_dedupe.py fails at collection — regex doesn't match post-refactor merge-bot.yml shape
> *This was generated by AI during triage.*

## Summary

\`tests/test_merge_bot_rollup_dedupe.py\` fails at pytest **collection** (not just at runtime — its module-level body crashes), which tanks the entire \`task test\` run for anyone executing the workspace suite locally. The test's regex-based extraction of the \`failed=\` jq filter from \`merge-bot.yml\` no longer matches the workflow's current shape.

## Root cause

The test pins both the \`failed\` and \`pending\` jq selectors in \`merge-bot.yml\`'s \`Inspect required checks\` step against fixture \`statusCheckRollup\` payloads. It does this by reading the workflow YAML at module load time, regex-extracting the jq programs, and shelling out to \`jq\` with the fixture data — see the docstring:

> *\"a future edit to the dedupe-or-predicate logic that drifts from this test will fail at the assert site, not at 'the workflow keeps running green but the bug is back'\".*

The extractor's regex (\`tests/test_merge_bot_rollup_dedupe.py\` lines 71-77) expects the historical shape:

\`\`\`bash
failed=\"\$(jq -r '
  ...filter body...
' <<<\"\${rollup}\")\"
\`\`\`

The workflow was refactored in #540 / #542 to a **per-context bash loop** that calls jq once per required check with \`--arg n \"\${ctx}\"\`, accumulating failures via shell string concatenation (\`merge-bot.yml\` lines 478-509). There is no longer a single \`failed=\"\$(jq -r '...')\"\` assignment to extract.

The dedupe property the test was guarding (\`sort_by(.startedAt // \"\") | (.[-1] // {})\`) **is still in the workflow** — it now lives inside the per-context jq invocation at line 493. So the regression guard's *intent* still matters; only its mechanism is broken.

## Why this wasn't caught in CI

The root \`tests/\` tree is not run by any current CI workflow. \`test-unit.yml\` is a per-package matrix scoped to \`packages/<svc>/tests/\` only; nothing else invokes pytest against the root tree. (See related observation noted alongside #579 — possibly the same Phase 5 cutover regression.)

## Possible shapes of the fix

- **Adapt the extractor to the new shape:** pull the per-context jq selector out of the loop body, plus a representation of the bash classification (success/pending/fail bucketing). Run the extracted jq inside a Python-side loop over the fixture rollups, replicating the workflow's iteration. Keeps the \"test exercises the actual filter\" property.
- **Run the loop body verbatim:** \`subprocess.run([\"bash\", \"-c\", <extracted-step-body>], env={\"REQUIRED_CONTEXTS\": ..., \"GH_TOKEN\": ...})\` against a mocked \`gh\` on PATH. More faithful to the workflow but bigger fixture overhead.
- **Hybrid:** test (a) the per-context jq selector for dedupe + classification correctness in isolation; (b) the bash loop's bucket-accumulation logic in a smaller separate test that doesn't need the full rollup payload.

The table-driven scenarios further down the file (fail-then-pass, pass-then-fail, mixed history, merge-bot self-filter, etc.) are still the right cases to pin — only the harness needs reworking.

## References

- Originating bug the test guards against: #347 (merge bot wedged on PR #346 by stale FAILURE entry in \`statusCheckRollup\`)
- Refactor that broke the harness: #540, #542
- Current broken regex: \`tests/test_merge_bot_rollup_dedupe.py\` lines 71-77
- Current workflow shape: \`.github/workflows/merge-bot.yml\` lines 478-509
- Related: #579 (root \`tests/\` tree not run in CI — would have caught this earlier)

aidanns:
==COMMIT_MSG==
PR #540 / #542 refactored merge-bot.yml's `Inspect required checks`
step from a pair of single jq pipelines (one for `failed`, one for
`pending`) to a per-context bash loop that calls jq once per
required check and buckets the result via a `case` statement. The
test's regex extractor was anchored on the old `failed="$(jq -r
'...' <<<"${rollup}")"` shape, so it returned no match against the
post-refactor workflow and the assert at module-load time aborted
pytest collection — tanking the entire `task test` run for anyone
exercising the workspace suite locally.

Rework the harness to extract the new shape:

- `_extract_classifier_loop` pulls the per-context loop body
  (initializers + `while ... done`) verbatim out of the
  `Inspect required checks` step. Anchored on the literal shape
  `fail=""\npending=""\nwhile IFS= read -r ctx; do ... done
  <<<"${REQUIRED_CONTEXTS}"` so a future refactor that moves the
  loop will surface here, not silently bypass the pin.
- `_classify(rollup, contexts)` runs the extracted loop under
  `bash -c` against a fixture rollup and a Python-supplied
  contexts list, then parses the `fail` / `pending` outputs back
  into Python lists. The extracted loop is the actual workflow
  bytes — not a hand-typed copy — so the test continues to fail
  at the assert site if the dedupe (`sort_by(.startedAt)` plus
  `(.[-1])`) is weakened or the merge-bot self-filter is
  removed.
- All seven existing scenarios (fail-then-pass, pass-then-fail,
  mixed history, merge-bot self-filter, legacy commit-status
  context, stale-IN_PROGRESS dedupe, genuinely-pending rerun)
  port over to the new harness with assertions on the
  `(failed, pending)` tuple. Two extra cases pin the empty-rollup
  and empty-contexts edge cases.
- `test_classifier_pins_dedupe_against_workflow_yaml` upgrades
  the structural pin: a future edit that drops `sort_by`, flips
  `(.[-1])` to `(.[0])`, or removes the `select((.workflowName
  // "") != "Merge Bot")` self-filter still surfaces here even
  if the data-driven scenarios happen to mask the regression
  (verified by mutation-injecting each of the three primitives
  against the workflow before committing).

Closes: #580
==COMMIT_MSG==

==NO_CHANGELOG==

## Review notes

Approach picked: option 2 from the issue body (`subprocess.run(["bash", "-c", <step-body>], env=...)`) over option 1 (replicate the loop in Python). Rationale: extracting the loop verbatim keeps the test pinned to the actual workflow bytes — including the `case` arm bucketing — without rebuilding the bash semantics in Python. The extra fixture overhead is one bash invocation per test (~50ms), which is well within the unit-test budget.

Mutation-test confirmation that the new harness catches the dedupe property: temporarily applied each of the three mutations below to `merge-bot.yml`, re-ran the test, observed the listed failures, then restored.

| Mutation | Tests that fail |
|---|---|
| `(.[-1] // {})` → `(.[0] // {})` | 7 of 11 (4 data-driven + structural pin + 2 collateral) |
| Drop `\| sort_by(.startedAt // "")` | 1 (structural pin) |
| `select((.workflowName // "") != "Merge Bot")` → `select(true)` | 2 (`test_failed_excludes_merge_bot_own_runs` + structural pin) |

The structural pin (`test_classifier_pins_dedupe_against_workflow_yaml`) is the catch-all: any single-line edit that removes `sort_by(.startedAt`, `(.[-1]`, or the `Merge Bot` self-filter substring trips it. Data-driven scenarios catch the semantic regressions on top.

### Test plan

- [ ] Local: `task test -- --fast` passes (sanity).
- [ ] Local: `uv run pytest tests/test_merge_bot_rollup_dedupe.py -v` returns 11 passed (was: 0 collected / 1 error before this PR).
- [ ] Local mutation sanity: temporarily apply each of the three mutations in the table above to `.github/workflows/merge-bot.yml`, re-run the test file, confirm the listed scenarios fail, restore the file.
- [ ] Local: `task lint`, `task format -- --check`, `task typecheck`, `task check` all green.
- [ ] CI: `Required checks passed` aggregator on `ci.yml` is green.

### DIFF
--- a/tests/test_merge_bot_rollup_dedupe.py
+++ b/tests/test_merge_bot_rollup_dedupe.py
@@ -5,34 +5,39 @@
 """Unit tests for the rollup dedupe predicate in merge-bot.yml.
 
 The merge bot (`.github/workflows/merge-bot.yml`) classifies required
-checks into `failed` / `pending` / `success` by piping the GraphQL
-`statusCheckRollup` payload through a jq filter. `statusCheckRollup`
-returns one entry per check *run*, not per check *name* — a retried
-check (workflow rerun, fix-then-push, close+reopen retrigger) appears
-multiple times, and the historical FAILURE entry never disappears
-from the rollup for the lifetime of the head SHA.
+checks into `failed` / `pending` / `success` by iterating the
+required-contexts list and, per context, piping the GraphQL
+`statusCheckRollup` payload through a jq filter that takes the latest
+run by `startedAt`. `statusCheckRollup` returns one entry per check
+*run*, not per check *name* — a retried check (workflow rerun,
+fix-then-push, close+reopen retrigger) appears multiple times, and the
+historical FAILURE entry never disappears from the rollup for the
+lifetime of the head SHA.
 
-The dedupe step (group by name+context, keep latest by startedAt)
+The dedupe step (sort by startedAt, take latest per context name)
 collapses each check name down to its most recent run before the
-failure / pending predicates evaluate the rollup. Without it, a
-fail-then-pass on the same SHA wedges merge-bot indefinitely — the
-bug surfaced on PR #346 / issue #347, where `changelog-lint` failed
-on the original `opened` event, was retriggered green via close+reopen,
-and the merge call was refused with `FAILED_CHECKS: changelog-lint`
-until the stale failed run was manually deleted via `gh run delete`.
+bash `case` classifier buckets it into success / pending / fail.
+Without it, a fail-then-pass on the same SHA wedges merge-bot
+indefinitely — the bug surfaced on PR #346 / issue #347, where
+`changelog-lint` failed on the original `opened` event, was
+retriggered green via close+reopen, and the merge call was refused
+with `FAILED_CHECKS: changelog-lint` until the stale failed run was
+manually deleted via `gh run delete`.
 
-The tests pin both selectors (failed *and* pending) by extracting the
-jq blocks straight out of the workflow YAML and feeding them
-hand-rolled rollup payloads. Pinning the predicates against the
-workflow file (rather than copy-pasting them into the test) means a
-future edit to the dedupe-or-predicate logic that drifts from this
-test will fail at the assert site, not at "the workflow keeps running
-green but the bug is back".
+The tests pin the dedupe by extracting the per-context bash loop
+straight out of the workflow YAML (the inner jq selector + the `case`
+classifier) and running it under bash against hand-rolled rollup
+payloads. Pinning the loop body against the workflow file (rather
+than copy-pasting it into the test) means a future edit to the
+dedupe-or-classifier logic that drifts from this test will fail at
+the assert site, not at "the workflow keeps running green but the
+bug is back".
 """
 
 from __future__ import annotations
 
 import json
+import os
 import re
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


aidanns:
Claude Reviewer: LGTM. Verified: (1) all 11 tests pass on the PR head; (2) the test extracts the per-context bash loop verbatim from `.github/workflows/merge-bot.yml` at runtime via `_extract_classifier_loop` and executes it under `bash -c` — not a hand-typed copy; (3) mutation-tested locally — replacing `sort_by(.startedAt // "")` with `select(true)` trips the structural pin, and flipping `(.[-1] // {})` to `(.[0] // {})` trips 7 of 11 scenarios — confirming a future refactor that weakens dedupe fails at an assert site, not silently; (4) scope is exactly one file (`tests/test_merge_bot_rollup_dedupe.py`); `merge-bot.yml` is untouched per the issue brief; (5) `ruff check`, `ruff format --check`, and `mypy` all pass on the new file; (6) PR title (`fix: repair merge-bot rollup-dedupe regression guard`) is 59 chars including the auto-appended `(#590)` suffix, uses the allowed `fix:` prefix; (7) `==COMMIT_MSG==` block has contiguous `Closes:` / `Co-Authored-By:` / `Signed-off-by:` trailers; (8) `no changelog` label applied; (9) the `test_failed_excludes_merge_bot_own_runs` rework is a nice touch — listing `Merge ==COMMIT_MSG== block as squash body` in the required contexts proves the self-filter short-circuits before classification (entry falls through to "missing" → `pending`), which is a stronger pin than the pre-refactor "filtered to nothing" assertion.

